PR #1, opened 2015-11-04
Core 7148 - Rename `donkey` to `terrain`

strootman:
As the title suggests, renamed all files, directories, and instances of `donkey` to `terrain`. Successfully tested locally


README.md
```diff
@@ -15,7 +15,7 @@ repositories under the iPlantArchive organization, should you need them.
 The top-level directories __lein-plugins__, __libs__, __services__, __tools__,
 __databases__, __ui__, and __ansible__ are the categories that the code falls
 into. Each direct subdirectory contains a separate project. For instance,
-__services/Donkey/__ contains the service that the Discovery Environment UI
+__services/Terrain/__ contains the service that the Discovery Environment UI
```
ianmcorvidae: Minor, but do we want to retain the capitalized subdirectory? We don't have it for much else.

README.md
```diff
@@ -15,7 +15,7 @@ repositories under the iPlantArchive organization, should you need them.
 The top-level directories __lein-plugins__, __libs__, __services__, __tools__,
 __databases__, __ui__, and __ansible__ are the categories that the code falls
 into. Each direct subdirectory contains a separate project. For instance,
-__services/Donkey/__ contains the service that the Discovery Environment UI
+__services/terrain/__ contains the service that the Discovery Environment UI
```
strootman: @ianmcorvidae @tedgin I made the change. No longer capitalized.

ansible/roles/ui/templates/nginx.conf.j2
```diff
@@ -62,7 +62,7 @@ http {
             proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
             proxy_set_header X-Forwarded_Proto $scheme;
 
-            proxy_pass http://{{ donkey.host }}:{{ kifshare.port }}/;
+            proxy_pass http://{{ terrain.host }}:{{ kifshare.port }}/;
```
ianmcorvidae: This is an odd line, to be going to the terrain host and the kifshare port. Not in your changes, I suppose

strootman: Heh. Separate issue methinks. Best to ask @dennisroberts71

services/metadata/conf/main/logback.xml
```diff
@@ -11,7 +11,7 @@
   </appender>
 
 
-  <Logger name="donkey" level="debug" />
+  <Logger name="terrain" level="debug" />
```
ianmcorvidae: Why is it we have configuration for the terrain logger in all of our services anyway? Presumably this should never be relevant except for terrain itself

services/terrain/doc/endpoints/misc.md
```diff
@@ -17,23 +17,23 @@
...
 
-The root path in Donkey can be used to verify that Donkey is actually running
+The root path in Terrain can be used to verify that Terrain is actually running
 and is responding. Currently, the response to this URL contains only a welcome
 message. Here's an example:
 
 ```
 $ curl -s http://by-tor:8888/
-Welcome to Donkey!  I've mastered the stairs!
+Welcome to Terrain!  I've mastered the stairs!
```
ianmcorvidae: Sadly this message doesn't make as much sense with the new name. Perhaps we should find something else humorous to say ;)

services/metadata/conf/main/logback.xml
strootman: Copy-pasta. If you blame it, my name will show up.
I can remove those if it seems appropriate.

services/terrain/doc/endpoints/terrain-v-metadactyl.md
```diff
@@ -1,28 +1,28 @@
...
+secured endpoints in Terrain. Specifically, the secured endpoints in Terrain
 actually require user authentication whereas those in metadactyl merely require
 information about the user making the request.
 
-## Calling Secured Donkey Endpoints
+## Calling Secured Terrain Endpoints
 
-All secured endpoints in Donkey require a query string parameter, `proxyToken`,
-whose value contains a CAS service ticket that Donkey can then use to verify
+All secured endpoints in Terrain require a query string parameter, `proxyToken`,
```
ianmcorvidae: This is pretty much just wrong now, cc: @dennisroberts71

services/terrain/doc/install.md
```diff
@@ -1,63 +1,63 @@
-# Installing and Configuring Donkey
+# Installing and Configuring Terrain
 
-Donkey is packaged as an RPM and published in iPlant's YUM repositories.  It
-can be installed using `yum install donkey` and upgraded using `yum upgrade
-donkey`.
+Terrain is packaged as an RPM and published in iPlant's YUM repositories.  It
+can be installed using `yum install terrain` and upgraded using `yum upgrade
+terrain`.
```
ianmcorvidae: Also wrong, though also out of scope. So much out of date documentation D:

services/terrain/doc/endpoints/misc.md
strootman: I'm open to suggestions :)

services/terrain/project.clj
```diff
@@ -7,13 +7,13 @@
       (string/trim (:out (sh "git" "rev-parse" "HEAD")))
       ""))
 
-(defproject org.iplantc/donkey "5.0.0-SNAPSHOT"
+(defproject org.iplantc/terrain "5.0.0-SNAPSHOT"
   :description "Framework for hosting DiscoveryEnvironment metadata services."
-  :url "https://github.com/iPlantCollaborativeOpenSource/Donkey"
+  :url "https://github.com/iPlantCollaborativeOpenSource/Terrain"
```
ianmcorvidae: Maybe point this to the new repo? This seems to be two iterations out of date. I'm happy to look into the other services, since I hadn't even thought about this line

services/terrain/doc/install.md
strootman: @dennisroberts71 how would you like to handle this? You want me to write up a bug?

set-versions
```diff
@@ -38,7 +38,7 @@ for project in $(find . -name project.clj -exec dirname '{}' \;); do
 
   cljproj=$(basename $project)
 
-  if [ "$cljproj" == "Donkey" ]; then
+  if [ "$cljproj" == "Terrain" ]; then
```
ianmcorvidae: continuing my bad habit of out-of-scope comments, I don't think we use this file any more and might want to remove it. not sure who would be the person to confirm/deny this

services/terrain/project.clj
```diff
@@ -7,13 +7,13 @@
       (string/trim (:out (sh "git" "rev-parse" "HEAD")))
       ""))
 
-(defproject org.iplantc/donkey "5.0.0-SNAPSHOT"
+(defproject org.iplantc/terrain "5.0.0-SNAPSHOT"
   :description "Framework for hosting DiscoveryEnvironment metadata services."
-  :url "https://github.com/iPlantCollaborativeOpenSource/Donkey"
+  :url "https://github.com/iPlantCollaborativeOpenSource/DE"
```
strootman: @ianmcorvidae Changed, but I think an issue needs to be made to capture others.

services/metadata/conf/main/logback.xml
ianmcorvidae: Probably out of scope. We can change it later, if we want/it matters, presumably.

set-versions
strootman: @johnworth will probably be the one to ask

And as far as "out of scope" is concerned, they're not out of scope comments.
As long as any other issues discovered are written up, it is all part of the review process

services/metadata/conf/main/logback.xml
strootman: If you want to write it up and put it into `logging/metrics` category, I'll take care of it later

ianmcorvidae: :+1:

set-versions
psarando: Are we no longer going to set versions in the project.clj files? I know we've been at `5.0.0` for a few releases now... Currently the `/status` endpoints use this version in their responses. Is that also changing when we use Docker labels for versions?

services/terrain/project.clj
psarando: I forgot about this project setting too. I think you're right and they should all now point to `DE`.

ianmcorvidae:
One other thing, just to make sure it's on your radar (probably is, but no hurt checking): Jenkins jobs: renaming them, pointing to correct directories, updating file include/exclude paths

set-versions
strootman: @psarando I think that is yet to be decided. I don't know if it is changing

ianmcorvidae: I was told to just use 5.0.0 for everything, and if that's the case there's not much reason to add it to the docker labels stuff (I was planning to just add git-ref info), and there's not much reason for this file. I'm not opposed to changing to using real version numbers again, but I wasn't around for whatever discussion led to the 5.0.0 decision in the first place so I didn't know that was relatively new/potentially malleable as a policy.

psarando: @johnworth could probably explain this better than I can, but these version numbers were used for RPM versions before we went with Docker, and the version got bumped for every QA drop, so they outpaced the DE version. `5.0.0` is just where these versions got to at DE v2.0 when we switched to Docker. Not sure why these versions haven't been updated since then.

ansible/roles/ui/templates/nginx.conf.j2
dennisroberts71: Yeah, that should probably be `kifshare.host`. I'll try to commit a fix for that today.

services/terrain/doc/endpoints/terrain-v-metadactyl.md
dennisroberts71: Yeah, the documentation needs a serious overhaul now that Donkey accepts JWTs.

services/terrain/doc/install.md
dennisroberts71: I'll file an issue to update the documentation.

set-versions
dennisroberts71: I'd actually like to see us start updating the versions again. It would be nice not to have to recompile all of the Clojure libraries every time I want to test an older version of one of the services on my machine.

strootman: I'll create an epic for this. We'll have to do it in an automated way, otherwise it will be too much manual, error-prone work.

psarando: Actually, we already have an automated script to update our services versions at the top level of the repo: [set-versions](https://github.com/iPlantCollaborativeOpenSource/DE/blob/master/set-versions).
It would be nice if it could also update the UI version in the ansible properties template too, but I'm not sure if that would be easy to accomplish.

databases/metadata/src/main/conversions/c192_2014090201.clj
```diff
@@ -58,16 +58,16 @@
...
-                               (get donkey-cfg "donkey.irods.home")
-                               (get donkey-cfg "donkey.irods.zone")
-                               (get donkey-cfg "donkey.irods.resc"))]
+    (let [fs-cfg (fs-init/init (get terrain-cfg "terrain.irods.host")
+                               (get terrain-cfg "terrain.irods.port")
+                               (get terrain-cfg "terrain.irods.user")
+                               (get terrain-cfg "terrain.irods.pass")
+                               (get terrain-cfg "terrain.irods.home")
+                               (get terrain-cfg "terrain.irods.zone")
+                               (get terrain-cfg "terrain.irods.resc"))]
```
psarando: I'm not sure about updating a conversion for a previous version.
I think it makes sense to leave these as `donkey` configs in this 1.9.2 conversion.

services/terrain/doc/endpoints/misc.md
dennisroberts71: The infinite is attainable with Terrain!

project-defs.edn
```diff
@@ -173,9 +173,9 @@
   :language :clojure
   :archive? true}
 
- "Donkey"
+ "Terrain"
```
psarando: Should this be lower-cased too? Not sure it matters here...

services/terrain/doc/install.md
```diff
@@ -74,41 +74,41 @@ modify this file so that it points to the correct hosts.
...
 log4j.appender.B.layout.ConversionPattern=%d{MM-dd@HH:mm:ss} %-5p (%13F:%L) %3x - %m%n
 
-# Use this appender when running Donkey in the background.
+# Use this appender when running Terrain in the background.
 log4j.appender.A=org.apache.log4j.RollingFileAppender
-log4j.appender.A.File=/var/log/donkey/donkey.log
+log4j.appender.A.File=/var/log/terrain/terrain.log
 log4j.appender.A.layout=org.apache.log4j.PatternLayout
 log4j.appender.A.layout.ConversionPattern=%d{MM-dd@HH:mm:ss} %-5p (%13F:%L) %3x - %m%n
 log4j.appender.A.MaxFileSize=10MB
```
psarando: @dennisroberts71: May as well get rid of these example configs too when the docs are overhauled.

dennisroberts71: :+1:

ui/de-lib/src/main/java/org/iplantc/de/client/services/SearchServiceFacade.java
```diff
@@ -18,7 +18,7 @@
  * <h2>Endpoints Used</h2>
  * <ul>
  * <li><a href=
- * "https://github.com/iPlantCollaborativeOpenSource/Donkey/blob/dev/doc/endpoints/filesystem/search.md#endpoints"
+ * "https://github.com/iPlantCollaborativeOpenSource/Terrain/blob/dev/doc/endpoints/filesystem/search.md#endpoints"
```
psarando: These URLs should become https://github.com/iPlantCollaborativeOpenSource/DE/blob/master/services/terrain/doc/endpoints/...

services/terrain/.gitignore
```diff
@@ -16,10 +16,10 @@
...
 /*.eld
 /target
-/donkey
-/donkey.spec
-/donkey*.tar.gz
-donkey.iml
+/terrain
+/terrain.spec
+/terrain*.tar.gz
+terrain.iml
```
dennisroberts71: I wonder if we should just make this `*.iml`. That would catch all IntelliJ idea files even if the project is named something else. It might not hurt to add `.idea/` to the list also.

PR #2, opened 2015-11-04
CORE-7221: Add labels to images denoting git-refs

ianmcorvidae:
You'll probably want to look at/review https://github.com/iPlantCollaborativeOpenSource/DE/compare/1bd3f81...eb25c4a (plus https://github.com/iPlantCollaborativeOpenSource/DE/commit/580f48b5181da90aeecab8b3deda57c138e4d614 -- just copying some stuff into the ui build script that's in the first block of commits for tools/services) first/primarily. Those commits set up the structure which the changes in this PR take advantage of, but I've done those first and directly to dev so that the baseIf images already have label information and all the build scripts are in place/nicely standardized.

This PR just makes the necessary updates to make all the services (except condor-log-monitor because we aren't even automatically building that any more), facepalm and filetool/porklock, and the UI build with these labels attached.

The end result here looks like:

```
mian@IPC-MCEWEN:~/Source/DE$ docker inspect -f '{{ json .Config.Labels }}' iplant-groups-with-buildenv-version | python -mjson.tool
{
    "org.iplantc.de.buildenv.git-ref": "eb25c4a84c9f3dcae874dbfb39b73ac98f3f05a4",
    "org.iplantc.de.iplant-groups.git-ref": "a61c588224de9fb367f5f6d6a436a31e3e47a6cf",
    "org.iplantc.de.javabase.git-ref": "720fa72ae7da270a52693b8cd62b618f381d8c02"
}

```

If #1 gets merged first, or the metadactyl rename happens, I'll fix this up with the new names, obviously.

### DIFF
--- a/docker/de-backend-buildenv/Dockerfile
+++ b/docker/de-backend-buildenv/Dockerfile
@@ -1,5 +1,7 @@
 FROM centos:centos7
 MAINTAINER John Wregglesworth <wregglej@gmail.com>
+ARG git_commit
+LABEL org.iplantc.de.buildenv.git-ref="$git_commit"
 
 ENV GOOS=linux GOARCH=amd64 GOVERSION=1.5
 ENV MVNVERSION=3.3.3
@@ -50,6 +52,5 @@ RUN npm install -g grunt-cli
 RUN lein help
 RUN git clone https://github.com/iPlantCollaborativeOpenSource/DE.git backend && \
     cd backend && \
-    git checkout %%GIT_COMMIT%% && \
+    git checkout $git_commit && \
     lein exec build-all.clj lein-plugins libs
-LABEL org.iplantc.de.buildenv.git-ref="%%GIT_COMMIT%%"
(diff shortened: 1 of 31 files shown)


psarando:
Should we also include the version defined by each `defproject` as a label as well? CORE-7221 mentions that the info reported by the `--version` flag of each service should be sped up, and these version numbers are reported by that flag. We will also start keeping our versions updated beyond `5.0.0` (see CORE-7241).

ianmcorvidae:
Yeah, true, we should probably add that information in as well if we're going to start updating it. I'll look into that.

ianmcorvidae:
Update! This now uses the fancy Docker 1.9 stuff @johnworth mentioned. Version stuff I'll do next (tomorrow, possibly).

docker/de-backend-buildenv/Dockerfile
```diff
@@ -50,6 +52,5 @@ RUN npm install -g grunt-cli
 RUN lein help
 RUN git clone https://github.com/iPlantCollaborativeOpenSource/DE.git backend && \
     cd backend && \
-    git checkout %%GIT_COMMIT%% && \
+    git checkout $git_commit && \
```
strootman: This is a nice touch.

ianmcorvidae: race conditions, begone :)

strootman:
We might want to also add a label to the `backwards-compat` image.

ianmcorvidae:
Hm, I'd avoided `backwards-compat` because it's pretty different from the others -- no jenkins job, rarely updated (for good reason!), etc. I suppose it won't hurt to add the git-ref label code since it defaults to `unknown` anyway, though.

strootman:
Sounds good.

PR #6, opened 2015-11-12
Changes to support local development

johnworth:
Facepalm's container is changed to be based off of Postgresql 9.5, which allows the unittest-dedb to be created without having to pull in the database tarballs, facepalm, and Postgresql separately.

The database container has been changed substantially. It no longer requires coordination between two separate containers in order to get running, you can now simply pull the unittest-dedb container image and run it, which will give you access to newly initialized instances of the de, notifications, and metadata databases.

de-db has been removed since it's not longer needed.

de-db-loader has been renamed to unittest-dedb.


tools/facepalm/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ubuntu:14.04
+FROM postgres:9.5
```
psarando: If prod is stuck using postgres 9.2, we probably should stick with that version in dev as well.

docker/unittest-dedb/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM discoenv/facepalm:dev
+
+COPY setup-pg-hbaconf.sh /docker-entrypoint-initdb.d/
+COPY setup-dev-database.sh /bin/setup-dev-database.sh
+COPY update-dev-database.sh /bin/update-dev-database.sh
+COPY pgpass /root/.pgpass
+RUN chmod 0600 /root/.pgpass
+RUN mkdir /data
+ENV POSTGRES_PASSWORD=notprod POSTGRES_USER=de PGDATA=/data
+ENTRYPOINT ["/docker-entrypoint.sh"]
```
psarando: Where does `docker-entrypoint.sh` come from? I don't see it in these commits, and I can't find it in facepalm's sources...

johnworth: It comes from the postgresql base image that facepalm uses. Since facepalm overrides the entrypoint to run its own stuff, I set the entrypoint back to docker-entrypoint.sh here so that postgres will start up.

docker/unittest-dedb/setup-dev-database.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+set -x
+set -e
+
+#psql -h 127.0.0.1 -U $POSTGRES_USER -d postgres -c "CREATE DATABASE de WITH OWNER $POSTGRES_USER;"
```
psarando: Is this line intentionally commented out?

docker/unittest-dedb/create-image.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/sh
+
+set -e
+set -x
+
+if [ $(docker ps | grep dedb | wc -l) -gt 0 ]; then
```
dennisroberts71: It might not hurt to make the regular expression a little more specific to avoid matching unrelated containers. Something like `grep '\sdedb$'` would match only containers with names that exactly match `dedb` (assuming that container names can't contain whitespace).

docker/unittest-dedb/setup-dev-database.sh
dennisroberts71: This may not ever be a problem, but in the case that `$POSTGRES_USER` is not equal to `de`, the `de` database will never be created. I didn't see a way to create a database only if it doesn't exist already. Another option might be to use `createuser --no-createdb` command to create the user, so that the command to create the database can always be executed without causing an error.

johnworth: I'm not sure it would be worth the effort to change this. The POSTGRES_USER is set in the Dockerfile and the setting is really only used when the image is built. You can override the POSTGRES_USER when firing up a container based on the image and you'll still have to connect to the database with the 'de' user. The POSTGRES_USER value is set in the Dockerfile.

PR #7, opened 2015-11-18
Systemd DE target - Core 7195

strootman:


ansible/deploy-all.yaml
```diff
@@ -4,6 +4,19 @@
...
+  sudo: True
+  gather_facts: false
+  tags:
+      - systemd
+      - systemd_target
+  roles:
+      - role: systemd-de-target 
+        up: false
+
+  include: systemd-de-target.yaml up=false
```
psarando: Are both the role and playbook-include required here?

strootman: No. Removed in 11fefba

PR #9, opened 2015-12-01
Enables creating logging, service, and nginx configs locally for any inventory

johnworth:
This set of changes is to allow us to generate config files for any environment locally.

For example:

    ansible-playbook -i inventories/production.cfg  --extra-vars "de_config_dir=$(pwd)/foo playbook=playbooks/local-services-cfg.yml owner=$(whoami) group=staff" --tags="service_config,logging_config" single-role.yaml

This would create the configuration files for the (non-existent) production environment and place them in a directory named "foo". The directory is created if it doesn't exist.

One thing that worries me is that I had to remove the following line in a couple of spots in order to get this to work.:

    when: "'{{group_name}}' in group_names"


ansible/roles/cfg-logging/tasks/main.yml
```diff
@@ -7,7 +7,5 @@
 
 - name: Place logging config file
   template: src={{src}} dest={{dest}} owner={{owner}} group={{group}} mode=0644
-  when: "'{{group_name}}' in group_names"
```
psarando: I think we need this protection in here (since the ansible/playbooks/10-services-cfg.yaml playbook uses the `services` host), so that configs for anon-files and kifshare don't get deployed in our other services host, and all our other service configs don't deployed on our anon-files/kifshare host, etc.

johnworth: That's what I figured. We'll need to have a discussion about whether to move forward with having a data container for the configs or not and how we would implement it if so. If it's one container per environment, then we don't really need the guard anymore, but if we do something else we probably will.

strootman: If we are deploying logging files to the host machine, this protection is needed so that the file is placed ONLY on the appropriate machine.
If we are getting configs from a docker container, then this protection is no longer needed.
The same holds for the checks in `cfg-service/tasks/main.yml`

ansible/roles/10-ui/meta/main.yaml
```diff
@@ -6,28 +6,35 @@ dependencies:
...
       src: ui.logback.xml.j2
+      tags:
+        - logging_config
 
     - role: cfg-service
       service_name_short: "{{de.http_server.service_name_short}}"
       group_name: ui
       src: ui/nginx.conf.j2
-      dest: /etc/nginx/nginx.conf
+      dest: "{{nginx_config}}/nginx.conf"
```
strootman: The variables used for this change need to be available at the `group_var` level, because this file is utilized in `ansible/roles/cfg-systemd-unit/templates/de-ui-nginx.service.j2`. 
```
de:
   ...
   http_server:
      ...
      config_dir: "{{de_config_dir}}/nginx"
```
Then, the line above will become `dest: "{{de.http_server.config_dir}}/nginx.conf`.

You would also have to change the volumes declaration in the nginx systemd file.
The line 
```-v /etc/nginx/nginx.conf:/etc/nginx/nginx.conf ``` could become 
```-v {{de.http_server.config_dir}}/nginx.conf:/etc/nginx/nginx.conf```

**Comment by strootman** [comment]
I don't see a way around the `group_name` checks without a custom inventory file. If the checks are in place, you could try passing the extra-var `group_name=all` on the command line. This will override the variable in all the roles, but I'm not sure if it will work.

**Comment by psarando** [comment]
If you've setup a `localhost.cfg` inventory, such as the one described in https://github.com/iPlantCollaborativeOpenSource/DE/blob/gh-pages-dev/ansible/work_instructions/local_configs.md, then shouldn't the `group_name` checks work as normal in these roles?

**Comment by strootman** [comment]
The trick is to get it to work without custom inventories.

**Comment by johnworth** [comment]
@psarando the problem that I'm trying to solve is how to create the configuration files for any inventory/environment locally all at once, not to create config files that work for a local deployment.

**Comment by tedgin** [comment]
Would it be possible to make the service_name, service_name_short, and compose_service computed values to reduce the number of variables?

**Comment by strootman** [comment]
@tedgin, the `service_name` and `service_name_short` variables are almost directly related to systemd usage.  I'm not sure if the intent is to drop systemd entirely, or to support is as an optional method for deploying the DE.
If we drop systemd entirely, we will have to be diligent in scrubbing all the group_vars.

File: ansible/DockerfileConfigs [hunk 3]
```diff
@@ -0,0 +1,16 @@
+FROM scratch
+
+COPY t /t
```
ianmcorvidae: Mentioned this to you at one point and it definitely shouldn't block anything, but if we want to save another ~MB (i.e., most of the size of these images), we could just do `FROM tianon/true`, which is a 125 byte image which contains and runs only a highly optimized x86 assembly version of `true`: https://github.com/tianon/dockerfiles/blob/master/true/true.asm

ianmcorvidae:
Not quite part of this PR, but related to it: Does the state of our jenkins builds and this PR produce a race condition? I think that we could end up in a situation where, if for some reason the build and deploy of a service happen before the appropriate de-configs build is finished, they'll be deployed with an old config image. Perhaps there's a jenkins plugin we can use to handle this for us -- we don't want to make every service build after any configuration change, but right now upstream/downstream builds are the only way we have to limit when things are built in Jenkins.

ansible/deploy-all.yaml
```diff
@@ -1,35 +1,38 @@
...
 ############################################
+# Third-party
+############################################
+- include: infra-amqp-brokers.yaml
+  when: not parasitic
+
+- include: infra-deploy-irods-elasticsearch.yaml
+  when: not parasitic
+
+############################################
```
strootman: This was removed in mainline. Should not re-add.

ansible/roles/10-ui/defaults/main.yml
```diff
@@ -0,0 +1,3 @@
+---
+local_config: "{{de_config_dir}}/nginx"
+nginx_config: "{{ local_config | default('/etc/nginx') }}"
```
strootman: This was renamed in mainline.

ansible/deploy-all.yaml
ianmcorvidae: Similarly other infra- playbooks below, I think

psarando: Either way, I thought it was decided by CORE-7345 that we will move all `infra-*` and `support-*` playbooks and roles into our internal ansible repo.

I'm not sure about the docker-compose related playbook and roles. We may still want to publish those?

ianmcorvidae:
(looking at the private changes alongside this) -- it appears you've removed the ability to configure heap sizes per-environment and set the hardcoded values to follow the non-production values except in the case of anon-files, which for some reason is set to the value for production instead. This will probably not work well under much of any load in dev and qa, and will under-utilize production's servers substantially (anon-files is the same, but most of the rest would be required to work in half the memory, and kifshare in a quarter of the memory); and of course, we'd been hoping to refine those values further in dev and qa, which this limits the possibility of doing reasonably

ansible/deploy-all.yaml
ianmcorvidae: It appears they were added only in the private repo; it presumably only works here because ansible-merge-tool copies everything together. So there's definitely something amiss, since this means we're publishing a playbook in the public repo which only we can run.

psarando: In that case, I don't think we should be including private playbooks from public playbooks.

johnworth:
Yeah, I meant to fix the heap-sizes so they're all the same as prod's heap sizes. Removing the configuration per-environment was on purpose. We've had an explosion in the number of variables (which are all global) that are scattered across multiple git repos, group_var files, playbooks, and roles. It's insane and I frankly got tired of tracking everything down, so I plopped in the values from de-2 (for the most part) with the intention of placing the prod values back in, which didn't happen before I left.

Basically, I want to keep ansible variables out of the docker-compose.yml files completely. I'd like us to be able to open up the docker-compose file at some point in the relatively near future and requiring ansible in order to use it would be annoying. We can always create per-environment docker-compose.yml files, if needed (I'd rather get all of the environments to have similar specs).

ansible/deploy-all.yaml
johnworth: Nuked the third party infra playbooks from deploy-all.yaml. Working on cleaning up the rest.

ansible/DockerfileConfigs
johnworth: Filed CORE-7399 internally to track this.

strootman:
While we may not be removing the systemd stuff completely, the `de-services-common` ansible role should not depend on the `de-systemd-de-target` role if the primary installation theme is docker-compose instead of systemd.

PR #11, opened 2015-12-02
Tool admin

ash3rz:
CORE-7209
Currently there is no way within Belphegor to make updates to the tools deployed to DE.  Requesting an update to Belphegor to be able to:

1. View a list of all the tools
2. Add a tool
3. Update an existing tool
4. Delete a tool


ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/toolAdmin/view/ToolAdminDetailsWindow.ui.xml
```diff
@@ -0,0 +1,87 @@
...
+            </container:child>
+            <container:child layoutData="{layoutData}">
+                <d:ToolImplementationEditor ui:field="implementationEditor"/>
+            </container:child>
+            <container:child layoutData="{layoutData}">
+                <d:ToolContainerEditor ui:field="containerEditor"/>
+            </container:child>
+        </container:VerticalLayoutContainer>
+    </g:FormPanel>
+</ui:UiBinder>
```
psarando: @ash3rz You can configure IntelliJ to always include a newline at the end of files.
Preferences > Editor > General, then check `Ensure line feed at file end on Save`
You can setup "Strip trailing spaces on Save" for Modified Lines there too.

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/toolAdmin/service/ToolAdminServiceFacade.java
```diff
@@ -0,0 +1,26 @@
...
+import java.util.List;
+
+/**
+ * Created by aramsey on 10/28/15.
+ */
+public interface ToolAdminServiceFacade {
+
+    void getTools(String searchTerm, AsyncCallback<List<Tool>> callback);
+
+    void getToolDetails(String toolId, AsyncCallback<ToolImportUpdateRequest> callback);
```
psarando: Since the new `ToolImportUpdateRequest` model is used to load tool details in addition to making import/update requests, and it already shares all the fields of the existing `Tool` model, we can add all the new required fields to the existing `Tool` model and use it everywhere this new `ToolImportUpdateRequest` model is used instead.

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/toolAdmin/view/ToolAdminViewImpl.java
```diff
@@ -0,0 +1,293 @@
...
+
+    }
+
+    private final class NameFilter implements Store.StoreFilter<Tool> {
+
+        private String filterText;
+
+        @Override
+        public boolean select(Store<Tool> store, Tool parent, Tool item) {
+            if (Strings.nullToEmpty(filterText).isEmpty()) {
```
psarando: A shortcut here could be `Strings.isNullOrEmpty(filterText)`.

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/toolAdmin/view/ToolAdminViewImpl.java
```diff
@@ -0,0 +1,292 @@
...
+            @Override
+            public void onSelect(SelectEvent event) {
+                Tool tool = detailsPanel.getTool();
+                if (detailsPanel.isValid()){
+                    presenter.updateTool(tool);
+                    dialogWindow.hide();
+                    try {
+                        grid.getSelectionModel().deselect(grid.getSelectionModel().getSelectedItem());
+                    } catch (UmbrellaException umb) {
+                        //do nothing -- throws error about DeviceList, VolumeList, and VolumesFromList having null IDs
```
psarando: There should be a better way of handling this than catching an `UmbrellaException`.
Since these IDs are not required in the add/update requests, maybe they can be removed from the UI's model, and the grid's listStore model-keys can be managed with AutoBean tags instead?

strootman: I agree with @psarando. The UI models should attempt to match the API-layer model as much as possible. There are other options using [interface segregation principle](https://en.wikipedia.org/wiki/Interface_segregation_principle), but I haven't looked into the code enough to weigh in.

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/toolAdmin/view/ToolAdminDetailsWindow.ui.xml
strootman: @psarando @ash3rz Didn't you guys configure these kind of settings [here](https://github.com/iPlantCollaborativeOpenSource/DE/tree/dev/ui/iplant/code-style/idea)? If not, then this should go in there as well.

psarando: Good point. So far that directory only contains the minimal required to import the [iPlant UI Layer Code Style.xml](https://github.com/iPlantCollaborativeOpenSource/DE/blob/dev/ui/iplant/code-style/idea/codestyles/iPlant%20UI%20Layer%20Code%20Style.xml) file, which is specific to Java code-style settings, but the `Ensure line feed at file end on Save` isn't part of those settings.
I can look into trying to include this setting in another file in that directory (maybe it's supposed to go in the [IntelliJ IDEA Global Settings](https://github.com/iPlantCollaborativeOpenSource/DE/blob/dev/ui/iplant/code-style/idea/IntelliJ%20IDEA%20Global%20Settings) file).

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/toolAdmin/view/ToolAdminDetailsWindow.ui.xml
```diff
@@ -0,0 +1,87 @@
+<!DOCTYPE ui:UiBinder SYSTEM "http://dl.google.com/gwt/DTD/xhtml.ent">
+<ui:UiBinder xmlns:ui="urn:ui:com.google.gwt.uibinder"
+             xmlns:container="urn:import:com.sencha.gxt.widget.core.client.container"
+             xmlns:gxt="urn:import:com.sencha.gxt.widget.core.client"
```
strootman: This namespace is not used.

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/toolAdmin/view/subviews/ToolImplementationEditor.ui.xml
```diff
@@ -0,0 +1,50 @@
+<!DOCTYPE ui:UiBinder SYSTEM "http://dl.google.com/gwt/DTD/xhtml.ent">
+<ui:UiBinder xmlns:ui="urn:ui:com.google.gwt.uibinder"
+             xmlns:container="urn:import:com.sencha.gxt.widget.core.client.container"
+             xmlns:gxt="urn:import:com.sencha.gxt.widget.core.client"
```
strootman: This namespace is not used.

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/toolAdmin/view/subviews/ToolVolumesFromListEditor.java
```diff
@@ -0,0 +1,221 @@
...
+    }
+
+    public boolean isValid(){
+        boolean valid = true;
+        for (ToolVolumesFrom volumesFrom : listStore.getAll()){
+            if (Strings.isNullOrEmpty(volumesFrom.getName()) || Strings.isNullOrEmpty(volumesFrom.getNamePrefix())){
+                return false;
+            }
+        }
+        return valid;
```
strootman: `valid` is always true. Either inline the value, or set `valid=false` in the if block above.

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/toolAdmin/service/impl/ToolAdminServiceFacadeImpl.java
```diff
@@ -0,0 +1,89 @@
...
+    private final String TOOLS_ADMIN = "org.iplantc.services.admin.tools";
+
+    @Inject
+    public ToolAdminServiceFacadeImpl() {
+    }
+
+    @Inject
+    private ToolAutoBeanFactory factory;
+    @Inject
+    private DiscEnvApiService deService;
```
strootman: These variables should go above constructor.

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/toolAdmin/service/impl/ToolListCallbackConverter.java
```diff
@@ -0,0 +1,32 @@
...
+import org.iplantc.de.client.models.tool.Tool;
+import org.iplantc.de.client.models.tool.ToolList;
+import org.iplantc.de.client.services.converters.AsyncCallbackConverter;
+
+import java.util.List;
+
+/**
+ * Created by aramsey on 10/28/15.
+ */
+
```
strootman: There are a number of places where the class javadoc has empty lines before the class declaration. For style consistency, the spaces should be removed.

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/toolAdmin/view/ToolAdminViewImpl.java
```diff
@@ -0,0 +1,285 @@
...
+
+import java.util.List;
+
+/**
+ * Created by aramsey on 10/27/15.
+ */
+
+
+public class ToolAdminViewImpl extends Composite
+        implements ToolAdminView, SelectionHandler {
```
strootman: This should have the class specifier. So `SelectionHandler` should be `SelectionHandler<Tool>`

ui/de-lib/src/main/java/org/iplantc/de/theme/base/client/admin/toolAdmin/ToolAdminViewDefaultAppearance.java
```diff
@@ -0,0 +1,412 @@
...
+
+    @Override
+    public String deleteToolSuccessText() {
+        return displayStrings.deleteToolSuccessText();
+    }
+
+    @Override
+    public String publicSubmitError() {
+        return iplantDisplayStrings.publicSubmitTip();
+    }
```
psarando: This error message is used in a few places, but I think its name a little confusing.
It's a good idea to reuse this `Please complete required fields` error message, but since it's now being used outside of the `Make app public` context, this property/method name should now be renamed to something more generic. Maybe `completeRequiredFieldsError`.

ash3rz: Should I rename publicSubmitTip in this branch with the current Jira or create a new ticket since it touches files outside this scope?

psarando: I would go ahead and do the rename in this pull request, since you're making the change to use it generically in code for this feature. So it's still a change related to this pull request.

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/toolAdmin/view/subviews/ToolTestDataParamsListEditor.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    private ModelKeyProvider<String> getModelKeyProvider() {
+        return new ModelKeyProvider<String>() {
+            @Override
+            public String getKey(String item) {
+                if (item == null) {
+                    return null;
+                }
+                return item;
```
psarando: This `null` check is redundant and `item` can just be returned here.

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/toolAdmin/presenter/ToolAdminPresenterImpl.java
```diff
@@ -0,0 +1,141 @@
...
+
+        });
+    }
+
+    @Override
+    public void updateTool(Tool tool) {
+        toolAdminServiceFacade.updateTool(tool, new AsyncCallback<Void>() {
+            @Override
+            public void onFailure(Throwable caught) {
+                ErrorHandler.post(caught);
```
psarando: The `POST /admin/tools` endpoint will return an `ERR_NOT_WRITEABLE` if the admin attempts to update container settings for a "public" tool (tools used in public apps), but we don't want to completely prevent admins from updating a public tool's container settings, so the endpoint provides an `overwrite-public` flag that may be included in the request to update these settings for public tools.

The UI is intended to use this flag in a similar way it's using the `force` flag with the copy metadata endpoint: see the [CopyMetadataCallback](https://github.com/iPlantCollaborativeOpenSource/DE/blob/dev/ui/de-lib/src/main/java/org/iplantc/de/diskResource/client/presenters/grid/GridViewPresenterImpl.java#L125-L171) for an example.

I would also encourage you to use a message similar to the `Danger Zone` message of the `POST /admin/tools` endpoint when presenting the admin an `overwrite-public` confirmation dialog.

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/toolAdmin/view/subviews/ToolVolumesFromListEditor.java
```diff
@@ -0,0 +1,221 @@
...
+    public void addVolumesFrom() {
+        ToolAutoBeanFactory factory = GWT.create(ToolAutoBeanFactory.class);
+        ToolVolumesFrom volumesFrom = factory.getVolumesFrom().as();
+        getVolumesFromTag(volumesFrom);
+
+        volumesFrom.setName("");
+        volumesFrom.setUrl("");
+        volumesFrom.setTag("");
+        volumesFrom.setNamePrefix("");
+        volumesFrom.setReadOnly(true);
```
psarando: For JSON models like this, it's safer to leave values `null` by default. The service may allow a field to be optional but may not allow blank values. So if the user doesn't fill in a value for fields like this, they should be left null so that when the model is serialized to JSON, the field will be omitted. Also, we generally don't want to save blank values in the database, so we should avoid submitting blank values from the UI (the services should validate a field isn't blank, but they don't always).

ui/de-lib/src/main/java/org/iplantc/de/client/models/tool/ToolVolumesFrom.java
```diff
@@ -0,0 +1,39 @@
...
+    void setTag(String tag);
+
+    @PropertyName("tag")
+    String getTag();
+
+    @PropertyName("url")
+    void setUrl(String url);
+
+    @PropertyName("url")
+    String getUrl();
```
psarando: FYI: The `@PropertyName` annotation isn't required when the method name after `get/set` matches the property name. So `@PropertyName("url")` isn't required for `getUrl`, but `@PropertyName("name_prefix")` is required for `getNamePrefix`

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/toolAdmin/view/subviews/ToolVolumesFromListEditor.java
ash3rz: I double checked and it seems that when the editor got flushed, if any of these values were empty strings they were not included in the outgoing JSON string.  Either way, I think this was leftover code from before I started using tags so I don't actually need to set values anymore.  I'll remove them and note to be cautious about setting blank values in the future.

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/toolAdmin/view/subviews/ToolTestDataParamsListEditor.java
```diff
@@ -0,0 +1,117 @@
...
+import com.sencha.gxt.widget.core.client.grid.ColumnModel;
+import com.sencha.gxt.widget.core.client.grid.Grid;
+import com.sencha.gxt.widget.core.client.grid.editing.GridEditing;
+import com.sencha.gxt.widget.core.client.grid.editing.GridInlineEditing;
+
+import java.util.ArrayList;
+import java.util.List;
+
+
+public class ToolTestDataParamsListEditor extends Composite
```
strootman: This class is no longer used. It should be removed.

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/toolAdmin/view/subviews/ToolTestDataEditor.java
```diff
@@ -0,0 +1,87 @@
...
+    interface EditorDriver extends SimpleBeanEditorDriver<ToolTestData, ToolTestDataEditor> {
+    }
+
+    private static ToolTestDataEditorBinder uiBinder = GWT.create(ToolTestDataEditorBinder.class);
+
+    private final EditorDriver editorDriver = GWT.create(EditorDriver.class);
+
+    //    @Ignore
+//    @UiField TextButton addParameterButton;
+//    @UiField ToolTestDataParamsListEditor paramsEditor;
```
strootman: If this class is not needed, these commented out lines should be removed as well. This goes for the lines below as well.

ash3rz: From what I understand, the implementation's params aren't stored anymore in the database, only input_files and output_files.  I thought perhaps leaving this here would be helpful if ever params were added back.  I can see how removing it would be cleaner though.

ui/de-lib/src/main/java/org/iplantc/de/theme/base/client/admin/toolAdmin/ToolAdminDisplayStrings.properties
```diff
@@ -0,0 +1,71 @@
...
+updateToolSuccessText = Tool successfully updated!
+dialogWindowDeleteBtnText = Delete
+deleteToolSuccessText = Tool successfully deleted!
+confirmOverwriteTitle = Public App Overwrite?
+confirmOverwriteDangerZone = Do not update container settings that are in use by tools in public apps unless it is certain the new container settings will not break reproducibility for those apps.\
+  <br>If required, the `overwrite-public` flag may be used to update these settings for public tools.\
+  <br><br> The following apps are currently using this tool:
+confirmOverwriteBody = Do you wish to use the overwrite-public flag now?
+deletePublicToolTitle = Tool in use by Public Apps
+deletePublicToolBody = This tool is already in use by public apps.  Deleting this tool is not an option while public apps are using it.
```
psarando: Actually, tools used in any app (public or private) cannot be deleted.

ui/de-lib/src/main/java/org/iplantc/de/theme/base/client/admin/toolAdmin/ToolAdminDisplayStrings.properties
```diff
@@ -0,0 +1,71 @@
...
+dialogWindowUpdateBtnText = Save/Update
+addToolSuccessText = Tool successfully added!
+updateToolSuccessText = Tool successfully updated!
+dialogWindowDeleteBtnText = Delete
+deleteToolSuccessText = Tool successfully deleted!
+confirmOverwriteTitle = Public App Overwrite?
+confirmOverwriteDangerZone = Do not update container settings that are in use by tools in public apps unless it is certain the new container settings will not break reproducibility for those apps.\
+  <br>If required, the `overwrite-public` flag may be used to update these settings for public tools.\
+  <br><br> The following apps are currently using this tool:
+confirmOverwriteBody = Do you wish to use the overwrite-public flag now?
```
psarando: Our admins don't need to know specifically about the `overwrite-public` flag.
We can simplify these messages for our admin users, and just say something like, "If required, the public tool may still be overwritten" and "Do you wish to overwrite the public tool?"

ash3rz: Ah ok, I wasn't sure if making it as close to the Swagger text was best or not.  I'll make it more general. Thanks!

strootman:
The grids in the `Tool Implementation` section do not have field labels like the grids in the `Container Details` section. I would suggest creating field labels with the same text as the column headers, and change the column headers to `File Name`.

strootman:
The `Container Volumes From` grid has a horizontal scroll bar (trivial). Fix if possible

strootman:
Selecting the row immediately opens the edit window. We should strive to make this consistent with the majority of our app. 

So, selecting the row should not automatically open the edit window (should not result in an immediate action other than selection). This is how the data, apps, and analyses windows operate, as well as the reference genome and system messages tabs. In the windows mentioned, the actionable selection comes from hovering over the name of the item in the grid.

strootman:
You can resize the edit window to an unusable size. Reasonable min width/height settings should be applied.

ui/de-lib/src/main/java/org/iplantc/de/theme/base/client/admin/toolAdmin/ToolAdminDisplayStrings.properties
```diff
@@ -0,0 +1,83 @@
...
+inputFilesHelp = The list of paths to test input files in iRODS
+outputFilesHelp = The list of paths to expected output files in iRODS
+containerDeviceHelp = Device information for the Docker container associated with the tool
+containerVolumeHelp = Volume information for the Docker container associated with the tool. Volumes are designated directories that exist outside the union file system to allow for persistent data and are accessible by the host.
+containerVolumesFromHelp = A list of Docker containers this tool`s container should import volumes from.
+entryPointHelp = The entrypoint defines what should be run when the image is run, and is ideal when a tool consists of a single main program.
+imageTagHelp = A tag is a label applied to a Docker image in a repository. tags are how various images in a repository are distinguished from each other.
+workingDirectoryHelp = The working directory or path for any instructions within the dockerfile.
```
psarando: I haven't seen the `Warning` messages from the Swagger docs for the `POST /admin/tools` endpoint included in the UI yet.

I think the `Use caution with Entrypoint settings` and `Use caution with Volumes settings` should be displayed somewhere appropriate, but obvious, in the UI where these settings can be changed.

ash3rz: Good catch - I added these warnings to the UI in the latest commit.

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/views/BelphegorView.ui.xml
```diff
@@ -48,6 +48,10 @@
              type="com.sencha.gxt.widget.core.client.TabItemConfig">
         <ui:attributes text="Metadata Templates"/>
     </ui:with>
+    <ui:with field="toolAdminConfig"
+             type="com.sencha.gxt.widget.core.client.TabItemConfig">
+        <ui:attributes text="Tool Admin"/>
+    </ui:with>
```
psarando: I suggest putting this tab next to the "Tool Requests" tab, rather than just at the end.

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/toolAdmin/presenter/ToolAdminPresenterImpl.java
```diff
@@ -0,0 +1,221 @@
...
+                updateView();
+            }
+        });
+    }
+
+    String getServiceError(Throwable caught) {
+        SimpleServiceError simpleServiceError =  AutoBeanCodex.decode(factory, SimpleServiceError.class, caught.getMessage())
+                     .as();
+        return simpleServiceError.getErrorCode();
+    }
```
psarando: Not all errors will return a JSON body, so attempting to parse JSON here may cause an exception in the UI.

For example, I didn't update my `de.properties` with the tool admin endpoint, so when attempting to save a tool, the server-side returned a `500 Internal Server Error The call failed on the server; see server log for details` string, which caused this method to throw a SyntaxError exception, but in the UI the form closed and no error was displayed to the user.

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/views/BelphegorView.ui.xml
ash3rz: :+1:  done

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/toolAdmin/view/subviews/ToolVolumesFromListEditor.java
```diff
@@ -0,0 +1,183 @@
...
+        columns.add(name);
+        columns.add(namePrefix);
+        columns.add(tag);
+        columns.add(url);
+        columns.add(readOnly);
+        ColumnModel<ToolVolumesFrom> cm = new ColumnModel<>(columns);
+
+        grid = new Grid<>(listStore, cm);
+
+        editing = new GridInlineEditing<>(grid);
```
psarando: `GridRowEditing` could be more user friendly here. It should probably also be used for the other editable grids in this form.

It may be questionable to use it for the Test Input and Output file grids since they only have 1 column each, but I think the row-editor will capture the `esc` key and simply stop editing the row. With the inline-editor, hitting escape while editing a cell caused the entire form to close and all work in-progress was lost.

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/toolAdmin/view/dialogs/ToolAdminDetailsDialog.java
```diff
@@ -0,0 +1,191 @@
...
+        setResizable(true);
+        setPixelSize(1000, 500);
+        setMinHeight(200);
+        setMinWidth(500);
+
+        addHelp(new HTML(appearance.toolAdminHelp()));
+
+        setPredefinedButtons(PredefinedButton.OK, PredefinedButton.CANCEL, PredefinedButton.CLOSE);
+        getOkButton().setText(appearance.dialogWindowUpdateBtnText());
+        getDeleteButton().setText(appearance.dialogWindowDeleteBtnText());
```
psarando: The `Delete` button should probably go next to the `Add` button in the main `Tool Admin` toolbar, rather than in the edit dialog. This would also be more consistent with the other tabs like `System Messages` and `Metadata Templates`.

ash3rz: I had thought about that, but wasn't sure if it's "dangerous" having the delete button on the main view.  Having the delete button within the edit dialog makes it very clear what you intend to delete.  But I agree, it's inconsistent.  If we'd rather go for consistency, I'll bump the button out to the main view.

psarando: It should be fine as long as the `Delete` button displays a confirmation dialog before performing the service call (which it already does).

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/toolAdmin/view/subviews/ToolVolumesFromListEditor.java
ash3rz: Done - I left the Input and Output file grids alone for now

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/toolAdmin/view/dialogs/ToolAdminDetailsDialog.java
ash3rz: Makes sense - this has been updated.

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/toolAdmin/view/subviews/ToolContainerEditor.java
```diff
@@ -0,0 +1,129 @@
...
+
+    private static ToolContainerEditorBinder uiBinder = GWT.create(ToolContainerEditorBinder.class);
+
+    @UiField TextField nameEditor;
+    @UiField TextField workingDirectoryEditor;
+    @Ignore
+    @UiField HTML entryPointWarningHTML;
+    @UiField TextField entryPointEditor;
+    @UiField IntegerField memoryLimitEditor;
+    @UiField IntegerField cpuSharesEditor;
```
psarando: These memory and CPU fields are optional in the service endpoint's request, so the user should be able to delete the values in these fields, and when the UI submits the request, these fields should be omitted from the JSON request. If the value is null in the UI model, this should allow the JSON field to be omitted in the request, but when I removed the values from these fields and saved the form, the UI threw an exception for these null values.

Some of the other optional Text fields should also be omitted in the request if the user removes the value in the form, and should not be submitted to the service with empty strings. This includes the image Tag and URL fields, and the container Name, Working Dir, Entry Point, and Network Mode fields.

ash3rz: I believe this should be fixed now.  This was actually helpful because I didn't even realize that Input Files and Output Files are required - I was sending empty arrays previously.  So fixing this caught that error.

psarando: :+1: 
Yep, the Input and Output arrays are required, but those arrays are allowed to be empty.
Though the Input and Output arrays shouldn't be submitted with empty strings inside them.

ash3rz: Oh ok, my original understanding on those arrays was correct then.  Looking at the swagger docs, is there any indication that while the array is required it could be an empty array?

ash3rz: Either way, I think my latest push should work as expected - empty strings are removed and empty arrays are removed except for input_files and output_files since they're required to at least be present.

psarando: The Swagger docs only tell you that they're required (`input_files (Array[string]):` instead of `input_files (Array[string], optional):` in the `Model` docs). I don't think there's an automated way for the Swagger docs to display that blank-strings aren't allowed in the arrays, so the English docs on those fields should mention if they aren't allowed.

Currently these endpoints aren't checking if the arrays contain blank strings, so they will just add empty strings to the database :-1: They may be updated to return errors if blank-strings are submitted by the client, so either way the UI shouldn't submit blank-strings if possible.

ianmcorvidae: We do have a `NonBlankString` schema which we use in some places, which could presumably be used here.

ianmcorvidae: (this is of course in services so is getting a bit in the weeds as far as the pull request in question. but anyway)

psarando: Also, the same goes for empty vs. non-empty arrays. I don't think Swagger can automatically display if empty arrays aren't allowed, so the English docs should mention if an array is required to contain elements. Otherwise, the array is allowed to be empty by default.

ash3rz: OK that answered my next question about whether it's possible to require an array that has at least one element.  Thanks!

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/toolAdmin/service/impl/ToolAdminServiceFacadeImpl.java
```diff
@@ -0,0 +1,111 @@
...
+            if ((temp.equals("") || temp.equals("\"\"")) && !key.matches("(input|output)_files")) {
+                StringQuoter.nullValue().assign(obj, key);
+            } else {
+                if (value.isIndexed() || value.isKeyed()) {
+                    removeEmptyValuesHelper(value);
+                }
+            }
+        }
+    }
+}
```
psarando: This seems like the hard-way to remove empty strings from the request (and potentially error-prone?).
I thought there might be a simpler way to set `NULL` values for specific fields in the AutoBean before it's encoded by the service facade, maybe with ValueChange handlers in the fields or the editor itself?
I thought we had done this already elsewhere, but I'm not finding an example outside of [TreeSelectionEditor.java](https://github.com/cyverse/DE/blob/tool-admin/ui/de-lib/src/main/java/org/iplantc/de/apps/widgets/client/view/editors/arguments/tree/TreeSelectionEditor.java) or [ArgumentEditorConverter.java](https://github.com/cyverse/DE/blob/tool-admin/ui/de-lib/src/main/java/org/iplantc/de/apps/widgets/client/view/editors/arguments/converters/ArgumentEditorConverter.java) right now.

psarando: A `ValueChangeHandler` could set a null value instead of an empty string for a `TextField`. For example:
```java
        workingDirectoryEditor.addValueChangeHandler(new ValueChangeHandler<String>() {
            @Override
            public void onValueChange(ValueChangeEvent<String> valueChangeEvent) {
                final String value = valueChangeEvent.getValue();
                if (value != null && value.isEmpty()) {
                    workingDirectoryEditor.setValue(null);
                }
            }
        });
```
This `ValueChangeHandler` could be extracted into its own class and reused in every `TextField` we don't want submitting empty-strings.

I'm not sure about the list editors, but the service endpoints shouldn't store empty values for the `container_devices`, `container_volumes`, and `container_volumes_from` arrays if they are just empty arrays, as long as we don't submit them with empty strings in their elements. So we don't have to null-out these arrays for now.

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/toolAdmin/view/subviews/ToolContainerEditor.java
psarando: Just to reiterate, the Input and Output File arrays are not required to contain any elements.
They are required in the request, but they may be empty arrays.
We just don't want them submitting empty string elements in the arrays.

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/toolAdmin/service/impl/ToolAdminServiceFacadeImpl.java
ash3rz: Cool, I'll start looking into this option - thanks!

psarando: :+1:

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/toolAdmin/view/subviews/ToolContainerEditor.java
ash3rz: Understood.  I was just surprised that it's possible to require an array but then also not require any elements within that same array.

psarando: Yeah, it's a little odd, and maybe those JSON keys should actually be made optional in the endpoint, since it doesn't exactly make sense to require an empty array instead of just allowing that key to be omitted.

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/toolAdmin/service/impl/ToolAdminServiceFacadeImpl.java
ash3rz: Does this also apply to the outermost fields in Tool?  It seems if I set a tool's `attribution`, for example, to some value, then update it to null, it gets omitted from the outgoing JSON body.  Then when I pull up that same tool, the old value for `attribution` is still there.

psarando: This is generally the way our `PATCH` endpoints work. If the value is not submitted in the `PATCH` request, then it is not updated. Most of the `container` subfields are different, where if the `container` subfield is not submitted as part of the `container`, it gets deleted.

From this `PATCH` endpoint's docs:
> Note: If the `container` object is omitted in the request, then existing container settings will not be modified, but if the `container` object is present in the request, then all container settings must be included in it. Any existing settings not included in the request's `container` object will be removed.

In all this discussion about not submitting blank strings, this point should have been clarified. Since the `container` object works this way, those fields should not be submitted with blank strings, but for other top-level fields like `attribution` or `version`, there's not another way to remove the value with this endpoint, so those fields can be submitted with empty strings in order to "delete" those values.

ash3rz: OK pushed another change.  I think only the `container` fields have to be checked, so I added the ValueChangeHandler there.  
Then for `container_devices`, `container_volumes`, and `container_volumes_from`, if you add a row with empty values, the form won't pass the `isValid()` method which means you'll get an error pop-up if you try to submit the form.  If there's a better way to handle this (like a way to keep the row within the grid red, which I haven't found unless the row is active), I'm open to ideas.

strootman:
Is this ready to merge?

ash3rz:
I sure hope so! :smile:

psarando:
Not essential for this PR, but I would split-up commits like that last `Catch JSON parse errors/remove empty values` commit. It seems like there are 2 or 3 unrelated changes that could each be their own commit: JSON parse errors vs. empty-strings vs. delete button change? (that last one isn't mentioned in the commit message).

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/toolAdmin/view/subviews/ToolContainerEditor.java
psarando: The `ToolImageEditor` is still allowing empty strings to submit for the image `tag` and `url`.

ash3rz:
I think I squashed a few changes into the wrong commit.  I'll clean this up.

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/toolAdmin/view/subviews/ToolContainerEditor.java
ash3rz: Should be fixed now.

psarando:
:+1:

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/toolAdmin/view/subviews/ToolContainerEditor.java
psarando: Almost there! Now when editing an existing `Volumes From` row, if the `tag` and `url` fields are deleted by the user, those fields are submitted as empty-strings in the JSON request.

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/toolAdmin/view/subviews/ToolDeviceListEditor.java
```diff
@@ -0,0 +1,139 @@
...
+        columns.add(hostPath);
+        columns.add(containerPath);
+        ColumnModel<ToolDevice> cm = new ColumnModel<>(columns);
+
+        grid = new Grid<>(listStore, cm);
+        grid.setHeight(100);
+
+        editing = new GridRowEditing<>(grid);
+        editing.addEditor(hostPath, new TextField());
+        editing.addEditor(containerPath, new TextField());
```
psarando: These TextFields should also `setAllowBlank(false)` similar to the `ToolVolumeListEditor`, since both these fields are required per device setting.

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/toolAdmin/view/subviews/ToolContainerEditor.java
ash3rz: Good catch - should be fixed.

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/toolAdmin/view/subviews/ToolDeviceListEditor.java
ash3rz: Not sure how I put `setAllowBlank` in one but not the other, but should be fixed now.

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/toolAdmin/view/subviews/ToolVolumesFromListEditor.java
```diff
@@ -0,0 +1,204 @@
...
+        }
+
+        @Override
+        public void onValueChange(ValueChangeEvent<String> event) {
+            String value = event.getValue();
+            if (value != null && value.isEmpty()) {
+                field.setValue(null);
+            }
+        }
+    }
```
psarando: It would be useful to refactor these 3 duplicated private classes into a public class that could be used elsewhere (in metadata templates, for example). Maybe under `org.iplantc.de.commons.client.widgets`?

psarando: If it's refactored as a public class, then it would also be a good idea to check for a null `field` before attempting to set its value.

ash3rz: Makes sense.  Would a name like `EmptyStringValueChangeHandler` make sense?

psarando: :+1: sounds good to me.

ash3rz: OK, this should be pushed.

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/toolAdmin/view/dialogs/DeleteAppDialog.java
```diff
@@ -0,0 +1,58 @@
...
+ */
+public class DeleteAppDialog extends IPlantDialog implements IsHideable {
+
+    private AppAutoBeanFactory appAutoBeanFactory = GWT.create(AppAutoBeanFactory.class);
+    private FlowLayoutContainer container;
+    @Inject ToolPublicAppListWindow publicAppListWindow;
+    @Inject ToolAdminView.ToolAdminViewAppearance appearance;
+
+    @Inject
+    public DeleteAppDialog(ToolAdminView.ToolAdminViewAppearance appearance, ToolPublicAppListWindow publicAppListWindow) {
```
psarando: This dialog should be named `DeleteToolDialog`.
I also doesn't need both OK and Cancel buttons, since it doesn't execute any actions. It could just have an OK button.

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/toolAdmin/view/dialogs/OverwriteAppDialog.java
```diff
@@ -0,0 +1,61 @@
...
+import com.google.inject.Inject;
+import com.google.web.bindery.autobean.shared.AutoBeanCodex;
+
+import com.sencha.gxt.core.client.dom.ScrollSupport;
+import com.sencha.gxt.widget.core.client.container.FlowLayoutContainer;
+
+/**
+ * @author aramsey
+ */
+public class OverwriteAppDialog extends IPlantDialog implements IsHideable {
```
psarando: This dialog should be named `OverwriteToolDialog`, since it's a warning about overwriting a tool and not an app.

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/toolAdmin/view/dialogs/DeleteAppDialog.java
ash3rz: OK should be fixed.

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/toolAdmin/view/ToolAdminViewImpl.java
```diff
@@ -0,0 +1,242 @@
...
+
+    }
+
+    private final class NameFilter implements Store.StoreFilter<Tool> {
+
+        private String filterText;
+
+        @Override
+        public boolean select(Store<Tool> store, Tool parent, Tool item) {
+            return !Strings.nullToEmpty(filterText).isEmpty() && item.getName()
```
psarando: FYI: `Strings.nullToEmpty(filterText).isEmpty()` can be reduced to `Strings.isNullOrEmpty(filterText)`.

psarando:
Looks good, nice work! :+1:

PR #12, opened 2015-12-07
7344 ansible restructuring

strootman:
This pr contains all the role and playbook renaming which segregates roles/playbooks into one of 4 tiers;
* util
* infrastructure
* support
* de (application)

CORE-7350
CORE-7345


ansible/host_vars/.gitignore
```diff
@@ -1 +0,0 @@
-localhost
```
psarando: Why remove this file?
Especially since our docs still reference it: http://iplantcollaborativeopensource.github.io/DE/ansible/work_instructions/local_configs.html (step 3).

ansible/deploy-all.yaml
```diff
@@ -35,31 +35,31 @@
...
-- include: playbooks/rsyslog-config.yaml
-- include: playbooks/10-logstash-forwarder.yaml
+- include: playbooks/support-rsyslog-config.yaml
+- include: playbooks/support-logstash-forwarder.yaml
 
 ############################################
 # Systems
 ############################################
-- include: systems.yaml
+- include: infra-systems.yaml
```
psarando: These `infra-*` and `support-*` playbooks were moved to our internal repo, so I don't think they can be called from this playbook anymore.

ansible/inventories/group_vars/all
```diff
@@ -32,11 +32,6 @@ logstash_forwarder:
 # environment
 parasitic: false
 
-denyhosts_allowed:
```
psarando: This var is still lurking in our internal repo, even though the role that uses it was removed completely.
The `group_vars/all` files of both repos should be cleaned up according to the changes of this pull-request, and if possible we should look into re-organizing the `group_vars/all` files so that as many internal playbooks as possible can be run solely from the internal repo.

psarando: I may have jumped the gun with this comment.
I see that even though this pull request is closed, the ticket for it is still in progress :construction: :construction_worker:

ansible/host_vars/.gitignore
strootman: My mistake. I'll put it back.

PR #13, opened 2015-12-08
Removed temporary ansible roles and playbooks

strootman:
Resolves: CORE-7259, CORE-7274, CORE-7289, CORE-7319


ianmcorvidae:
:shipit: looks good to me

psarando:
The `TMP-services-cleanup.yaml` include needs to be removed from `ansible/deploy-all.yaml`.

psarando:
Looks good now.

PR #14, opened 2015-12-08
Removed the iplant user from all Cfg Mgmt

strootman:


docker/javabase/Dockerfile
```diff
@@ -12,10 +12,10 @@ RUN apk add --update curl \
...
-RUN busybox adduser -s /bin/sh -u 1337 -D -h /home/iplant iplant
+#RUN mkdir -p /home/iplant
+#RUN busybox adduser -s /bin/sh -u 1337 -D -h /home/iplant iplant
 
-USER iplant
-RUN mkdir -p /home/iplant/logs/
-WORKDIR /home/iplant
+#USER iplant
+#RUN mkdir -p /home/iplant/logs/
+#WORKDIR /home/iplant
```
ianmcorvidae: Delete these lines for real; we can get them from history if we need them back.

services/NotificationAgent/Dockerfile
```diff
@@ -14,6 +14,5 @@ LABEL org.iplantc.de.notificationagent.git-ref="$git_commit" \
       org.iplantc.de.notificationagent.version="$version" \
       org.iplantc.de.buildenv.git-ref="$buildenv_git_commit"
 
-USER iplant
 ENTRYPOINT ["java", "-Dlogback.configurationFile=/etc/iplant/de/logging/notificationagent-logging.xml", "-cp", ".:notificationagent-standalone.jar", "notification_agent.core"]
```
ianmcorvidae: This line is depending on things being copied into `/home/iplant` and the workdir being `/home/iplant` -- probably those things will need to be changed to just use `/root` or `/`

ianmcorvidae: (this of course applies to basically all the service files)

ansible/playbooks/grouper.yaml
```diff
@@ -0,0 +1,16 @@
...
+        username: Grouper
+        icon: ":fish:"
+        msg: "Deploying Grouper"
+
+      - role: grouper
+
+      - role: notify-chat
+        username: Grouper
+        icon: ":fish:"
+        msg: ":heavy_check_mark: Done deploying Grouper"
```
psarando: Wasn't this just removed by https://github.com/iPlantCollaborativeOpenSource/DE/pull/12?

ianmcorvidae:
It won't let me comment on the exact lines but you should remove all the things chowning things to the iplant user when you change where files are copied and run from, etc.

ansible/roles/de-services-common/tasks/main.yaml
```diff
@@ -1,27 +1,11 @@
...
-  user: name=iplant shell=/bin/bash groups=iplant append=yes
-  tags:
-      - init_users
-      - init_iplant_user
-
-- name: add the docker group to iplant user
+- name: add the current user to the docker group
   sudo: yes
-  user: name=iplant groups=docker append=yes
+  user: name={{ ansible_ssh_user }} groups=docker append=yes
```
ianmcorvidae: does this make sense? if I'm reading this right it'll add the person running ansible to the docker group, which should either be something they manage themselves or something done automatically by LDAP or such; it certainly doesn't affect our deployment process

services/condor-log-monitor/Dockerfile
```diff
@@ -9,9 +9,6 @@ LABEL org.iplantc.de.condor-log-monitor.git-ref="$git_commit" \
 
 RUN useradd -m -U -s /bin/bash iplant
```
ianmcorvidae: This is part of the same process. But I also think clm might just be in the jex-events image now, so... not sure if it matters? @johnworth would know I guess

services/NotificationAgent/Dockerfile
strootman: Yes, the ansible tasks have been updated to apply owner/group as `root`

ansible/playbooks/grouper.yaml
strootman: Yes. will push a change

ansible/roles/de-services-common/tasks/main.yaml
strootman: You are reading it right, and boy is it convenient!!
And no, it isn't directly needed by the deployment. I'll remove it.

strootman:
@ianmcorvidae I just pushed a changed that removed the chowns

ansible/roles/grouper/templates/iplant-grouper.service
```diff
@@ -0,0 +1,29 @@
...
+ExecStop=/usr/bin/docker stop {{grouper.container_name}}
+Restart=on-failure
+StartLimitInterval=60s
+StartLimitBurst=3
+
+SyslogIdentifier={{grouper.container_name}}
+
+[Install]
+WantedBy={{systemd.target}}
+
```
psarando: These elk and grouper files, removed in https://github.com/iPlantCollaborativeOpenSource/DE/pull/12, are also getting re-added in this pull request.

strootman: Files have been removed. Rebasing onto dev helped identify what was needed to be removed.

services/NotificationAgent/Dockerfile
```diff
@@ -5,7 +5,6 @@ VOLUME ["/etc/iplant/de"]
 
 COPY conf/main/logback.xml /home/iplant/
 COPY target/notificationagent-standalone.jar /home/iplant/
```
ianmcorvidae: The entrypoint below, in conjunction with these lines' copying locations, still depends on the `WORKDIR /home/iplant` removed from `javabase` (it adds `.` to the classpath, which is the `/home/iplant` here by way of the `WORKDIR`). They should be copied into whatever directory is the default workdir (`/`, I believe, for `jeanblanchard/java:8` which is the base of `javabase`) or another alternately configured working directory such as `/root`.

services/jex-events/Dockerfile
```diff
@@ -10,9 +10,6 @@ LABEL org.iplantc.de.jex-events.git-ref="$git_commit" \
       org.iplantc.de.jex-events.version="$version" \
       org.iplantc.de.buildenv.git-ref="$buildenv_git_commit"
 
-USER iplant
```
ianmcorvidae: There is a `useradd` command further up in this dockerfile which should be removed.

strootman:
I've removed the `useradd`s, as well as updated the paths for files which are being copied in.

@ianmcorvidae If you find something visually, excellent, but I'm going to start testing these changes out. From building the images to deploying to an environment. I'll keep this PR updated with progress.

ianmcorvidae:
I don't see anything else, checking quickly.

PR #15, opened 2015-12-10
Update compojure-api to latest (0.24.2)

ianmcorvidae:


libs/iplant-clojure-commons/project.clj
```diff
@@ -16,7 +16,6 @@
                   :exclusions [commons-logging]]
                  [me.raynes/fs "1.4.6"]
                  [medley "0.7.0"]
-                 [metosin/compojure-api "0.23.1"]
```
ianmcorvidae: This was in this project.clj twice. So I removed this one.

psarando:
Tested this locally with a few `apps` endpoints and `data-info` endpoints.
Looks good to me.

PR #19, opened 2015-12-17
CORE-7370: modify all JVM services to have descriptive process names

dennisroberts71:
This turned out to be a lot simpler than I expected because the JDK version of the `java` executable (as opposed to the JRE version) doesn't re-execute the command. Because of this, we can use a symbolic link to change the process name of all of our JVM services that run in Docker containers.

The JEX was a little bit different because it doesn't run in a container. This was also relatively simple however. This change involved creating a bash wrapper script that uses `exec -a` to set the process name.

### DIFF
--- a/ansible/roles/de-jex/tasks/main.yaml
+++ b/ansible/roles/de-jex/tasks/main.yaml
@@ -46,10 +46,15 @@
     - jex
     - docker_pull
 
+- name: create /usr/local/bin/jex
+  sudo: yes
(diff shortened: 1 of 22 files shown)


tedgin:
Does this mean that the entire JDK needs to be in each of the docker images?

dennisroberts71:
It does mean that a pared-down version of the JDK will be in in each Docker image, but that was the case before also.

psarando:
Looks good!

PR #21, opened 2016-01-12
iRODS AVU endpoint migration

ianmcorvidae:
I'll include a variety of comments below, interleaved. In general, this is about removing the jargon-related calls from terrain.services.filesystem.metadata, but a variety of other things were added in order to allow terrain to continue coordinating things, since chunks of that code interact with the metadata service as well.

### DIFF
--- a/services/data-info/project.clj
+++ b/services/data-info/project.clj
@@ -54,4 +54,5 @@
          :init data-info.core/lein-ring-init
          :port 60000
          :auto-reload? false}
-  :uberjar-exclusions [#".*[.]SF" #"LICENSE" #"NOTICE"])
+  :uberjar-exclusions [#".*[.]SF" #"LICENSE" #"NOTICE"]
+  :jvm-opts ["-Dlogback.configurationFile=/etc/iplant/de/logging/data-info-logging.xml"])
--- a/services/data-info/project.clj
+++ b/services/data-info/project.clj
@@ -23,7 +23,6 @@
                  [com.cemerick/url "0.1.1"]
                  [dire "0.5.3"]
                  [me.raynes/fs "1.4.6"]
-                 [medley "0.7.0"]
                  [metosin/compojure-api "0.24.2"]
                  [net.sf.json-lib/json-lib "2.4" :classifier "jdk15"]
                  [org.apache.tika/tika-core "1.11"]
--- a/services/data-info/src/data_info/routes/domain/common.clj
+++ b/services/data-info/src/data_info/routes/domain/common.clj
@@ -11,11 +11,19 @@
 (def DataIdPathParam (describe UUID "The data item's UUID"))
 
 (s/defschema Paths
-  {:paths (describe [NonBlankString] "A list of IRODS paths")})
+  {:paths (describe [NonBlankString] "A list of iRODS paths")})
 
 (s/defschema OptionalPaths
   (-> Paths
     (->optional-param :paths)))
+
+(s/defschema DataIds
+  {:ids (describe [UUID] "A list of iRODS data-object UUIDs")})
+
+(s/defschema OptionalPathsOrDataIds
+  (-> (merge DataIds Paths)
+      (->optional-param :paths)
+      (->optional-param :ids)))
 
 (def ValidInfoTypesEnum (apply s/enum (hm/supported-formats)))
 (def ValidInfoTypesEnumPlusBlank (apply s/enum (conj (hm/supported-formats) "")))
--- a/services/data-info/src/data_info/routes/domain/stats.clj
+++ b/services/data-info/src/data_info/routes/domain/stats.clj
@@ -62,8 +62,13 @@
   {(describe s/Keyword "The iRODS data item's path")
    (describe (s/conditional #(contains? % :file-size) FileStatInfo :else DirStatInfo) "The data item's info")})
 
+(s/defschema DataIdsMap
+  {(describe s/Keyword "The iRODS data item's ID")
+   (describe (s/conditional #(contains? % :file-size) FileStatInfo :else DirStatInfo) "The data item's info")})
+
 (s/defschema StatusInfo
-  {:paths (describe PathsMap "Paths info")})
+  {(s/optional-key :paths) (describe PathsMap "Paths info")
+   (s/optional-key :ids) (describe DataIdsMap "IDs info")})
 
 ;; Used only for display as documentation in Swagger UI
 (s/defschema StatResponsePathsMap
@@ -71,5 +76,11 @@
    :/path/from/request/to/a/file   (describe FileStatInfo "A file's info")})
 
 ;; Used only for display as documentation in Swagger UI
+(s/defschema StatResponseIdsMap
+  {:some-folder-uuid (describe DirStatInfo "A folder's info")
+   :some-file-uuid   (describe FileStatInfo "A file's info")})
+
+;; Used only for display as documentation in Swagger UI
 (s/defschema StatResponse
-  {:paths (describe StatResponsePathsMap "A map of paths from the request to their status info")})
+  {(s/optional-key :paths) (describe StatResponsePathsMap "A map of paths from the request to their status info")
+   (s/optional-key :ids) (describe StatResponseIdsMap "A map of ids from the request to their status info")})
--- a/services/data-info/src/data_info/routes/domain/stats.clj
+++ b/services/data-info/src/data_info/routes/domain/stats.clj
@@ -1,5 +1,5 @@
 (ns data-info.routes.domain.stats
-  (:use [common-swagger-api.schema :only [describe NonBlankString]]
+  (:use [common-swagger-api.schema :only [describe NonBlankString StandardUserQueryParams]]
         [data-info.routes.domain.common])
   (:require [schema.core :as s])
   (:import [java.util UUID]))
@@ -8,6 +8,11 @@
 (def PermissionEnum (s/enum :read :write :own))
 (def DataItemIdParam (describe UUID "The UUID of this data item"))
 (def DataItemPathParam (describe NonBlankString "The IRODS paths to this data item"))
+
+(s/defschema StatQueryParams
+  (assoc StandardUserQueryParams
+         (s/optional-key :validation-behavior)
+         (describe PermissionEnum "What level of permissions on the queried files should be validated?")))
 
 (s/defschema DataStatInfo
   {:id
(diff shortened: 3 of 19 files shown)


services/data-info/project.clj
```diff
@@ -54,4 +54,5 @@
          :init data-info.core/lein-ring-init
          :port 60000
          :auto-reload? false}
-  :uberjar-exclusions [#".*[.]SF" #"LICENSE" #"NOTICE"])
+  :uberjar-exclusions [#".*[.]SF" #"LICENSE" #"NOTICE"]
+  :jvm-opts ["-Dlogback.configurationFile=/etc/iplant/de/logging/data-info-logging.xml"])
```
ianmcorvidae: Unrelated; I wanted to test logging and we have these lines for the rest of our services.

services/data-info/src/data_info/routes/domain/common.clj
```diff
@@ -11,11 +11,19 @@
...
   (-> Paths
     (->optional-param :paths)))
 
+(s/defschema DataIds
+  {:ids (describe [UUID] "A list of iRODS data-object UUIDs")})
+
+(s/defschema OptionalPathsOrDataIds
+  (-> (merge DataIds Paths)
+      (->optional-param :paths)
+      (->optional-param :ids)))
```
ianmcorvidae: This is the first of several changes which enable the stat-gatherer endpoint to also provide information when given data-object UUIDs, not just paths.

services/data-info/src/data_info/routes/domain/stats.clj
```diff
@@ -9,6 +9,11 @@
 (def DataItemIdParam (describe UUID "The UUID of this data item"))
 (def DataItemPathParam (describe NonBlankString "The IRODS paths to this data item"))
 
+(s/defschema StatQueryParams
+  (assoc StandardUserQueryParams
+         (s/optional-key :validation-behavior)
+         (describe PermissionEnum "What level of permissions on the queried files should be validated?")))
```
ianmcorvidae: This is the start of several changes which allow the stat-gatherer endpoint to have a configurable level of file validation: just readability (as it always has), writeability, or ownership. This enabled removing some of the validators which depended on an iRODS connection.

services/data-info/src/data_info/services/metadata.clj
```diff
@@ -36,13 +37,170 @@
...
+(defn metadata-get
+  "Returns the metadata for a path. Filters out system AVUs
+   if :system true is not passed to it, and replaces
+   units set to ipc-reserved with an empty string."
+  [user data-id & {:keys [system] :or {system false}}]
+  (with-jargon (cfg/jargon-cfg) [cm]
+    (validators/user-exists cm user)
+    (let [path (get-readable-path cm user data-id)]
+      {:irods-avus (list-path-metadata cm path :system system)
+       :path path})))
```
ianmcorvidae: Most of this file is copied from the terrain version. Note however that this endpoint returns the path, since upstream endpoints wanted to return it to the UI.

services/data-info/src/data_info/services/metadata.clj
```diff
@@ -36,13 +37,170 @@
...
+      (authorized-avus irods-avus)
+
+      (doseq [del-avu delete-irods-avus]
+        (common-metadata-delete cm path del-avu))
+      (doseq [avu irods-avus]
+        (common-metadata-add cm path avu))
+
+      {:path path
+       :user user
+       :type type})))
```
ianmcorvidae: This endpoint returns the file type, so that terrain has that information in order to make a request to the metadata service.

services/data-info/src/data_info/services/page_tabular.clj
```diff
@@ -27,12 +27,15 @@
 
 (defn- seek-prev-line
   [chunk start-pos]
+  (log/debug "seeking for a line-ending backwards starting at" start-pos)
```
ianmcorvidae: The changes in this file are primarily just adding logging lines such as this. There's some lurking but very unlikely to be triggered bugs in our CSV parsing which this may be relevant for in the future.

(specifically: if the chunk size provided is small enough, and/or the line of a tabular data file long enough, that the chunk includes only one linebreak, the endpoint will return the partial line that linebreak ends as though it is a complete line. it's unlikely to come up in practice because our real-world code paths have a minimum chunk size of 8KB)

services/data-info/src/data_info/services/stat.clj
```diff
@@ -88,21 +89,31 @@
...
-    (validators/all-paths-readable cm user paths)
-    {:paths (into {} (map (juxt keyword (partial path-stat cm user)) paths))}))
+    (let [uuid-paths (map (juxt (comp keyword str) (partial uuid/get-path cm)) uuids)
+          all-paths (into paths (map second uuid-paths))]
+      (validators/all-paths-exist cm all-paths)
+      (case (keyword validation)
+            :own (validators/user-owns-paths cm user all-paths)
+            :write (validators/all-paths-writeable cm user all-paths)
+            :read (validators/all-paths-readable cm user all-paths)
+            (validators/all-paths-readable cm user all-paths))
```
ianmcorvidae: These five lines are the implementation of the configurable validation. It's making the assumption that if you can own you can write and if you can write you can read, which should be uncontroversial.

services/data-info/src/data_info/services/stat.clj
```diff
@@ -88,21 +89,31 @@
...
+      {:paths (into {} (map (juxt keyword (partial path-stat cm user)) paths))
+       :ids (into {} (map (juxt first #(path-stat cm user (second %))) uuid-paths))})))
```
ianmcorvidae: The other changes in this function are all for supporting ids being passed alongside paths.

services/terrain/src/terrain/services/filesystem/metadata.clj
```diff
@@ -377,44 +200,59 @@
...
-    paths value-lists)))
+      (validate-batch-add-attrs user (map #(get-in path-info-map [% "id"]) paths) irods-attrs))
+  (mapv (partial bulk-add-file-avus user dest-dir template-id template-attrs attrs)
+    path-info-map value-lists)))
+
+(defn- keyword-to-int [kw] (Integer/parseInt (name kw)))
+
+(defn- deparse-csv-line
+  [line]
+  (mapv second (into (sorted-map-by #(< (keyword-to-int %1) (keyword-to-int %2))) line)))
```
ianmcorvidae: This function basically inverts the code in data-info which formats the CSV into something nice for the UI, and turns it back into what opencsv would have returned originally. That is, from a vector of maps from column numbers to column contents, it turns them back into ordered vectors of column contents without explicit column numbers.

services/terrain/src/terrain/services/filesystem/metadata.clj
```diff
@@ -377,44 +200,59 @@
...
+                           ;; It could probably be larger without issue, but this is the expected largest
+                           ;; value for the underlying endpoint in practice. Ideally, it'll never matter
+                           ;; and everything will fit in one page.
+        get-page  (fn [page] (service-response->json (data-raw/read-tabular-chunk user path-uuid separator page chunk-size)))]
+    (loop [page 1 max-pages nil csv []]
+      (let [{max-pages :number-pages new-csv :csv :as res} (get-page page)
+            csv (concat csv (map deparse-csv-line (remove empty? new-csv)))]
+        (if (< page (Integer/parseInt max-pages))
+            (recur (+ page 1) (Integer/parseInt max-pages) csv)
+            csv)))))
```
ianmcorvidae: The loop logic won't come up unless the metadata CSV file is bigger than a megabyte, which should hopefully be unlikely.

services/terrain/src/terrain/util/config.clj
```diff
@@ -502,6 +502,8 @@
        (cfg/env-setting "TREE_URLS_PORT")
        (tree-urls-base-url)))))
 
+(defn tree-urls-attr [] "ipc-tree-urls")
```
ianmcorvidae: I changed this to be configurable as well as changed it to be an administrative AVU (starting with ipc-). This value serves only as a cache and wasn't even being properly fetched before, so this has no downsides other than potentially leaving around some unused 'tree-urls' (the previous AVU name) AVUS.

services/data-info/src/data_info/services/stat.clj
```diff
@@ -3,6 +3,7 @@
             [clojure.tools.logging :as log]
             [dire.core :refer [with-pre-hook! with-post-hook!]]
             [slingshot.slingshot :refer [throw+]]
+            [medley.core :only [remove-vals]]
```
dennisroberts71: This is trivial, but I didn't see `remove-vals` called in this file.

ianmcorvidae: Good point. It looks like this was the only thing using medley in data-info at all, actually, so I've also removed it from project.clj

services/terrain/src/terrain/services/filesystem/metadata.clj
```diff
@@ -143,36 +42,15 @@
...
+    (metadata-client/set-metadata-template-avus data-id data-type (or metadata {}))
+    (select-keys modification-data [:user :path])))
 
 (defn- find-attributes
-  [cm attrs path]
-  (let [matching-avus (get-attributes cm attrs path)]
+  [attrs user uuid]
+  (let [{:keys [irods-avus path]} (service-response->json (data-raw/get-avus user uuid))
+        matching-avus (filter #(contains? attrs (:attr %)) irods-avus)]
     (if-not (empty? matching-avus)
```
dennisroberts71: This isn't your code, but `when-not` can be used here.

services/terrain/src/terrain/services/filesystem/metadata.clj
```diff
@@ -208,33 +79,19 @@
...
-  "Deletes an AVU from path on behalf of a user. attr and value should be strings."
-  [user path attr value]
-  (with-jargon (icat/jargon-cfg) [cm]
-    (validators/user-exists cm user)
-    (validators/path-exists cm path)
-    (validators/path-writeable cm user path)
-    (authorized-avus [{:attr attr :value value :unit ""}])
-    (delete-metadata cm path attr value)
-    {:path path :user user}))
+  (let [dest-items (-> (data-raw/collect-stats user :ids dest-ids :validation-behavior "write") :body json/decode (get "ids") vals walk/keywordize-keys)
```
dennisroberts71: I wasn't aware of `keywordize-keys`. Nice!

dennisroberts71:
This pull request looks good to me.

PR #22, opened 2016-01-13
Remove de target CORE-7379

strootman:
This PR relates only to CORE-7379 and its subtasks


ansible/roles/util-cfg-systemd-unit/templates/elk-logstash.service.j2
```diff
@@ -1,26 +0,0 @@
...
-        --log-driver={{docker.log_driver}} --log-opt syslog-tag={{elk.logstash.service_name_short}} \
-        {{docker.user}}/{{elk.logstash.image_name}} logstash -f /config-dir
-ExecStop=/usr/bin/docker stop {{elk.logstash.container_name}}
-Restart=on-failure
-StartLimitInterval=60s
-StartLimitBurst=3
-
-[Install]
-WantedBy={{systemd.target}}
-
```
dennisroberts71: I really need to reread the Ansible stuff in the near future. I assume that the logstash forwarder stuff is also now managed by docker-compose...

strootman: No, it won't be.  Currently, the ELK stuff is undergoing a refactor, and will be considered as infrastructure. Most, if not all, of the ELK apps will no longer be dockerized.

Do I need to make the ticket/sub-tasks more clear?

ansible/roles/de-services-cfg/README.md
```diff
@@ -22,9 +22,8 @@ N/A
 Dependencies
 ------------
 
-util-cfg-logging
```
dennisroberts71: It appears that util-cfg-logging isn't referenced at all anymore, although it could potentially be used to create local logging config files. Would it be reasonable to remove the role or is there a reason to keep it around?

strootman: Good catch. It should go away.

ansible/roles/util-cfg-systemd-unit/templates/elk-logstash.service.j2
dennisroberts71: Cool. To be honest, I didn't read the linked issues, which would have answered my question. The issues described well enough.

PR #23, opened 2016-01-13
CORE-4831 Add logic to enable/disable Notification's DeleteAll btn

ash3rz:
Added method to enable/disable Delete All notifications button
Added listStore size check on grid refresh to determine if DeleteAll button should be enabled


dennisroberts71:
I'm not an expert on our UI code by any means, but this change looks good to me.

strootman:
:shipit:

PR #24, opened 2016-01-14
CORE-6211: migrate /secured/fileio/download 

ianmcorvidae:
This makes it use the data-info logic already in services.filesystem.updown, for the display-download/special-download URL.

(This is a pretty small pull request, put up in the interest of continuing to make us use PRs better)

### DIFF
--- a/services/terrain/src/terrain/routes/fileio.clj
+++ b/services/terrain/src/terrain/routes/fileio.clj
@@ -10,8 +10,8 @@
   []
   (util/optional-routes [config/data-routes-enabled]
 
-    (GET "/fileio/download" [:as {:keys [params]}]
-      (fio/download params))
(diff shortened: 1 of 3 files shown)


services/terrain/src/terrain/routes/fileio.clj
```diff
@@ -10,8 +10,8 @@
   []
   (util/optional-routes [config/data-routes-enabled]
 
-    (GET "/fileio/download" [:as {:keys [params]}]
-      (fio/download params))
+    (GET "/fileio/download" [:as req]
+      (util/controller req fio/download :params))
```
ianmcorvidae: This construction using `terrain.util/controller` matches the things we're doing in `routes.filesystem`, which seemed easier than discerning exactly what differences mattered and didn't.

PR #26, opened 2016-01-20
CORE-7417: add a separate role to generate .pgpass files

dennisroberts71:
From a monster pull request to a mini pull request. :smile:

### DIFF
--- a/ansible/roles/util-pgpass/tasks/main.yaml
+++ b/ansible/roles/util-pgpass/tasks/main.yaml
@@ -1,0 +1,4 @@
+---
+- name: place .pgpass
+  sudo: yes
+  template: src=postgres.pgpass.j2 dest=~/.pgpass mode=0600 backup=yes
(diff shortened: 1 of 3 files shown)


ansible/roles/util-pgpass/tasks/main.yaml
```diff
@@ -0,0 +1,4 @@
+---
+- name: place .pgpass
+  sudo: yes
+  template: src=postgres.pgpass.j2 dest=~/.pgpass mode=0600 backup=yes
```
strootman: If there is ever a chance that we will want to use this to place a .pgpass in a different user's home directory, then you'll want to remove the `sudo: yes`.

Otherwise, keep the `sudo` and be explicit with the `dest`, making it `/root/.pgpass`. Makes it unambiguous.

dennisroberts71: Good suggestion. We're using .pgpass within docker files, so I doubt if we'll ever want to put a .pgpass file in another user's home directory. I'll make the destination explicit.

dennisroberts71: I made the destination `~root/.pgpass`. It's still explicit and it will work in the (unlikely) case where root's home directory is something other than `/root`.

strootman:
:shipit:

PR #28, opened 2016-01-21
Sharing services

dennisroberts71:
These are the changes to implement the app permission listing service.

### DIFF
--- a/services/apps/src/apps/routes/domain/permission.clj
+++ b/services/apps/src/apps/routes/domain/permission.clj
@@ -8,7 +8,7 @@
 
 (defschema UserPermission
   {:user       (describe String "The user ID")
-   :permission (describe String "The permission level assigned to the user")})
+   :permission (describe (enum "read" "write" "own" "") "The permission level assigned to the user")})
 
 (defschema AppPermissions
   {:id          (describe String "The app ID")
--- a/services/apps/src/apps/routes/domain/permission.clj
+++ b/services/apps/src/apps/routes/domain/permission.clj
@@ -1,17 +1,17 @@
 (ns apps.routes.domain.permission
-  (:use [common-swagger-api.schema :only [describe]]
+  (:use [common-swagger-api.schema :only [describe NonBlankString]]
         [schema.core :only [defschema optional-key enum]])
   (:import [java.util UUID]))
 
 (defschema AppIdList
-  {:apps (describe [String] "A List of app identifiers")})
+  {:apps (describe [NonBlankString] "A List of app identifiers")})
 
 (defschema UserPermission
-  {:user       (describe String "The user ID")
+  {:user       (describe NonBlankString "The user ID")
    :permission (describe (enum "read" "write" "own" "") "The permission level assigned to the user")})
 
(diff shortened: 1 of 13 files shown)


services/apps/src/apps/routes/domain/permission.clj
```diff
@@ -0,0 +1,18 @@
...
+  (:use [common-swagger-api.schema :only [describe]]
+        [schema.core :only [defschema optional-key enum]])
+  (:import [java.util UUID]))
+
+(defschema AppIdList
+  {:apps (describe [String] "A List of app identifiers")})
+
+(defschema UserPermission
+  {:user       (describe String "The user ID")
+   :permission (describe String "The permission level assigned to the user")})
```
ianmcorvidae: It'd be better if this was an `s/enum` of the available permissions, if that's something we can expect to be able to know in advance.

services/apps/src/apps/routes/domain/permission.clj
```diff
@@ -0,0 +1,18 @@
+(ns apps.routes.domain.permission
+  (:use [common-swagger-api.schema :only [describe]]
+        [schema.core :only [defschema optional-key enum]])
+  (:import [java.util UUID]))
+
+(defschema AppIdList
+  {:apps (describe [String] "A List of app identifiers")})
```
ianmcorvidae: Can this be a UUID instead? Not sure if we need to worry about agave here.

services/apps/src/apps/routes/domain/permission.clj
```diff
@@ -0,0 +1,18 @@
...
+
+(defschema UserPermission
+  {:user       (describe String "The user ID")
```
ianmcorvidae: I'd have this be `NonBlankString` from `common-swagger-api.schema` instead (same for other strings that shouldn't be blank), just for some extra validation/documentation.

services/apps/src/apps/routes/domain/permission.clj
dennisroberts71: I considered making it an `enum`, but Grouper supports arbitrary action names. That said, we are sticking to `read`, `write`, `own`, and the empty string right now. I suppose we could make it an enum and change it later if necessary.

services/apps/src/apps/routes/domain/permission.clj
dennisroberts71: We do have to worry about agave here. Agave app IDs won't be UUIDs. We're not going to be supporting permission listing for apps immediately, but we will be supporting them.

services/apps/src/apps/routes/domain/permission.clj
ianmcorvidae: We depend on those three being the only available things when calculating precedence anyway, looking lower, so I think it'd make sense to enforce it.

dennisroberts71: Yeah, that's a good point. making the change now.

ianmcorvidae:
Aside from those couple of relatively minor comments, lgtm.

PR #29, opened 2016-01-21
Add Permanent ID request database tables

psarando:


databases/metadata/src/main/constraints/12_permanent_id_request_types.sql
```diff
@@ -0,0 +1,14 @@
+SET search_path = public, pg_catalog;
+
+--
+-- permanent_id_requests table primary key.
```
dennisroberts71: This is trivial, but it looks like `permanent_id_requests` should be `permanent_id_request_types` in this comment.

dennisroberts71:
Other than the comment I mentioned above, this looks good to me.

PR #30, opened 2016-01-21
Add Permanent ID request CRUD service endpoints

psarando:
Please let Paul merge this pull request when it's ready.

### DIFF
--- a/services/metadata/src/metadata/persistence/permanent_id_requests.clj
+++ b/services/metadata/src/metadata/persistence/permanent_id_requests.clj
@@ -12,9 +12,9 @@
   "Adds a where clause to a query, filtering out all conditions for which the value is nil."
   [query clause]
   (let [clause (remove-nil-values clause)]
-    (if clause
-      (where query (remove-nil-values clause))
-      query)))
+    (if (empty? clause)
+      query
+      (where query clause))))
 
 (defn- list-permanent-id-requests-subselect
   "Creates a subselect query that can be used to list Permanent ID Requests."
@@ -33,7 +33,7 @@
     (join [:permanent_id_request_status_codes :status_codes]
           {:status_codes.id :statuses.permanent_id_request_status_code})
     (where-if-defined {:requested_by user})
-    (order :statuses.date_assigned :DESC)))
+    (order :statuses.date_assigned :ASC)))
 
 (defn- list-permanent-id-requests-query
   "Lists the Permanent ID Requests that have been submitted by the user."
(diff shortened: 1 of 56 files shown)


services/metadata/src/metadata/persistence/permanent_id_requests.clj
```diff
@@ -0,0 +1,200 @@
...
+  ((comp :name first)
+    (select [:permanent_id_requests :r]
+            (fields :status_codes.name)
+            (join [:permanent_id_request_statuses :statuses]
+                  {:statuses.permanent_id_request :r.id})
+            (join [:permanent_id_request_status_codes :status_codes]
+                  {:status_codes.id :statuses.permanent_id_request_status_code})
+            (where {:r.id request-id})
+            (order :statuses.date_assigned :DESC)
+            (limit 1))))
```
dennisroberts71: I usually leave out the top-level table (`permanent_id_requests` in this case) in these types of queries, but the effect is equivalent. I do it just to remove one of the join levels, which is probably done by the query optimizer anyway. The upshot is that no change is required, but I normally do something slightly different. :grinning:

services/metadata/src/metadata/routes/domain/permanent_id_requests.clj
```diff
@@ -0,0 +1,80 @@
...
+                                          SortFieldOptionalKey
+                                          StandardUserQueryParams]]
+        [clojure-commons.error-codes]
+        [metadata.routes.domain.common])
+  (:require [metadata.persistence.permanent-id-requests :as db]
+            [schema.core :as s])
+  (:import [java.util UUID]))
+
+(def PermanentIDRequestIdParam (describe UUID "The Permanent ID Requests's UUID"))
+(def PermanentIDRequestTypeEnum (apply s/enum (map :type (db/list-permanent-id-request-types))))
```
dennisroberts71: I'd be a little bit worried about this causing some errors if the namespace happens to be loaded before the configuration file has been read.

psarando:
Closing to open a new PR onto `dev`.

services/metadata/src/metadata/routes/domain/permanent_id_requests.clj
ianmcorvidae: It does, but we also already do this elsewhere: https://github.com/cyverse/DE/blob/dev/services/metadata/src/metadata/routes/domain/template.clj#L9

ianmcorvidae: (by "it does" I mean it does error out since it doesn't have a database configuration, but it really only comes up when you're doing things like running linters which load the namespaces before configuration)

dennisroberts71: I'm not really worried about linters and compilers, but I was a little bit worried about the namespace being loaded directly. Upon closer inspection, it appears that `metadata.core` explicitly loads the namespace after the configuration file has been loaded, though. Sorry for the false alarm.

dennisroberts71: Well, what I should have said was that `metadata.core` explicitly loads `metadata.routes`, which is the only namespace that uses any of the schema definition namespaces.

PR #34, opened 2016-01-22
Add Permanent ID request CRUD service endpoints

psarando:
See previous comments in https://github.com/cyverse/DE/pull/30

### DIFF
--- a/services/terrain/src/terrain/clients/data_info.clj
+++ b/services/terrain/src/terrain/clients/data_info.clj
@@ -18,6 +18,7 @@
             [terrain.services.filesystem.stat :as st]
             [terrain.services.filesystem.users :as users]
             [terrain.services.filesystem.uuids :as uuids]
+            [terrain.services.filesystem.validators :as validators]
             [terrain.util.config :as cfg]
             [terrain.util.service :as svc])
   (:import [clojure.lang IPersistentMap ISeq Keyword]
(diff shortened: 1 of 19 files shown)


services/terrain/src/terrain/services/permanent_id_requests.clj
```diff
@@ -0,0 +1,149 @@
...
+  (->> data-item
+       (validate-owner user)
+       validate-staging-dest))
+
+(defn- get-requested-data-item
+  "Gets data-info stat for the given ID and checks if the data item is valid for a Permanent ID request."
+  [user data-id]
+  (->> data-id
+       (data-info/stat-by-uuid user)
+       (validate-data-item user)))
```
ianmcorvidae: (originally saw this on the other PR but it got added here)

It may make sense to specifically pass the validation behavior to data-info while getting the stat information, rather than using the owner (understanding, of course, that I added that to the endpoint while you were writing this, so it's quite reasonable you didn't), to reduce the number of data-info requests that need to be made.

services/terrain/src/terrain/clients/data_info.clj
```diff
@@ -148,11 +150,12 @@
 
 (defn rename
   "Uses the data-info set-name endpoint to rename a file within the same directory."
-  [params body]
-  (assertions/assert-valid (= (ft/dirname (:dest body)) (ft/dirname (:source body)))
+  [{:keys [user]} {:keys [source dest]}]
```
dennisroberts71: :+1: adding destructuring here really improves the readability.

PR #35, opened 2016-01-22
Add admin EZID submission service endpoints

psarando:
See previous comments in https://github.com/cyverse/DE/pull/31

### DIFF
--- a/services/terrain/src/terrain/clients/ezid.clj
+++ b/services/terrain/src/terrain/clients/ezid.clj
@@ -1,0 +1,84 @@
+(ns ^{:author psarando} terrain.clients.ezid
+  (:use [ring.util.http-response :only [charset]]
+        [slingshot.slingshot :only [try+ throw+]])
+  (:require [cemerick.url :as curl]
+            [clj-http.client :as http]
+            [clojure.string :as string]
+            [terrain.util.config :as config]))
+
+(defn- auth-params
+  []
+  [(config/ezid-username) (config/ezid-password)])
+
+(defn- ezid-uri
+  [& components]
+  (str (apply curl/url (config/ezid-base-url) components)))
+
+(defn- anvl-escape
+  [value]
+  (-> value
+      (string/replace #"%"  "%25")
+      (string/replace #"\n" "%0A")
+      (string/replace #"\r" "%0D")
+      (string/replace #":"  "%3A")))
+
+(defn- anvl-unescape
+  [value]
+  (-> value
+      (string/replace #"%3A" ":")
+      (string/replace #"%0D" "\r")
+      (string/replace #"%0A" "\n")
+      (string/replace #"%25" "%")))
+
+(defn- anvl-decode
+  "Decodes an ANVL string into a map."
+  [anvl]
+  (let [anvl-lines (when (string? anvl) (->> anvl string/split-lines (remove string/blank?)))
+        keywordize-fn #(update-in % [0] keyword)]
+    (if (and anvl-lines (every? #(.contains % ":") anvl-lines))
+      (into {}
+        (map (fn [kv] (keywordize-fn (mapv anvl-unescape (string/split kv #": *" 2))))
+             anvl-lines))
+      anvl)))
+
+(defn- anvl-encode
+  "Encodes the given metadata map as an ANVL string."
+  [metadata]
+  (->> metadata
+       (map (fn [[k v]] (str (anvl-escape (name k)) ": " (anvl-escape v))))
+       (string/join "\n")))
+
+(defn- ezid-post
+  "Posts a request to the EZID API and parses its ANVL response."
+  [body & uri-parts]
+  (->> (charset {:body       (anvl-encode body)
+                 :basic-auth (auth-params)}
+                "utf-8")
+       (http/post (apply ezid-uri uri-parts))
+       :body
+       anvl-decode))
+
+(defn- format-anvl-error
+  "Adds the ANVL message to the given error map."
+  [error-map anvl]
+  (let [decoded-anvl (anvl-decode anvl)]
+    (if (map? decoded-anvl)
+      (merge error-map decoded-anvl)
+      (assoc error-map :error decoded-anvl))))
+
+(defn mint-id
+  "Mints a new permanent ID under the given shoulder with the given metadata.
+   http://ezid.cdlib.org/doc/apidoc.html#operation-mint-identifier"
+  [shoulder metadata]
+  (try+
+    (let [response (ezid-post metadata "shoulder" shoulder)
+          keywordize-fn #(update-in % [0] (comp keyword string/upper-case))]
+      (when-not (and (map? response) (contains? response :success))
+        (throw+ {:type :clojure-commons.exception/request-failed
+                 :error "Could not parse EZID response."
+                 :response response}))
+      (into {}
+        (mapv #(keywordize-fn (string/split % #" *: *"))
+              (string/split (:success response) #" *\| *"))))
+    (catch [:status 400] {:keys [body] :as bad-request}
+      (throw+ (format-anvl-error {:type :clojure-commons.exception/bad-request} body)))))
--- a/services/terrain/src/terrain/services/admin.clj
+++ b/services/terrain/src/terrain/services/admin.clj
@@ -76,6 +76,18 @@
       (log/error (ce/format-exception e))
       false)))
 
+(defn perform-ezid-check
+  []
+  (try
+    (let [ezid-status-url (str (url/url (config/ezid-base-url) "status"))
+          status          (:body (client/get ezid-status-url))]
+      (log/info "HTTP Status from EZID: " status)
+      status)
+    (catch Exception e
+      (log/error "Error performing EZID status check:")
+      (log/error (ce/format-exception e))
+      false)))
+
 
 (defn- status-irods
   [status]
@@ -101,6 +113,10 @@
     (merge status {:notificationagent (perform-notificationagent-check)})
     status))
 
+(defn status-ezid
+  [status]
+  (merge status {:ezid (perform-ezid-check)}))
+
 (defn status
   "Returns JSON containing the Terrain's status."
   [request]
@@ -109,5 +125,6 @@
     (status-jex)
     (status-apps)
     (status-notificationagent)
+    (status-ezid)
     success-response))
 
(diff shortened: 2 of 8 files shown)


services/terrain/src/terrain/clients/ezid.clj
```diff
@@ -0,0 +1,84 @@
...
+      (string/replace #"\r" "%0D")
+      (string/replace #":"  "%3A")))
+
+(defn- anvl-unescape
+  [value]
+  (-> value
+      (string/replace #"%3A" ":")
+      (string/replace #"%0D" "\r")
+      (string/replace #"%0A" "\n")
+      (string/replace #"%25" "%")))
```
ianmcorvidae: This is perhaps finicky but I'd love these escape/unescape to be more declarative.

I believe that this will work:

```clojure
(def ^:private anvl-replacements
  {"\n" "%0A"
   "\r" "%0D"
   "%" "%25"
   ":" "%3A"})

(defn- anvl-escape [value]
  (reduce #(string/replace %1 (re-pattern (first %2)) (second %2)) value anvl-replacements))

(defn- anvl-unescape [value]
  (reduce #(string/replace %1 (re-pattern (second %2)) (first %2)) value anvl-replacements))
```

I wasn't completely sure this would work with `\r` and `\n` but repl testing seems to suggest it will: 

```
=> (reduce #(string/replace %1 (re-pattern (second %2)) (first %2)) "a%3Aoe%25ui%0D%0A" {"\n" "%0A" "\r" "%0D" "%" "%25" ":" "%3A"})
"a:oe%ui\r\n"
```

ianmcorvidae: er, I guess other testing I did was more representative:

```
=> (reduce #(string/replace %1 (re-pattern (first %2)) (second %2)) "aoeui\n" {"\n" "%0A"})
"aoeui%0A"
=> (reduce #(string/replace %1 (re-pattern (second %2)) (first %2)) "aoeui%0A" {"\n" "%0A"})
"aoeui\n"
```

services/terrain/src/terrain/clients/ezid.clj
```diff
@@ -0,0 +1,84 @@
...
+
+(defn- anvl-decode
+  "Decodes an ANVL string into a map."
+  [anvl]
+  (let [anvl-lines (when (string? anvl) (->> anvl string/split-lines (remove string/blank?)))
+        keywordize-fn #(update-in % [0] keyword)]
+    (if (and anvl-lines (every? #(.contains % ":") anvl-lines))
+      (into {}
+        (map (fn [kv] (keywordize-fn (mapv anvl-unescape (string/split kv #": *" 2))))
+             anvl-lines))
```
ianmcorvidae: I think when I found it for a different PR you mentioned this, but using `clojure.walk/keywordize-keys` may be nicer here. The only issue would be if this is a deeply-nested structure and you don't want to keywordize keys in nested maps, but I think I recall that not being possible with ANVL anyway.

services/terrain/src/terrain/clients/ezid.clj
psarando: This might be a good idea, except we still need to guarantee order, since we don't want to escape everything else before attempting to escape `%`. So maybe a vector of vectors could work instead of a map?

ianmcorvidae: Ah, good point. an array-map would probably also work, but in any case when it's turned into a seq by `reduce` it's always going to be a sequence of 2-vectors so doing that directly seems good

services/terrain/src/terrain/clients/ezid.clj
ianmcorvidae: Er, hm. Looking again I realize your way is better, since it means you're not creating a map and then recreating the map with `keywordize-keys`, just creating it right the first time. It is a bit harder to read, but that's probably no big deal.

services/terrain/src/terrain/services/admin.clj
```diff
@@ -101,6 +113,10 @@
     (merge status {:notificationagent (perform-notificationagent-check)})
     status))
 
+(defn status-ezid
```
ianmcorvidae: This, and perform-ezid-check, should be made private in this namespace to match the other status checks.

services/terrain/src/terrain/clients/ezid.clj
```diff
@@ -0,0 +1,84 @@
...
+    (if (map? decoded-anvl)
+      (merge error-map decoded-anvl)
+      (assoc error-map :error decoded-anvl))))
+
+(defn mint-id
+  "Mints a new permanent ID under the given shoulder with the given metadata.
+   http://ezid.cdlib.org/doc/apidoc.html#operation-mint-identifier"
+  [shoulder metadata]
+  (try+
+    (let [response (ezid-post metadata "shoulder" shoulder)
```
ianmcorvidae: Speaking completely from a position of ignorance about ezid/etc., is it okay that we're not doing any validation of the metadata before we post it to ezid?

services/terrain/src/terrain/services/permanent_id_requests.clj
```diff
@@ -38,11 +44,21 @@
              :path staging-dest})))
   data-item)
 
+(defn- validate-publish-dest
+  [{:keys [path] :as data-item}]
+  (let [publish-dest (ft/path-join (config/permanent-id-publish-dir) (ft/basename path))]
+    (when (data-info/path-exists? (config/irods-user) publish-dest)
+      (throw+ {:type :clojure-commons.exception/exists
+               :error "A folder with this name has already been published."
+               :path publish-dest})))
```
ianmcorvidae: This block might be another use case for a validator endpoint in data-info, to push the particular logic of which error codes to throw for files which already exist into it. I suppose this error message would be harder to preserve that way, though.

services/terrain/src/terrain/clients/ezid.clj
psarando: I'm not sure what kind of validation we can do. The EZID API will return a 400 status with an error message if the metadata doesn't pass its own validation (such as missing a `datacite.creator` attribute), and that error message will get returned to the UI.

services/terrain/src/terrain/services/permanent_id_requests.clj
```diff
@@ -38,11 +44,21 @@
              :path staging-dest})))
   data-item)
 
+(defn- validate-publish-dest
```
ianmcorvidae: I notice that both this and `validate-staging-dest` are checking existence, so it might save a bit to only call `existence-marker` once.

services/terrain/src/terrain/services/permanent_id_requests.clj
psarando: I'm not sure this validation makes sense in `data-info`.
We have a folder in a user's home directory, and we're trying to see if a folder of the same name exists in the public `curated` folder (or in the `staging` folder) before successfully creating the request, or before attempting to create a DOI for the folder, since the folder will end up in the `curated` destination once the DOI creation process is complete.

This is all `permanent-id-request` service logic.

ianmcorvidae: What I mean is, checking that the path exists and throwing a `clojure-commons.exception/exists` error is all the sort of stuff data-info is doing for all sorts of other things -- this is just `validators/path-not-exists`, basically. The only thing that's different is the more-specific error message about a folder with that name already having been published. Or, put differently, the link between a path existing when it shouldn't and the `clojure-commons.exception/exists` error type is the concern of data-info, despite the reason the path should not exist being unrelated.

psarando: This validation isn't throwing the exception only because we're trying to move the folder and the name in the destination already exists. For example, when the user makes the initial request, we validate that name doesn't exist in the `curated` folder, but it's only moving into the `staging` folder for now.

So this exception doesn't have to be `clojure-commons.exception/exists`. It could be a `bad-request` or even eventually become a `permanent-id-request` specific error code.

ianmcorvidae: (long response incoming; I think what I'm saying isn't coming through concisely. sorry about that!)

I understood what it's doing as far as the validation; none of this is happening at the same time as things actually moving around to all of the paths being validated in iRODS. I think that it's reasonable to have it where it is for the sake of the more specific message even if it doesn't do anything else (like being a different code), but this seemed like a place to have a discussion about the boundaries and roles of the microservices -- whether data-info is only tasked with providing information in cases like this, and thus only taking action or throwing errors when it comes to modifying things in iRODS (i.e.: get the info from data-info, throw the error in terrain) or if its role also can include validating things in iRODS, like a file existing or not. Another version of this might say that data-info's scope includes validation as a primary task, even for operations being performed elsewhere, but that it can only throw generic errors, such that this bit of code might call a validator endpoint and catch an ERR_EXISTS (if it gets one) to rethrow as a more descriptive error.

For my own part, parallel to this work, I'm trying to figure out what to do with the urlupload endpoint -- it needs to validate some very similar things to this (user exists in irods, destination directory writeable, destination file doesn't already exist). Since we don't want jargon in terrain, of course this will require some calls to data-info, but it's not clear which calls exactly: a call each to existence-marker and to stat-gatherer, with terrain needing to parse the permissions and existence information out and throw its own errors; a call or several to endpoints which check things and either return success responses or throw errors; or just one call and let data-info call out to apps itself to start the job. Past discussions have largely if not completely ruled out the third (it could still make sense to have data-info call out to apps, if we consider the fact we're using an app to do urlupload as an implementation detail for something which is fundamentally an action in iRODS i.e. belongs in data-info), but which of the first/second is ideal isn't clear to me. What you're advocating would be the first, which is possibly easier to explain but more code and more requests into data-info, and what I've been advocating is the second, which seems to be harder to explain (unless that's just me being bad at explaining things, which is possible :P) but reduces the cross-talk between the services that's needed as well as the need for terrain to parse information out of data-info requests solely for validation. Your case the number of calls to data-info is probably the same either way, but it could still be a cleaner separation of concerns or less code one way or the other.

Anyway, how it is is fine, since the point here was the discussion. But I'm still curious to thresh out these boundaries better.

psarando: OK, thanks for clarifying. I think I see where you're coming from, now.
This does deserve a design discussion.

PR #38, opened 2016-01-22
CORE-7446: migrate /secured/fileio/upload

ianmcorvidae:
This also lays the groundwork for migrating save and saveas, which in turn will lay the groundwork for anything else that wants to save files into iRODS (e.g. re-migrating the metadata save stuff partly out of data-info to remove its dependency on the metadata service).

### DIFF
--- a/services/data-info/src/data_info/services/write.clj
+++ b/services/data-info/src/data_info/services/write.clj
@@ -1,0 +1,59 @@
+(ns data-info.services.write
+  (:use [clj-jargon.init :only [with-jargon]])
+  (:require [clojure-commons.file-utils :as ft]
+            [clojure-commons.error-codes :as ce]
+            [clj-jargon.item-info :as info]
+            [clj-jargon.item-ops :as ops]
+            [ring.middleware.multipart-params :as multipart]
+            [data-info.services.stat :as stat]
+            [data-info.util.config :as cfg]
+            [data-info.util.validators :as validators]))
+
+(defn- save-file-contents
+  "Save an istream to a destination. Relies on upstream functions to validate."
+  [cm istream user dest-path]
+  (ops/copy-stream cm istream user dest-path)
+  dest-path)
+
+(defn- create-at-path
+  "Create a new file at dest-path from istream.
+
+   Error if the path exists or if the destination directory does not exist or is not writeable."
+  [cm istream user dest-path]
+  (let [dest-dir (ft/dirname dest-path)]
+    (validators/user-exists cm user)
+    (validators/path-not-exists cm dest-path)
+    (validators/path-exists cm dest-dir)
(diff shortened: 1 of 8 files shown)


services/terrain/src/terrain/clients/data_info/raw.clj
```diff
@@ -97,6 +97,15 @@
 
 ;; CREATE
 
+(defn upload-file
+  [user dest-path filename content-type istream]
+  (http/post (str (url/url (cfg/data-info-base) "data"))
+             {:query-params {:user user
+                             :dest dest-path}
+              :multipart [{:part-name "file"
+                           :name filename
```
ianmcorvidae: `:part-name` defines what key it ends up under, but `:name` defines the filename that'll be used. This is not at all documented in clj-http. Multipart requests are annoying.

services/terrain/src/terrain/services/fileio/controllers.clj
```diff
@@ -48,10 +44,8 @@
      req  - the ring request map"
   [{:keys [user dest]} ^IPersistentMap req]
   (let [store                   (partial store-from-form user dest)
-        {{file "file"} :params} (multipart/multipart-params-request req {:store store})]
-    (when-not file
-      (throw+ {:error_code ERR_MISSING_FORM_FIELD :field "file"}))
-    (success-response {:file (data/path-stat user file)})))
+        {{file-info "file"} :params} (multipart/multipart-params-request req {:store store})]
```
ianmcorvidae: The value that the middleware's store function returns ends up in the key relating to the part name, i.e. 'file'; so, file-info here is the data-info response, which matches the former return value.

services/data-info/src/data_info/services/write.clj
```diff
@@ -0,0 +1,59 @@
...
+   Error if the path exists or if the destination directory does not exist or is not writeable."
+  [cm istream user dest-path]
+  (let [dest-dir (ft/dirname dest-path)]
+    (validators/user-exists cm user)
+    (validators/path-not-exists cm dest-path)
+    (validators/path-exists cm dest-dir)
+    (validators/path-writeable cm user dest-dir)
+    (save-file-contents cm istream user dest-path)))
+
+(defn- overwrite-path
```
dennisroberts71: It appears that this function is never called. I assume that it will be called at some point in the future, but I thought I should point it out in case that assumption is incorrect.

ianmcorvidae: Ah, yes, that's true; I hadn't thought about it being unused just yet. It'll be used for the `/secured/fileio/save` endpoint, ultimately.

dennisroberts71:
:shipit: Looks good to me.

PR #39, opened 2016-01-25
CORE-6213: migrate /secured/fileio/saveas

ianmcorvidae:
This builds on top of #38, which means it's probably better to review that first. The changes in this PR, from that one, are all in the most-recent commit, should you be reviewing it before that one's merged (at which point this PR should no longer show those changes).


dennisroberts71:
:shipit: Looks good to me.

PR #40, opened 2016-01-25
CORE-7456: omit the requesting user from permission listing results

dennisroberts71:
This is a tiny change to omit the requesting user's permissions from app permission listings.


ianmcorvidae:
looks straightforward to me. :shipit:

PR #41, opened 2016-01-25
CORE-6212: migrate /secured/fileio/save (file overwrite)

ianmcorvidae:
This institutes another data-info endpoint which takes the file as a multipart request much like the upload (because that's a lot easier than trying to get compojure-api to let you use a non-json body, plus it's consistent) and overwrites it, at `PUT /data/:data-id`. It then uses it for `/secured/fileio/save` much like how `saveas` was done, turning the JSON input into a buffered input stream to send to the data-info endpoint.

### DIFF
--- a/ansible/roles/util-cfg-service/templates/terrain.properties.j2
+++ b/ansible/roles/util-cfg-service/templates/terrain.properties.j2
@@ -89,7 +89,6 @@ terrain.garnish.type-attribute = ipc-filetype
 
 # File IO settings
 terrain.fileio.url-import-app = 1E8F719B-0452-4D39-A2F3-8714793EE3E6
-terrain.fileio.max-edit-file-size = {{ max_edit_file_size }}
 
 # Filesystem settings
 terrain.data-info.base-url            = http://{{ data_info.host }}:{{ data_info.port }}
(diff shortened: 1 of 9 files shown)


ansible/roles/util-cfg-service/templates/terrain.properties.j2
```diff
@@ -89,7 +89,6 @@ terrain.garnish.type-attribute = ipc-filetype
 
 # File IO settings
 terrain.fileio.url-import-app = 1E8F719B-0452-4D39-A2F3-8714793EE3E6
-terrain.fileio.max-edit-file-size = {{ max_edit_file_size }}
```
ianmcorvidae: As we discussed at one point, I'm letting nginx handle this from here on in. This was only enforced for this endpoint anyway (not upload or saveas) and nginx is almost certainly better at it.

services/terrain/src/terrain/clients/data_info.clj
```diff
@@ -148,6 +148,12 @@
   (log/warn "checking to see if" path "can be created")
   (st/can-create-dir? user path))
 
+(defn overwrite-file
```
ianmcorvidae: This has a version in the plain client, rather than the raw client, to convert the path to a UUID.

strootman:
Code looks straight forward at a glance. If no one else gets to this, I'll review more thoroughly tomorrow.

ianmcorvidae:
(reopen)

psarando:
I didn't catch any obvious issues :shipit:

PR #42, opened 2016-01-26
Core 3833 and Core 5472

ash3rz:
There are two commits here.  One for each ticket.


psarando:

PR #44, opened 2016-01-26
Update pb hosts to prevent action on parasitic env

strootman:
The intent is to create an intersection of host groups with the
`[systems]` host group. The purpose is to prevent the accidental
execution of playbooks against "non-managed" hosts in the inventory
(i.e. I don't want to accidentally install/re-configure elasticsearch
for my inventory).

CORE-7460


dennisroberts71:
Looks straight-forward to me. :shipit:

PR #45, opened 2016-01-26
Core 6052

tedgin:
See https://pods.cyverse.org/jira/browse/CORE-6052

In brief this is an API change to terrain's data search endpoint, `/secured/filesystem/index`.  A new optional query parameter has been added, `includeTrash`, that allows DE user's trash to be included or excluded in the search results. 

As part of this change, all search results unreachable from the DE, like those from /iplant and /iplant/trash have been excluded.

### DIFF
--- a/services/terrain/project.clj
+++ b/services/terrain/project.clj
@@ -15,15 +15,15 @@
   :manifest {"Git-Ref" ~(git-ref)}
   :uberjar-name "terrain-standalone.jar"
   :dependencies [[org.clojure/clojure "1.7.0"]
-                 [org.clojure/tools.nrepl "0.2.10"]
+                 [org.clojure/tools.nrepl "0.2.12"]
                  [cheshire "5.5.0"]
                  [clj-http "2.0.0"]
                  [clj-time "0.11.0"]
-                 [clojurewerkz/elastisch "2.1.0"]
+                 [clojurewerkz/elastisch "2.2.0"]
                  [com.cemerick/url "0.1.1" :exclusions [com.cemerick/clojurescript.test]]
-                 [commons-net "3.3"]                               ; provides org.apache.commons.net
+                 [commons-net "3.4"]                               ; provides org.apache.commons.net
                  [compojure "1.4.0"]
-                 [metosin/compojure-api "0.24.2"]
+                 [metosin/compojure-api "0.24.3"]
                  [de.ubercode.clostache/clostache "1.4.0" :exclusions [org.clojure/core.incubator]]
                  [dire "0.5.3"]
                  [me.raynes/fs "1.4.6"]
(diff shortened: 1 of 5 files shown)


services/terrain/project.clj
```diff
@@ -15,15 +15,15 @@
...
                  [clj-http "2.0.0"]
                  [clj-time "0.11.0"]
-                 [clojurewerkz/elastisch "2.1.0"]
+                 [clojurewerkz/elastisch "2.2.0"]
                  [com.cemerick/url "0.1.1" :exclusions [com.cemerick/clojurescript.test]]
-                 [commons-net "3.3"]                               ; provides org.apache.commons.net
+                 [commons-net "3.4"]                               ; provides org.apache.commons.net
                  [compojure "1.4.0"]
-                 [metosin/compojure-api "0.24.2"]
+                 [metosin/compojure-api "0.24.3"]
```
ianmcorvidae: I'd not update this unless we're also updating the rest of our compojure-api imports. Especially notable is iplant-clojure-commons (of course, a patch release shouldn't be an issue, but compojure-api can change heavily).

services/terrain/src/terrain/persistence/search.clj
```diff
@@ -110,46 +110,55 @@
...
+                                                      path
+                                                      (str path \/))))
+        filter-tag  (fn [tag] (query/term :id {:type  "tag"
+                                               :id    tag
+                                               :path  "targets.id"
+                                               :cache false}))
+        perm-filter (query/nested :path   "userPermissions"
+                                  :filter (query/term "userPermissions.user" memberships))]
+    (query/bool :must   (query/bool :must perm-filter :should (map filter-path in-folders))
+                :should (map filter-tag tags))))
```
ianmcorvidae: I think I'm misunderstanding, but why is the folder path a `should` query? If I'm reading [the docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-bool-query.html) correctly, this means results could still be returned from bad paths if they match the tag filter.

Relatedly, where (if anywhere) do we set the [minimum_should_match](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-minimum-should-match.html) parameter? I guess since we have a `must` query it won't default to needing minimum-of-one, but I'd still be curious if this is just for scoring or if we require some to match.

tedgin: The elasticsearch docs are a little confusing. I had to test it out manually to make sure that `must` and `should` work together as I hoped.  The `must` and `must not` clauses must always pass.  If there is a should `clause` at least `minimum_should_match` or all of entries should pass for a result to be returned.

The `minimum_should_match` parameter defaults to `1`.

ianmcorvidae: Cool, thanks. Couldn't find the default for that parameter.

I realize that I also hadn't properly read the nesting, which was part of my confusion, so in case anyone else runs into it, it's `(:must (:must <permissions> :should <in folder>) :should <tags>)` which with the minimum of 1 works out to "must match perms and at least one folder and at least one tag".

ianmcorvidae:
LGTM, other than I'd probably pull out the compojure-api version bump since it's unrelated to this. Your call, however, so :shipit:

services/terrain/project.clj
tedgin: I'll remove that change and push again.

PR #46, opened 2016-01-26
Core 7441

dennisroberts71:
This one was interesting because grouper (:fish:) doesn't treat duplicate permission assignments as no-ops. I worked around this by removing any matching permission assignments (that is, any permission assignments that match on everything but the action name) before assigning the new permissions. Note: this only works for us because we never expect to have direct permission assignments to more than one action name.

### DIFF
--- a/services/apps/src/apps/clients/iplant_groups.clj
+++ b/services/apps/src/apps/clients/iplant_groups.clj
@@ -1,10 +1,12 @@
 (ns apps.clients.iplant-groups
   (:use [medley.core :only [remove-vals]]
-        [slingshot.slingshot :only [throw+]])
+        [slingshot.slingshot :only [try+]])
   (:require [apps.util.config :as config]
+            [apps.util.service :as service]
             [cemerick.url :as curl]
             [clj-http.client :as http]
             [clojure.string :as string]
+            [clojure.tools.logging :as log]
             [clojure-commons.exception :as cx]
             [kameleon.uuids :refer [uuidify]]))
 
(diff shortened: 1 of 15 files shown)


services/apps/src/apps/clients/iplant_groups.clj
```diff
@@ -136,3 +138,25 @@
...
+               :content-type :json
+               :as           :json}))
+  nil)
+
+(defn share-app
+  "Shares an app with a user."
+  [app-id subject-id level]
+  (try+
+   (share-app* app-id subject-id level)
+   (catch :status {:keys [body]}
```
ianmcorvidae: Might use `clj-http-error?` from clojure-commons for this rather than just `:status`, it'll be a bit more specific.

dennisroberts71: :+1: I'll change that.

services/apps/src/apps/clients/iplant_groups.clj
```diff
@@ -136,3 +138,25 @@
...
+     (let [reason (:grouper_result_message (service/parse-json body))]
```
ianmcorvidae: Should this perhaps have a default of some sort in case the error isn't directly from grouper -- e.g., if grouper-nginx sends something which errors? Getting "unable to share <uuid> with <subject>: " as an error wouldn't be as nice as something like "unable to share <uuid> with <subject>: non-grouper error caught"

services/apps/src/apps/routes/apps.clj
```diff
@@ -70,6 +70,19 @@
...
+         :query [params SecuredQueryParams]
+         :body [body (describe perms/AppSharingRequest "The app sharing request.")]
+         :return perms/AppSharingResponse
+         :summary "Share one or more apps with one or more users."
+         :description "This endpoint allows the caller to share multiple apps with multiple users. The
+         authenticated user must have ownership permission to every app in the request body for this
+         endpoint to succeed. Note: this is a potentially slow operation and the response is returned
+         synchronously. The DE UI handles this by allowing the user to continue working while the request
+         is being processed. When calling this endpoint, please be sure that the response timeout is long
+         enough."
```
ianmcorvidae: What happens if the response timeout isn't long enough for a particular request? As in, what'll get returned, what will and won't happen in the backend, etc.? (I assume the UI should know how to handle that case as well, but that's not this PR)

services/apps/src/apps/service/apps/de/permissions.clj
```diff
@@ -23,6 +25,16 @@
...
 
+(defn- get-permission-error
+  [user required-level app-id]
+  (try+
+   (let [perms (iplant-groups/load-app-permissions user [app-id])]
+     (when (lacks-permission-level perms required-level app-id)
+       (str "insufficient privileges for app: " (str app-id))))
+   (catch :status {:keys [body]}
+     (let [reason (:grouper_result_message (service/parse-json body))]
+       (str "unable to load permissions for " app-id ": " reason)))))
```
ianmcorvidae: (same comments as the other error block down here I think)

dennisroberts71:
I've addressed all of the comments. Merging the request now.

PR #48, opened 2016-01-27
CORE-7468 Ensure data curators get own perms on published folders.

psarando:
If the Permanent ID Request curated folder needs to be created, the data
curators group will get own permissions for it.
The data curators group also get own permissions for folders published
to the curated folder.

### DIFF
--- a/services/terrain/src/terrain/services/permanent_id_requests.clj
+++ b/services/terrain/src/terrain/services/permanent_id_requests.clj
@@ -67,7 +67,7 @@
 (defn- validate-publish-dest
   [{:keys [path] :as data-item}]
   (let [publish-dest (ft/path-join (config/permanent-id-publish-dir) (ft/basename path))
-        paths-exist (parse-service-json (data-info/check-existence {:user (config/irods-user)}
+        paths-exist (parse-service-json (data-info/check-existence {:user (config/permanent-id-curators-group)}
                                                                    {:paths [publish-dest]}))]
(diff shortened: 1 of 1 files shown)


services/terrain/src/terrain/services/permanent_id_requests.clj
```diff
@@ -98,9 +98,11 @@
 (defn- create-publish-dir
   "Creates the Permanent ID Requests publish directory, if it doesn't already exist."
   []
-  (let [publish-path (config/permanent-id-publish-dir)]
+  (let [publish-path (config/permanent-id-publish-dir)
+        curators     (config/permanent-id-curators-group)]
     (when-not (data-info/path-exists? (config/irods-user) publish-path)
-      (data-info-client/create-dirs (config/irods-user) [publish-path]))))
+      (data-info-client/create-dirs (config/irods-user) [publish-path])
+      (data-info/share (config/irods-user) [curators] [publish-path] "write"))))
```
ianmcorvidae: I believe this is fine here, but note that this will set the permission to exactly that, which I'm mentioning because it might be mildly unintuitive -- if the curators have ownership this'd revoke that. I think that'd only come up if for some reason the irods user and the curators are the same, so I'll assume it's fine unless you can think of something else.

ianmcorvidae:
Looks reasonable as long as you can't think of anything which would break with the way sharing works there. :shipit:

psarando:
Looks like we need requirements from the data commons team on whether they require `write` or `own` permission on these folders.

PR #50, opened 2016-01-28
Core 7443

dennisroberts71:
This is the implementation of the service to revoke permissions for apps.

### DIFF
--- a/services/iplant-groups/src/iplant_groups/clients/grouper.clj
+++ b/services/iplant-groups/src/iplant_groups/clients/grouper.clj
@@ -714,16 +714,15 @@
 (defn- remove-existing-permissions
   [username attribute-def-name & [params]]
   (->> (get-permission-assign-ids username attribute-def-name params)
-       (map (fn [[permission-type ids]] (remove-permission-assign-ids username permission-type ids)))
-       dorun))
+       (mapv (fn [[permission-type ids]] (remove-permission-assign-ids username permission-type ids)))))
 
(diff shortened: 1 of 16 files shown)


services/iplant-groups/src/iplant_groups/clients/grouper.clj
```diff
@@ -714,16 +714,15 @@
 (defn- remove-existing-permissions
   [username attribute-def-name & [params]]
   (->> (get-permission-assign-ids username attribute-def-name params)
-       (map (fn [[permission-type ids]] (remove-permission-assign-ids username permission-type ids)))
-       dorun))
+       (mapv (fn [[permission-type ids]] (remove-permission-assign-ids username permission-type ids)))))
 
-(defn- remove-existing-role-permissions
```
ianmcorvidae: It appears you used the membership version of this outside of this file, but not the role version. I'm ambivalent on whether it'd make sense to add the parallel endpoints now for consistency, make this private for consistency, or neither, so your call but something to consider.

ianmcorvidae:
Other than the minor for-your-consideration comment, looks good to me, so :shipit: when ready.

services/iplant-groups/src/iplant_groups/clients/grouper.clj
dennisroberts71: Yeah, I was ambivalent about that too. I think I'll leave it for now and make the function private in the next PR.

PR #51, opened 2016-01-28
Core 7469

dennisroberts71:
This PR fixes the problem where app searches don't include private apps that have been shared with the user.


ianmcorvidae:
Looks reasonable. :shipit:

PR #52, opened 2016-01-28
CORE-4791 & CORE-6747

ash3rz:
CORE-4791 is adding an announcement to show to the user when they've successfully copied an App.  Prior to this, there was no indication the app was copied.  The user was immediately presented with the Edit App window which made it unclear if the app was already copied or if they had to save the edit window first.

CORE-6747 is adding a message to indicate to the user when they've clicked on the HPC app category that they'll be redirected to the Agave auth page instead of a `302` error message.


psarando:
I don't see any obvious issues :shipit:

ash3rz:
Awesome - thanks!

PR #53, opened 2016-01-28
CORE-7470: added a new action name (that is, permission level) to the…

dennisroberts71:
… permission definitions for apps and analyses

This is a partial pull request that gets the DE back into a working state with the recent Grouper updates that I made.

### DIFF
--- a/tools/sharkbait/src/sharkbait/permissions.clj
+++ b/tools/sharkbait/src/sharkbait/permissions.clj
@@ -5,6 +5,7 @@
            [edu.internet2.middleware.grouper.misc SaveMode]
            [edu.internet2.middleware.grouper.permissions PermissionFinder]))
 
+(def originator "originator")
 (def read  "read")
 (def write "write")
 (def own   "own")
@@ -21,9 +22,10 @@
   "Sets the allowed actions for a permission definition."
   [permission-def]
   (let [delegate (.getAttributeDefActionDelegate permission-def)]
-    (.configureActionList delegate ["read" "write" "own"])
-    (add-implied-action delegate "own" "write")
-    (add-implied-action delegate "write" "read"))
+    (.configureActionList delegate [read write own originator])
+    (add-implied-action delegate originator own)
+    (add-implied-action delegate own write)
+    (add-implied-action delegate write read))
   (.store permission-def))
 
 (defn create-permission-def
--- a/services/apps/src/apps/service/apps/de/permissions.clj
+++ b/services/apps/src/apps/service/apps/de/permissions.clj
@@ -6,13 +6,18 @@
             [clojure.string :as string]
             [clojure-commons.exception-util :as cxu]))
 
-(def permission-precedence (into {} (map-indexed (fn [i v] (vector v i)) ["own" "write" "read"])))
+(def permission-precedence (into {} (map-indexed (fn [i v] (vector v i)) ["originator" "own" "write" "read"])))
+
+(def top-permission-precedence (permission-precedence "own"))
 
 (defn get-permission-level
   ([perms app-id]
      (get-permission-level (perms app-id)))
   ([perms]
-     (first (sort-by permission-precedence (map :action_name perms)))))
+     (->> (map :action_name perms)
+          (sort-by permission-precedence)
+          (remove (fn [level] (< (permission-precedence level) top-permission-precedence)))
+          first)))
 
 (defn has-permission-level
   [perms required-level app-id]
--- a/tools/sharkbait/src/sharkbait/apps.clj
+++ b/tools/sharkbait/src/sharkbait/apps.clj
@@ -7,28 +7,28 @@
             [sharkbait.roles :as roles]))
 
 (defn- extract-username
-  "Extracts the username of an owner from an app."
+  "Extracts the username of an originator from an app."
   [app]
   (->> (into [] (.getArray (:users app)))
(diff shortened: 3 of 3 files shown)


psarando:
What's an `originator` permission for?

dennisroberts71:
That's the permission level assigned to the person who initially created the entity. For example, an app integrator or a job submitter. The idea for apps is that it will allow us to distinguish apps that a user integrated from apps to which the user has been given `own` permission.

tools/sharkbait/src/sharkbait/permissions.clj
```diff
@@ -21,9 +22,10 @@
   "Sets the allowed actions for a permission definition."
   [permission-def]
   (let [delegate (.getAttributeDefActionDelegate permission-def)]
-    (.configureActionList delegate ["read" "write" "own"])
-    (add-implied-action delegate "own" "write")
-    (add-implied-action delegate "write" "read"))
+    (.configureActionList delegate [read write own originator])
+    (add-implied-action delegate originator own)
```
ianmcorvidae: Won't this cause issues once we want to use this to allow labels-only editing, or am I missing the intention?

I'd assumed this was the solution for the problem that came up in the meeting yesterday, that originators can still edit labels even after the app is published but not the rest -- but, with this line they'd always have ownership permission, so it wouldn't limit them and we'd have to remove the permission.

I'd figured we'd do something like set an explicit permission for the limited public editing we'll still allow, plus an originator thing like this, and set up the inheritance with two branches: own -> write -> limited write -> read as well as originator -> limited write; that way, we grant both ownership and originator to the originator and we can still remove all ownership and write permissions when an app is published and the originator retains their limited write permission.

But I suspect I'm misunderstanding :)

dennisroberts71: Yeah, this isn't the permission level for editing the labels of public apps; that one's still coming. The purpose of this level is to allow us to distinguish apps that the user integrated from apps to which the user has been given `own` permissions. We _could_ do that by filtering out apps in the user's `Apps Under Development` category, but that will stop working when we do away with the current category system. This seemed like a more sustainable option.

ianmcorvidae: Alright, makes sense. I think "originator" might be what's confusing me, because it sounds like a provenance term which you'd expect to remain valid even after it's made public (which this can't be, given the inheritance). So I wonder if there's maybe a better name -- I'm still not completely sure what the user-facing role of this level would be, which would determine that I think, but something like "lead" might make sense if the goal is just something like "ownership, but can't be revoked by other owners". Given you know what your plans are, though, I'm sure you have better ideas, heh

psarando: I guess I don't see the reason for this permission, but then again I don't see how it would hurt having it.
Is it supposed to prevent the originator from losing `own` permissions by other owners?

dennisroberts71: Yeah, I guess I could have explained the idea a little more clearly. I'm trying to enable the population of the `Shared with Me` category in the DE. This category will contain all apps that have been shared with the user either directly or (eventually) via user-managed groups. The problem is that when `read`, `write` and `own` are the only permission levels, there's no way to distinguish apps that the user has integrated from apps to which the user has been given `own` permissions by other users.

There are other options for ensuring that apps the user has integrated don't appear in the `Shared with Me` category, of course. We could:

1. Filter out apps that are in the user's `Apps Under Development` category.
2. Add a real `Shared with Me` category in the DE database.
3. Add a way to reliably identify the original integrator of an app in the DE database.
4. Add an `app-integrators` role to grouper that is reserved for permission assignments given to the original integrator of an app.

There are probably other options, but those are the ones that immediately come to mind. The problem with the first and second alternatives is that they will no longer work when the current app category system goes away. The third alternative would work, but the DE database seems like the wrong place to put this information since we're placing all other information about app sharing in Grouper. The last alternative is something that I hadn't considered before I added the new action. It might be a reasonable alternative, but I haven't been able to identify any advantage that this method has over the new action name.

I should mention that this action name is intended to be completely hidden from users. Its only purpose is to help the apps service determine which apps should go in the Shared with Me virtual category.

I'm not particularly attached to the `originator` name. It just seemed general enough to mean either `integrator` or `submitter`. The term, `lead`, doesn't really help me to identify the purpose of the action name either. I'm open to suggestions on this.

The original intent of the new action name wasn't to prevent other users from revoking permissions from the original integrator, but that is another potential use of it.

psarando: OK, that makes more sense. It might have helped if I read the JIRA title for this issue first `Update the app category listing endpoints to include a Shared With Me category` :confounded:

psarando: I think the `originator` label for this permission works fine.

dennisroberts71: Hehe, no problem.

ianmcorvidae: Okay, that does make more sense. I can get over my initial confusion of thinking it was a statement about provenance, especially with its particular text being non-user-facing, so that seems fine to me now that I understand. :shipit:

dennisroberts71: Cool, I'll probably stick with `originator` then.

PR #54, opened 2016-01-29
Refactor data-info routes.

ianmcorvidae:
A lot of my early migration work I put lots of things into the `data-info.routes.data` namespace and the `data-info.routes.domain.data` namespace, but those files were getting really big. So this splits out the AVU operations and the move/rename operations into their own namespaces, which are much more reasonably sized.

It also adds a quick note about the metadata save endpoint (which remains in the data namespace), since it should be migrated partly back up into terrain in order to decouple the data-info and metadata services.

### DIFF
--- a/services/data-info/src/data_info/routes.clj
+++ b/services/data-info/src/data_info/routes.clj
@@ -13,6 +13,7 @@
             [data-info.routes.filetypes :as filetypes-routes]
             [data-info.routes.users :as users-routes]
             [data-info.routes.navigation :as navigation-routes]
+            [data-info.routes.rename :as rename-routes]
             [data-info.routes.status :as status-routes]
             [data-info.routes.stats :as stat-routes]
             [data-info.routes.trash :as trash-routes]
@@ -45,6 +46,7 @@
      log-validation-errors]
     status-routes/status
     data-routes/data-operations
+    rename-routes/rename-routes
     avus-routes/avus-routes
     exists-routes/existence-marker
     filetypes-routes/filetypes-operations
--- a/services/data-info/src/data_info/routes.clj
+++ b/services/data-info/src/data_info/routes.clj
@@ -7,6 +7,7 @@
         [ring.util.response :only [redirect]])
   (:require [compojure.route :as route]
             [clojure-commons.exception :as cx]
+            [data-info.routes.avus :as avus-routes]
             [data-info.routes.data :as data-routes]
             [data-info.routes.exists :as exists-routes]
             [data-info.routes.filetypes :as filetypes-routes]
@@ -44,6 +45,7 @@
      log-validation-errors]
     status-routes/status
     data-routes/data-operations
+    avus-routes/avus-routes
     exists-routes/existence-marker
     filetypes-routes/filetypes-operations
     users-routes/permissions-gatherer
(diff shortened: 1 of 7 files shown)


psarando:
> since it should be migrated partly back up into terrain in order to decouple the data-info and metadata services.

I think we still need to have a design discussion about this. Should `terrain` be the master coordinator? Or should it be a dumb router and we let services interact with each other more? In case 2, does this mean adding feature flags to our other services too (like turning off apps interactions in data-info)?

ianmcorvidae:
I'm not sure either, but the past discussions I've had (with @dennisroberts71 some months ago, mostly) have suggested it's probably better to make data-info be dumber and smaller and coordinate between services higher up; that's how I migrated the metadata-related endpoints which coordinate between the two services (copy and csv-parser). But I would be curious to hear more thoughts on what defines the edges of the services; beyond "jargon needs to be in data-info and not in terrain" I haven't had much of a clear picture as far as intentions.

I have, of course, thought about it a lot. I hope you heard the wall of text coming ;)

One edge of the spectrum of options as I see it would be something in the vein of "anything which is addressed by a UUID or path in iRODS, or creates something new in iRODS, should happen through data-info. This would, for example, push all the metadata operations (AVUS or template/database) through data-info if they concern objects in iRODS, with data-info calling out to metadata for the pieces which concern it. I like this option because of the validation behavior -- anything for which a UUID or path should be validated goes through data-info, so it can just be trusted to verify that. Of course, there's nothing stopping other services from calling out to the metadata service directly even for iRODS data object-related tasks other than trusting ourselves not to. This would, obviously, keep this endpoint in data-info.

Another angle might be that the metadata service should do everything metadata-related, whether template/DB stuff or iRODS stuff, which would essentially work the opposite direction -- terrain would call out to metadata, which would call out to data-info for validating things exist and for operating on iRODS AVUs. This has some of the same issues that I'd say putting it in terrain does, with validation, where the order you call things in matters a lot. And, the metadata service is a fairly nice little wrapper right now and it feels like polluting it (but that might be because of my own predispositions that I feel that way). This tactic would pull this endpoint out of data-info, but it'd put it into metadata instead.

A third angle would be that both microservices should be kept as small and comprehensible as possible, which is pretty much the angle I've been taking (and, I think, what I've seen come out of discussions, but I'm not sure -- I've been taking this strategy as a mix of that impression plus the idea that it's the option which optimizes for getting jargon out of terrain faster, i.e. the only actual goal I'm strictly aware of). That is, don't put things in metadata or data-info which are composites of several other operations (except for bulk operations of some description -- I can't decide if a general bulk operations endpoint like the elasticsearch one makes sense or not, but at the very least endpoints for "do X on a bunch of targets"); limit yourself to simple operations and mostly straightforward REST. There's of course still vagueness there -- should data-info have CSV parsing, or should something higher-up just call a more general get-chunk endpoint and parse it itself (I've gone with putting the CSV parsing in data-info, thus far). This strategy would say that data-info is suitable for getting the iRODS AVUs and for writing the file to be created, but not for querying the metadata service and deciding how to combine and format the data, and is what I've gone with thus far.

The urlupload endpoint will also make for an interesting case of this. For that, I'd say that the first strategy would add the apps configuration and client into data-info and it'd all go there. The second wouldn't really be applicable, unless we decided anything that happens to use an app should go in apps. The third would say only the checking of the path's existence, user's existence, etc. should happen in data-info, or possibly even that data-info should only provide information and terrain should parse it out and decide if errors should be thrown (which is what I was asking about on your PR the other day).

psarando:
These changes seem like a straight forward refactor.
I didn't catch any obvious errors :shipit:

PR #55, opened 2016-01-29
CORE-6086: make clj-jargon.item-ops/copy-stream ownership-setting behavior configurable

ianmcorvidae:
Ticket title: You can steal ownership of files in a write-shared folder.

This was the case because both creating and overwriting files use `clj-jargon.item-ops/copy-stream`, which unconditionally called `set-owner` for the user making the request (which is to say, it allowed stealing ownership of any writeable file). This changes it to take an extra argument which can optionally turn off this behavior, and then uses this argument in `data-info.services.write` to explicitly specify when ownership should be granted (a new file) and shouldn't (modifying an existing one).


psarando:

PR #56, opened 2016-01-29
CORE-7472: Add heuristomancer and UI support for the SFF format.

ianmcorvidae:
The test data file is downloaded from the SRA; it seems to be the most common test file used for this sort of thing, as it appears on some various documentation pages around the web. However, this code does also work on the private, non-publishable sff file I was given to test with initially.


ianmcorvidae:
note: I used http://www.ncbi.nlm.nih.gov/Traces/trace.cgi?cmd=show&f=formats&m=doc&s=format#sff for the format details, if folks are reviewing this and want to understand the ebnf

tedgin:
Looks straightforward :shipit:

PR #57, opened 2016-01-29
Core 7470

dennisroberts71:
I resorted to using app categories to determine which apps should go into the `Shared with me` category. The additional action name degraded Grouper's performance too much, and the other methods that I could think of using the DE database weren't any better than this option.

### DIFF
--- a/libs/kameleon/src/kameleon/app_listing.clj
+++ b/libs/kameleon/src/kameleon/app_listing.clj
@@ -309,3 +309,39 @@
   (-> (get-app-listing-base-query workspace favorites-group-index query-opts)
       (add-public-apps-by-user-where-clause email)
       (select)))
+
+(defn visible-app-group-id-subselect
+  [workspace-id]
+  (subselect [:app_category_listing :c]
+             (fields :id)
+             (where (or {:is_public true}
+                        {:workspace_id workspace-id}))))
+
+(defn- is-visible-app-subselect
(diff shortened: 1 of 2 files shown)


libs/kameleon/src/kameleon/app_listing.clj
```diff
@@ -309,3 +309,32 @@
...
+  [{workspace-id :id root-category-id :root_category_id} favorites-group-index app-id-keyword]
+  (subselect [:app_category_app :aca]
+             (join [:app_category_listing :ac] {:aca.app_category_id :ac.id})
+             (where {:aca.app_id app-id-keyword})
+             (where {:ac.id [not= (get-fav-group-id-subselect root-category-id favorites-group-index)]})
+             (where (or :ac.is_public {:ac.workspace_id workspace-id}))))
+
+(defn list-shared-apps
+  "Lists apps that have been shared with a user. For the time being, this works by listing all apps
+  in th :app-ids parametr that are not in one of the categories that the user can access. When the
```
ianmcorvidae: nearly irrelevantly small but may as well spell out 'the' and 'parameter' with all their vowels

dennisroberts71: Heh, I don't seem to recll that. :laughing:

ianmcorvidae:
Aside from the pressing vowel-theft issue I see you've already resolved, this looks straightforward :laughing:

:shipit:

PR #58, opened 2016-01-30
CORE-7467 Store Permanent ID with completed Requests.

psarando:
On successful completion, terrain now sends the new ID with the completion status, and the EZID API response is emailed to data curators with the completion notification.
Permanent ID Request details now include the original_path and a permanent_id for completed requests.

### DIFF
--- a/services/terrain/src/terrain/services/permanent_id_requests.clj
+++ b/services/terrain/src/terrain/services/permanent_id_requests.clj
@@ -259,16 +259,21 @@
           {:keys [irods-avus metadata]} (metadata-get user folder-id)
           ezid-metadata                 (parse-valid-ezid-metadata irods-avus metadata)
           ezid-response                 (ezid/mint-id shoulder ezid-metadata)
+          identifier                    (get ezid-response (keyword type))
           publish-path                  (publish-data-item user folder (ezid-response->avus ezid-response))]
-      (email/send-permanent-id-request-complete type publish-path))
+      (email/send-permanent-id-request-complete type
+                                                publish-path
+                                                (json/encode ezid-response {:pretty true}))
+      identifier)
     (catch Object e
       (log/error e)
       (update-permanent-id-request request-id nil (json/encode {:status status-code-failed}))
-      (throw+ e)))
-  (update-permanent-id-request request-id nil (json/encode {:status status-code-completion})))
+      (throw+ e))))
 
 (defn create-permanent-id
   [request-id params body]
   (create-publish-dir)
-  (complete-permanent-id-request (:shortUsername current-user)
-                                 (admin-get-permanent-id-request request-id nil)))
+  (let [identifier (complete-permanent-id-request (:shortUsername current-user)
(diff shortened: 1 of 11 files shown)


services/terrain/src/terrain/services/permanent_id_requests.clj
```diff
@@ -259,16 +259,21 @@
           {:keys [irods-avus metadata]} (metadata-get user folder-id)
           ezid-metadata                 (parse-valid-ezid-metadata irods-avus metadata)
           ezid-response                 (ezid/mint-id shoulder ezid-metadata)
+          identifier                    (get ezid-response (keyword type))
           publish-path                  (publish-data-item user folder (ezid-response->avus ezid-response))]
-      (email/send-permanent-id-request-complete type publish-path))
+      (email/send-permanent-id-request-complete type
+                                                publish-path
+                                                (json/encode ezid-response {:pretty true}))
```
ianmcorvidae: I guess that the `identifiers` parameter in the email is this json-encoded response? What'll be in that response, other than the identifiers themselves (and should we perhaps update the variable name the template uses, so it's clearer it'll be more than just identifiers)?

ianmcorvidae:
Generally looks pretty reasonable, aside from my one (minor) question.

services/terrain/src/terrain/services/permanent_id_requests.clj
psarando: A successful response should only contain the new identifiers:
```json
{
  "DOI" : "10.5072/FK2RX9826N",
  "ARK" : "/b5072/fk2rx9826n"
}
```

psarando: Maybe we should return the unparsed EZID API response in the admin email, in case its format changes or our code doesn't parse the new ID correctly for some reason...

ianmcorvidae: I guess only one of these is stored in the DB, though? I guess both forms (hopefully) are added to the file metadata, so it's not going to cause issues in any case, and if they all look like this they're easy to convert between. Still curious though :)

ianmcorvidae:
Looks good to me, anyway. :shipit:

services/terrain/src/terrain/services/permanent_id_requests.clj
psarando: Yes, only the requested type is stored in the DB.

ianmcorvidae: I think you may be right about passing the unparsed response in the email, though I'm reluctant to say it since it's so nicely decoupled at present :)

psarando: Good point. I also forgot that we throw an error if we can't parse the response correctly, which should contain the unparsed response:
https://github.com/cyverse/DE/blob/dev/services/terrain/src/terrain/clients/ezid.clj#L76-L79

PR #59, opened 2016-02-01
CORE-7474 Open BED, BOWTIE, and GTF files in Tabular View

ash3rz:
As the title says, this is just allowing the BED, BOWTIE, and GTF file formats to open in Tabular View since they're tab-delimited file types.


ianmcorvidae:
:shipit: looks good to me

PR #60, opened 2016-02-01
Sharing

sri2k1us:
Please review my UI changes for App sharing


ui/de-lib/src/main/java/org/iplantc/de/analysis/client/presenter/AnalysesPresenterImpl.java
```diff
@@ -291,6 +295,11 @@ public void onShowAllSelected() {
     }
 
     @Override
+    public void onShareSelected(List<Analysis> selected) {
+        aSharingDialgoProvider.get().show(selected);
```
ianmcorvidae: Probably this should be "Dialog", even though it's just a variable name, rather than "Dialgo"

ui/de-lib/src/main/java/org/iplantc/de/analysis/client/views/AnalysisColumnModel.java
```diff
@@ -49,36 +50,69 @@
...
-            public void setValue(Analysis object, String value) {
-                object.setStatus(value);
-            }
-
-            @Override
-            public String getPath() {
-                return "status";
-            }
-        }, 75);
+                                                                       @Override
```
ianmcorvidae: I guess this huge indentation is our usual code style? I haven't worked with the java much obviously, but this seems... excessive

psarando: Our IntelliJ code style tries to align method arguments with the parentheses.

psarando: If the `new ColumnConfig<>(` was on a new line, its args wouldn't be indented as much.
I think our code style might also allow the args to start on the next line, which wouldn't indent them as much either.

ui/de-lib/src/main/java/org/iplantc/de/apps/client/presenter/sharing/AppSharingPresenter.java
```diff
@@ -0,0 +1,341 @@
...
+                this.usernames = usernames;
+            }
+
+            @Override
+            public void onFailure(Throwable caught) {
+                ErrorHandler.post(caught);
+            }
+
+            @Override
+            public void onSuccess(FastMap<Collaborator> results) {
```
ianmcorvidae: I think it has the potential to be confusing that a lot of the internal variables in this function reference data sharing and disk resources rather than apps; reading the source of parseNameFromPath it seems like that shouldn't be necessary here, either, but I may be misunderstanding what it does (but I'm not sure what a 'path' would constitute, for an app)

ui/de-lib/src/main/java/org/iplantc/de/apps/client/presenter/sharing/AppSharingPresenter.java
```diff
@@ -0,0 +1,341 @@
...
+    private JSONArray buildPathArr(List<Sharing> shareList) {
+        JSONArray pathArr = new JSONArray();
+        int index = 0;
+        for (Sharing s : shareList) {
+            pathArr.set(index++, new JSONString(s.getId()));
+        }
+        return pathArr;
+    }
+
+    private JSONArray buildPathArrWithPermissions(List<Sharing> shareList) {
```
ianmcorvidae: Similar to the other comment -- these aren't really path-related, they're arrays of apps. So the naming seems confusing.

ui/de-lib/src/main/java/org/iplantc/de/apps/client/presenter/toolBar/AppsToolbarPresenterImpl.java
```diff
@@ -130,4 +136,9 @@ public void onSuccess(String result) {
     public void onRequestToolSelected(RequestToolSelected event) {
         newToolRequestDialogProvider.get().show();
     }
+
+    @Override
+    public void onShareAppSelected(ShareAppsSelected event) {
+        appSharingDialgoProvider.get().show(event.getSelectedApps());
```
ianmcorvidae: El Diaglo rears his head again! :imp:

strootman:
Where are your tests?!

ui/de-lib/src/main/java/org/iplantc/de/analysis/client/presenter/sharing/AnalysisSharingPresenter.java
```diff
@@ -0,0 +1,94 @@
...
+    final AnalysisSharingView view;
+    private final SharingPermissionsPanel permissionsPanel;
+    private final List<Analysis> selectedAnalysis;
+    private Appearance appearance;
+    private FastMap<List<Sharing>> dataSharingMap;
+    private FastMap<List<JSONObject>> sharingList;
+    private final JsonUtil jsonUtil;
+    private final CollaboratorsUtil collaboratorsUtil;
+
+    public AnalysisSharingPresenter(final AnalysisServiceFacade aService,
```
strootman: No injection? We are trying to apply that paradigm across the UI.

ui/de-lib/src/main/java/org/iplantc/de/theme/base/client/apps/AppsMessages.properties
```diff
@@ -54,3 +54,6 @@ url = URL:
 appUrl = App URL
 copyAppUrl = Copy App URL
 copyAppSuccessMessage = ''{0}'' was successfully copied, and the copy is now open for editing.
+share = Share
+shareCollab = Share with Collaborators...
+sharePublic = Share with Public...
```
ianmcorvidae: This seems a bit stilted -- maybe "Make Public" or so? I think it doesn't matter much to the users that it's being made public by way of a sharing system, unlike with the collaborators where they'd probably go looking for something that says "share" on it.

ui/de-lib/src/main/java/org/iplantc/de/apps/client/presenter/sharing/AppSharingPresenter.java
sri2k1us: Agreed! I think these are result stealing logic from Data Sharing.

ianmcorvidae:
I can't comment too much on the code, having not worked with this much, so I think I've done the review I can and others should make sure there's no logic bugs to be concerned about. Hopefully my few nitpicks about style and variable names can help the code be more easily read though.

ui/de-lib/src/main/java/org/iplantc/de/client/sharing/SharingPresenter.java
```diff
@@ -0,0 +1,20 @@
+package org.iplantc.de.client.sharing;
```
strootman: I'd prefer that this code be kept in `org.iplantc.de.sharing.client`. Just because the package has this pattern does not mean that the code underneath it *MUST* be a GWT module. If we end up having more sharing views, widgets, presenters, etc which need to be used across the app, it would make more sense for it to be in a namespace which matches the rest of our app.

ui/de-lib/src/main/java/org/iplantc/de/analysis/client/presenter/sharing/AnalysisSharingPresenter.java
strootman: Next time, just open the PR earlier, instead of the day before. You can do rolling reviews on your code as you develop it. 
If you would have opened this up for review earlier, I would have just pounded out the beginning of a test structure for your code and given it to you to build off of.
I don't think there is a lot of time to do that right now, unfortunately. 
:(

ui/de-lib/src/main/java/org/iplantc/de/analysis/client/views/dialogs/AnalysisSharingDialog.java
```diff
@@ -0,0 +1,106 @@
...
+
+    private SharingPresenter sharingPresenter;
+
+    @Inject
+    CollaboratorsUtil collaboratorsUtil;
+    @Inject
+    JsonUtil jsonUtil;
+
+    @Inject
+    AnalysisSharingDialog(final AnalysisServiceFacade analysisService) {
```
strootman: Here, you could inject `AnalysisSharingPresenter`, and assign it to your `SharingPresenter` variable.
Same holds for the `DataSharingDialog` and `DataSharingPresenter`.

ui/de-lib/src/main/java/org/iplantc/de/theme/base/client/apps/AppsMessages.properties
sri2k1us: Agreed. Will change to "Make Public..."

ui/de-lib/src/main/java/org/iplantc/de/client/services/AppUserServiceFacade.java
```diff
@@ -82,4 +82,12 @@ void rateApp(App app,
                  AsyncCallback<AppFeedback> callback);
 
     void deleteRating(App app, AsyncCallback<AppFeedback> callback);
+
+    void getPermissions(List<App> apps, AsyncCallback<String> callback);
+
+    void shareApp(JSONObject request, AsyncCallback<String> callback);
+    
+    void unshareApp(JSONObject request, AsyncCallback<String> callback);
```
strootman: The `shareApp` and `unshareApp` method signatures are the definition of ambiguous. Couple that with an absence of documentation, and you have no idea what is supposed to be in that `JSONObject`.

Also, nothing seems to be done with the reponses.

So, I suggest the following changes
`void shareApp(Map<String, List<Sharing>> userSharingMap, AsyncCallback<Void> callback);`
`void unshareApp(Map<String, List<Sharing>> userUnsharingMap, AsyncCallback<Void> callback);`
Then, you remove the `AppSharingPresenter.buildSharingJson()` and `AppSharingPresenter.buildUnSharingJson()` logic out of `AppSharingPresenterImpl` and into `AppUserServiceFacadeImpl`.

We've been trying to migrate that sort of json manipulation/construction out of the presenters and into the service facades anyway.

strootman: This refactor should also be applied to `DiskResourceServiceFacade.shareDiskResource(...)` and `DiskResourceServiceFacade.unshareDiskResource(...)`
Change the signatures to;
`void shareDiskResource(Map<String>, List<Sharing>> userSharingMap, AsyncCallback<Void> callback);`
`void unshareDiskResource(Map<String>, List<Sharing>> userUnsharingMap, AsyncCallback<Void> callback);`

ui/de-lib/src/main/java/org/iplantc/de/client/services/AppUserServiceFacade.java
```diff
@@ -82,4 +82,12 @@ void rateApp(App app,
                  AsyncCallback<AppFeedback> callback);
 
     void deleteRating(App app, AsyncCallback<AppFeedback> callback);
+
+    void getPermissions(List<App> apps, AsyncCallback<String> callback);
```
strootman: It seems like this method should be updated as well;
`void getPermissions(List<App> apps, AsyncCallback<Map<String, List<Sharing>> callback);`
Then the logic in the `AppSharingPresenter.LoadPermissionsCallback.GetUserInfoCallback.onSuccess()` can be moved into a callback converter in `AppUserServiceFacadeImpl.getPermissions(..)`.

This also holds true for `DiskResourceServiceFacade.getPermissions(..)` except a little more;
`void getPermissions(HasPaths diskResourceIds, AsyncCallback<Map<String, List<Sharing>> callback);`
And the logic to construct the request will come from `DataSharingPresenterImpl.buildPermissionsRequestBody()`.

ui/de-lib/src/main/java/org/iplantc/de/client/models/sharing/Sharing.java
```diff
@@ -9,29 +14,67 @@
  */
 public class Sharing {
```
strootman: If you are going to use a class like this, it would be better to implement as a [Javascript Overlay Type](http://www.gwtproject.org/doc/latest/DevGuideCodingBasicsOverlay.html). It's not very clear in that documentation how to serialize the json into the overlay type, but you can find a good example in the _"Working with JSON"_ section, [Parsing JSON](http://www.gwtproject.org/doc/latest/DevGuideCodingBasicsJSON.html#parsing).

ui/de-lib/src/main/java/org/iplantc/de/apps/client/views/toolBar/AppsViewToolbarImpl.java
```diff
@@ -44,336 +58,456 @@
 /**
  * @author jstroot
  */
-public class AppsViewToolbarImpl extends Composite implements AppsToolbarView {
```
strootman: Great googly moogly!!! We *SERIOUSLY* need to get our IDE styles synced up!!!

ui/de-lib/src/main/java/org/iplantc/de/analysis/client/views/AnalysisColumnModel.java
```diff
@@ -108,22 +143,26 @@ public String getPath() {
     }
 
     @Override
-    public HandlerRegistration addAnalysisAppSelectedEventHandler(AnalysisAppSelectedEvent.AnalysisAppSelectedEventHandler handler) {
+    public HandlerRegistration
```
strootman: This PR has soo many unnecessary style changes.

ui/de-lib/src/main/java/org/iplantc/de/analysis/client/presenter/sharing/AnalysisSharingPresenter.java
ash3rz: I presume it has to do with the `TODO`s but a few of the variables defined here are not used.

ui/de-lib/src/main/java/org/iplantc/de/apps/client/presenter/sharing/AppSharingPresenter.java
```diff
@@ -0,0 +1,341 @@
...
+        public void onFailure(Throwable caught) {
+            permissionsPanel.unmask();
+            ErrorHandler.post(caught);
+        }
+
+        @Override
+        public void onSuccess(String result) {
+            JSONArray permissionsArray =
+                    jsonUtil.getArray(jsonUtil.getObject(result), "apps"); //$NON-NLS-1$
+            if (permissionsArray != null) {
```
ash3rz: I'm probably missing something, but instead of String couldn't this be a type of bean like `App` or `AppList`?  It would eliminate the need for all the JSONArrays and JSONObject manipulations in this file.

strootman: I made some comments in this direction.

ui/de-lib/src/main/java/org/iplantc/de/client/models/sharing/Sharing.java
sri2k1us: Not sure if we want to use Javascript Overlay type here. Sharing is just POJO that i am using store and retrieve information throughout the sharing logic That's never parsed into JSON. That being said I might need a new Model just for encode and decode the information that I send over wire.

ui/de-lib/src/main/java/org/iplantc/de/analysis/client/views/AnalysisColumnModel.java
sri2k1us: Not sure if that style changes were intentional. Guess me switching to IDEA and importing styles changed formatting in quite few places.

ui/de-lib/src/main/java/org/iplantc/de/apps/client/presenter/sharing/AppSharingPresenter.java
sri2k1us: Agreed. Most of the comments points towards a general refactoring that needs to done throughout Data and app Sharing. In the first pass, I have just tried to extend Data sharing logic to do app sharing. I will get more refactoring done for the next release as this is the last drop for this one.

ui/de-lib/src/main/java/org/iplantc/de/analysis/client/presenter/sharing/AnalysisSharingPresenter.java
sri2k1us: Analysis sharing code is not complete yet as there are no services to talk to. I left UI changes in place and just commented out the menu items for now.

ui/de-lib/src/main/java/org/iplantc/de/client/models/sharing/Sharing.java
strootman: This class is already used to serialize json data. In the `SharingPresenter` callbacks. Along with my other comments above, serializing/deserializing of this class can be performed as-is in the service layer. When that happens, I think the (de)serializing logic can be simplified with the use of turning this class into an overlay type.

PR #62, opened 2016-02-02
CORE-7473 Add metadata template AVUs from completed Permanent ID Request info.

psarando:
Refactored the /permanent-id-requests/:request-id/ezid endpoint to save
the new identifiers as metadata template AVUs instead of iRODS AVUs.
Also saves the current year as the publicationYear attribute, which is
also submitted with the EZID API request metadata.

### DIFF
--- a/services/terrain/src/terrain/services/permanent_id_requests.clj
+++ b/services/terrain/src/terrain/services/permanent_id_requests.clj
@@ -20,9 +20,23 @@
 (def ^:private status-code-completion "Completion")
 (def ^:private status-code-failed "Failed")
 
+(def ^:private ezid-target-attr "_target")
+
 (defn- parse-service-json
   [response]
   (-> response :body service/decode-json))
+
+(defn- format-staging-path
+  [path]
+  (ft/path-join (config/permanent-id-staging-dir) (ft/basename path)))
+
+(defn- format-publish-path
+  [path]
+  (ft/path-join (config/permanent-id-publish-dir) (ft/basename path)))
+
+(defn- format-metadata-target-url
+  [path]
+  (str (ft/rm-last-slash (config/permanent-id-target-base-url)) (format-publish-path path)))
 
 (defn- validate-ezid-metadata
   [ezid-metadata]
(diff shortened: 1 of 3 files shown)


services/terrain/src/terrain/services/permanent_id_requests.clj
```diff
@@ -28,6 +29,12 @@
   (when (empty? ezid-metadata)
     (throw+ {:type :clojure-commons.exception/bad-request
              :error "No metadata found for Permanent ID Request."}))
+  (let [identifier (get ezid-metadata (config/permanent-id-identifier-attr))]
+    (when-not (empty? identifier)
+    (throw+ {:type :clojure-commons.exception/bad-request
+             :error (str "The " (config/permanent-id-identifier-attr)
+                         " metadata attribue already contains a value: "
```
ianmcorvidae: s/attribue/attribute/

services/terrain/src/terrain/services/permanent_id_requests.clj
```diff
@@ -257,13 +300,17 @@
...
           ezid-response                 (ezid/mint-id shoulder ezid-metadata)
           identifier                    (get ezid-response (keyword type))
-          publish-path                  (publish-data-item user folder (ezid-response->avus ezid-response))]
+          alt-identifiers               (dissoc ezid-response (keyword type))
+          publish-avus                  (format-publish-avus metadata ezid-metadata identifier alt-identifiers)
+          publish-path                  (publish-data-item user folder)]
       (email/send-permanent-id-request-complete type
                                                 publish-path
                                                 (json/encode ezid-response {:pretty true}))
+      (publish-metadata folder template-id publish-avus)
```
ianmcorvidae: This is minor but it seems like you shouldn't send the email saying it's done until you've published the metadata too.

psarando: Actually, this is intentional, because the email will contain the new identifier(s), so if something goes wrong when setting the metadata, the admin can manually set the metadata from the email.

ianmcorvidae: Ah, interesting. I still think it'd be better to be able to tell them as much, probably by wrapping `publish-metadata` in a try-catch block and putting something in the email if it appears to have failed.

ianmcorvidae:
Looks good in general; don't think it needs to be blocked by my comments or anything, just minor improvements. :shipit:

psarando:
I'm going to go ahead and merge this now so the build is ready in the morning, but feel free to leave any additional comments.

PR #66, opened 2016-02-02
CORE-6052 fix search to exclude-include trash using query parameter.

sri2k1us:
Please review my changes for changes.


ianmcorvidae:
This looks straightforward to me despite java-related verbosity, so I'd say :shipit:

psarando:
Looks good to me. Although I think we should avoid formatting code that we didn't update (except for imports that are auto-organized).

PR #67, opened 2016-02-03
CORE-7266 address PR comments. Refactor code use auto beans instead o…

sri2k1us:
Addressed some PR comments.

### DIFF
--- a/ui/de-lib/src/main/java/org/iplantc/de/apps/client/presenter/sharing/AppSharingPresenter.java
+++ b/ui/de-lib/src/main/java/org/iplantc/de/apps/client/presenter/sharing/AppSharingPresenter.java
@@ -5,10 +5,20 @@ package org.iplantc.de.apps.client.presenter.sharing;
 
 import org.iplantc.de.apps.client.views.sharing.AppSharingView;
 import org.iplantc.de.client.models.apps.App;
+import org.iplantc.de.client.models.apps.AppAutoBeanFactory;
+import org.iplantc.de.client.models.apps.sharing.AppPermission;
+import org.iplantc.de.client.models.apps.sharing.AppSharingAutoBeanFactory;
+import org.iplantc.de.client.models.apps.sharing.AppSharingRequest;
+import org.iplantc.de.client.models.apps.sharing.AppSharingRequestList;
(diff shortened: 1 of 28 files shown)


ui/de-lib/src/main/java/org/iplantc/de/apps/client/presenter/sharing/AppSharingPresenter.java
```diff
@@ -82,24 +89,21 @@ public void onFailure(Throwable caught) {
...
-                    loadPermissions(id,appName, user_arr);
-                }
-
-                final List<String> usernames = new ArrayList<>();
-                usernames.addAll(sharingList.keySet());
-                collaboratorsUtil.getUserInfo(usernames, new GetUserInfoCallback(usernames));
+            AutoBean<AppUserPermissionsList> abrp =
+                    AutoBeanCodex.decode(shareFactory, AppUserPermissionsList.class, result);
+            AppUserPermissionsList appPermsList = abrp.as();
+            sharingList = new FastMap<>();
```
ash3rz: The next refactor that could be done is moving some of these decoding steps to a callback converter so the service returns the `AppUserPermissionsList`, for example, directly instead of a String.  Not sure if refactoring the appropriate methods in `appService` to use a callback converter might result in refactors needed elsewhere though.

ianmcorvidae:
As a general question, this appears to touch only the app-sharing-related endpoints. Presumably the same sorts of refactoring can be applied to the data-sharing endpoints from which these were originally copied. Is that out of scope for this pull request/for your work on this, or should it be added here?

ui/de-lib/src/main/java/org/iplantc/de/apps/client/presenter/sharing/AppSharingPresenter.java
strootman: I think that moving the bean manipulation/building to the service layer is important, but not as part of the release. I don't think the work will be too intensive.

strootman:
I agree with @ianmcorvidae. This code was copy-pasta from the old data-sharing endpoints. If you are going to perform this refactoring here, it makes complete sense to do it to the original code (since it is functionally identical).

PR #68, opened 2016-02-04
Core 7485

dennisroberts71:
This PR implements app sharing and unsharing notifications.


services/apps/src/apps/clients/notifications.clj
```diff
@@ -27,6 +28,14 @@
              {:content-type :json
               :body (cheshire/encode m)}))
 
+(defn- guarded-send-notification
```
ianmcorvidae: not for this PR, but presumably we should use this for more things in this namespace too

dennisroberts71: :+1: yeah, I think it would make the rest of the namespace a little clearer.

services/apps/src/apps/clients/notifications/app_sharing.clj
```diff
@@ -0,0 +1,73 @@
...
+  {:grouped   {:singular "%2$s has %1$sd %5$d app with you."
+               :plural   "%2$s has %1$sd %5$d apps with you."}
+   :ungrouped {:singular "%2$s has %1$sd the following app with you: %4$s"
+               :plural   "%2$s has %1$sd the following apps with you: %4$s"}})
+
+(def failure-formats
+  {:grouped   {:singular "%5$d app could not be %1$sd with %3$s."
+               :plural   "%5$d apps could not be %1$sd with %3$s."}
+   :ungrouped {:singular "The following app could not be %1$sd with %3$s: %4$s"
+               :plural   "The following apps could not be %1$sd with %3$s: %4$s"}})
```
ianmcorvidae: ick, complicated numerically-indexed format strings. Once again not for this PR most likely, but it might be nice to make our own lives easier at some point in the future by using a better formatting tool, perhaps [clostache](https://github.com/fhd/clostache) which we already use for building kifshare URLs.

services/apps/src/apps/clients/notifications.clj
dennisroberts71: I also want to spend a little time separating the notification formatting from the sending. That namespace seems to be getting a little cluttered.

ianmcorvidae: :+1: sounds good to me!

services/apps/src/apps/clients/notifications/app_sharing.clj
dennisroberts71: Meh, yeah, a better formatting tool might be nice, but I still though that this was better than piecing the strings together using `(str)`

ianmcorvidae: (arguably, a full-fledged i18n library might be ideal, but I have no idea where proper i18n/l10n are on our roadmap, if they're anywhere)

ianmcorvidae: and yeah, agreed. I'm just being picky, like you can only be when reviewing someone else's code ;)

dennisroberts71: Yes, that would be nice. I don't think that they are on our roadmap right now.

ianmcorvidae:
nice-for-the-future comments aside this looks good to me :shipit:

services/apps/src/apps/clients/notifications/app_sharing.clj
dennisroberts71: :laughing:

dennisroberts71:
:+1: merging...

PR #69, opened 2016-02-04
CORE-6215: migrate /secured/filesystem/anon-files

ianmcorvidae:
The first commit removes an AVU-based hack that we used to use for revoking permissions on unsharing, but don't any more. Since the code was also written to circumvent a jargon bug at the time, it also used the units field as the attribute field, so it added very unuseful AVUs, and only to user home directories, which usually the AVUs can't even be seen on.

This also migrates the start of some sharing stuff, but that will probably be refactored as I migrate the actual sharing endpoints, since I'd like to make those more file-centric. Currently the function is written to take only one permission, a list of users, and a list of files, and it assigns each user that permission on each file. With a future PR, I'll most likely change this to expect a single file (per invocation) and a list of user-permission pairs.

### DIFF
--- a/services/data-info/src/data_info/routes.clj
+++ b/services/data-info/src/data_info/routes.clj
@@ -14,6 +14,7 @@
             [data-info.routes.users :as users-routes]
             [data-info.routes.navigation :as navigation-routes]
             [data-info.routes.rename :as rename-routes]
+            [data-info.routes.sharing :as sharing-routes]
             [data-info.routes.status :as status-routes]
             [data-info.routes.stats :as stat-routes]
             [data-info.routes.trash :as trash-routes]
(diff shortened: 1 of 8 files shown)


psarando:
Is there a particular reason the `share-withs` and `fpaths` arguments are swapped for the `share` and `share-paths` functions? Was this just copied that way from terrain?

Otherwise, I didn't catch any issues with this PR. :shipit:

ianmcorvidae:
Ah, right. I'd initially started refactoring the share function to be user-focused as mentioned, and switched the arguments to reflect that. But, I pulled out most of those changes. I'll put them back the previous order tomorrow just to make the steps of the process more distinct.

PR #70, opened 2016-02-04
CORE-7481 Fix Permanent ID Request metadata validations.

psarando:
This validation fix should be required for the 2.4 release.


ianmcorvidae:
:shipit:, straightforward

PR #71, opened 2016-02-04
CORE-5291: After delete, return path in trash

ianmcorvidae:
This modifies only the data-info endpoints, but since terrain serves as a pure
passthrough for these endpoints, it will also be reflected in the terrain
endpoints.

One thing I'd like someone familiar with the UI to confirm or deny for me is if this will cause problems for the UI. This result (for all affected endpoints) is parsed with `decode(HasPaths.class, json)`, but I don't know if this will ignore unrecognized keys (in this case, the `"trash-paths"` key) or error. Obviously, I'm hoping the former!

### DIFF
--- a/services/data-info/src/data_info/services/trash.clj
+++ b/services/data-info/src/data_info/services/trash.clj
@@ -35,7 +35,8 @@
   [cm p user]
   (let [trash-path (randomized-trash-path user p)]
     (move cm p trash-path :user user :admin-users (cfg/irods-admins))
-    (set-metadata cm trash-path "ipc-trash-origin" p paths/IPCSYSTEM)))
+    (set-metadata cm trash-path "ipc-trash-origin" p paths/IPCSYSTEM)
+    trash-path))
 
 (defn- home-matcher
   [user path]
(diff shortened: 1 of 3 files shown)


ash3rz:
I can't speak for the services stuff, but regarding the UI I believe you're correct - the parser in that scenario would ignore the `trash-paths` key.

services/data-info/src/data_info/services/trash.clj
```diff
@@ -69,10 +71,13 @@
         ;;; If the file isn't already in the user's trash, move it there
         ;;; otherwise, do a hard delete.
         (if-not (.startsWith p (paths/user-trash-path user))
-          (move-to-trash cm p user)
+          (do (let [trash-path (move-to-trash cm p user)]
+              (reset! trash-paths
+                      (assoc @trash-paths p trash-path))))
```
dennisroberts71: You could use `(swap!)` here: `(swap trash-paths assoc p trash-path)`

ianmcorvidae: Ah, good idea. I'd just copied the way this code works from lower in the file for the restore functionality, which I'll probably change too.

dennisroberts71: :+1:

dennisroberts71:
This looks good to me. :shipit: when ready.

PR #73, opened 2016-02-05
Core 7485

dennisroberts71:
…tions

I *think* I caught all of the corner cases in this one. Making the code readable while catching all of the corner cases was a little challenging.


ianmcorvidae:
Looks reasonable to me :shipit:

dennisroberts71:
Thanks for the review. Merging the code now.

PR #74, opened 2016-02-08
CORE-7491 set context click property altering loading de properties f…

sri2k1us:
Please review my changes for setting context click option.


ui/de-webapp/src/main/java/org/iplantc/de/admin/client/Belphegor.java
```diff
@@ -6,6 +6,7 @@
 import org.iplantc.de.client.models.UserInfo;
 import org.iplantc.de.commons.client.ErrorHandler;
 import org.iplantc.de.commons.client.requests.KeepaliveTimer;
+import org.iplantc.de.conf.GwtRpcConfig;
```
psarando: This looks like an unused import.

sri2k1us: Bah! Yes. Thanks for catching that!

psarando:
:shipit:

PR #75, opened 2016-02-08
CORE-7490 fix data advanced search when search using owned by field.

sri2k1us:
Please review my changes to fix advanced search when searching on owned by field.


psarando:
:shipit:

PR #76, opened 2016-02-08
CORE-7495: fix a problem that was causing an index out of range error…

dennisroberts71:
… in jex-events


ianmcorvidae:
LGTM, pretty straightforward. If only it were as easy to find this sort of thing *before* it broke things, right?

:ship: :rocket: :guitar: :metal:

dennisroberts71:
Hehe, yep.

PR #77, opened 2016-02-08
Core 7365

strootman:
This PR removes some ansible group vars and a single, deprecated, config file.

This should be a non-functional change.

There is a corresponding branch in the private ansible repo as well.
Which is, again, removing some deprecated group vars and adding the systemd.services group var from this repo.


ianmcorvidae:
Looks straightforward. Probably doesn't need to go into `dev` but I don't think it'll hurt to put it there either. :ship:

strootman:
This doesn't have to go into `dev` right now, but it shouldn't hurt anything. 
However, the private repo stuff I'd prefer to go into `dev`, since the other work I have for `2.4` is dependent on it.

strootman:
@ianmcorvidae If you're satisfied that this is a non-functional change, should I go ahead and merge? Or should I close this PR and open against the `2.5` branch?

psarando:
Looks like some of these groups vars are still used by `ansible/roles/util-cfg-service/templates/logstash-forwarder/logstash-forwarder-conf.json.j2` and `ansible/roles/util-cfg-systemd-unit/defaults/main.yml`

psarando:
Looks good to me now :shipit:

strootman:
@psarando pushed change for `util-cfg-systemd-unit`

PR #78, opened 2016-02-09
Removed more elk references

strootman:
These should be the last of the elk references. It should also correct the `docker-gc.properties` build error.


psarando:
Looks good :shipit:

PR #83, opened 2016-02-09
CORE-4345 and CORE-4832

ash3rz:
CORE-4345 In the Notifications window, now if you hit the "Delete All" button, it will only delete all the notifications within the category you are currently viewing (Data, Analysis, etc).  If the "ALL" category is selected, then all notifications regardless of the category will get deleted.

CORE-4832 If you click either the "See All Notifications" or "See New Notifications" button in the Notification drop down menu, the Notification window will open and now the drop down will close itself instead of staying open.

### DIFF
--- a/ui/de-lib/src/test/java/org/iplantc/de/client/services/impl/MessageServiceFacadeImplTest.java
+++ b/ui/de-lib/src/test/java/org/iplantc/de/client/services/impl/MessageServiceFacadeImplTest.java
@@ -1,0 +1,94 @@
+package org.iplantc.de.client.services.impl;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotEquals;
+import static org.mockito.Mockito.when;
+
+import org.iplantc.de.client.models.DEProperties;
+import org.iplantc.de.client.models.UserInfo;
+import org.iplantc.de.client.models.notifications.NotificationAutoBeanFactory;
+import org.iplantc.de.client.models.notifications.NotificationCategory;
+import org.iplantc.de.client.util.DiskResourceUtil;
+import org.iplantc.de.shared.services.DiscEnvApiService;
+
+import com.google.gwt.user.client.rpc.AsyncCallback;
+import com.google.gwtmockito.GwtMockitoTestRunner;
+
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+
+/**
+ * @author aramsey
+ */
+@RunWith(GwtMockitoTestRunner.class)
+public class MessageServiceFacadeImplTest {
+
+    @Mock NotificationAutoBeanFactory notesFactoryMock;
+    @Mock DEProperties dePropertiesMock;
+    @Mock DiscEnvApiService deServiceFacadeMock;
+    @Mock UserInfo userInfoMock;
+    @Mock DiskResourceUtil diskResourceUtilMock;
+    @Mock AsyncCallback<String> asyncCallbackMock;
+
+    private MessageServiceFacadeImpl uut;
+
+    @Before
+    public void setUp() {
+
+        when(dePropertiesMock.getMuleServiceBaseUrl()).thenReturn("sampleUrl");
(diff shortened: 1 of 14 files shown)


dennisroberts71:
LGTM, but someone who's more familiar with the UI code should probably review it.

ash3rz:
Thanks Dennis!

strootman:
You can haz test?
https://github.com/cyverse/DE/blob/CORE-4345_4832/ui/de-lib/src/test/java/org/iplantc/de/desktop/client/presenter/DesktopPresenterImplTest.java

ash3rz:
Oops, I'll take a look at tests - thanks!

ui/de-lib/src/test/java/org/iplantc/de/desktop/client/views/DesktopNotifications_DesktopViewTest.java
```diff
@@ -38,9 +56,7 @@
     }
 
     @Test public void notificationsNotMarkedSeenWhenNotificationBtnSelectedWithGreaterThan10Unseen() {
-        DesktopViewImpl uut = new DesktopViewImpl(tourStringsMock, windowManagerMock);
-        verifyViewInit(uut);
-        final DesktopView.Presenter mockPresenter = mock(DesktopView.Presenter.class);
+
         uut.setPresenter(mockPresenter);
```
psarando: Is this already done in the `setUp`?

ash3rz:
I added a separate section for Notification tests and put them there.  The refactoring I started to put in here I've removed and I'll put into a separate PR for CORE-6341.

One issue is that it seems like the Notification category filtering that's enabled on `GET` isn't fully copied over to `DELETE`.  So, for example, if you're viewing only the `Permanent ID Requests` notifications and hit the Delete All button, nothing actually happens.  I'm not sure how involved it is to fix that, so I'm curious whether this would be a good first step into the backend side of things for me, or if I should file a separate ticket for that fix for someone else.

psarando:
I don't think a separate issue should be opened to add filtering to the DELETE endpoint. That seems like it's still part of the CORE-4345 issue. Or at least it could be a sub-task of CORE-4345.

ui/de-lib/src/test/java/org/iplantc/de/desktop/client/views/DesktopNotifications_DesktopViewTest.java
ash3rz: Ah, I missed this line.  I'll remove it - thanks for catching!

ash3rz:
Sure, makes sense.  Some of the filters are already there - `Data` and `Analysis` for example - so it's just a matter of getting the rest of the filters into the DELETE endpoint.  If it's not too difficult, I can try to take a stab at it.

ui/de-lib/src/test/java/org/iplantc/de/client/services/impl/MessageServiceFacadeImplTest.java
```diff
@@ -0,0 +1,94 @@
...
+        assertNotEquals(filter.toLowerCase(), NotificationCategory.ALL.toString().toLowerCase());
+        assertEquals(filter.toLowerCase(), NotificationCategory.ANALYSIS.toString().toLowerCase());
+    }
+
+    @Test
+    public void testDeleteAll_filterPermID() {
+        String filter = "PERMANENT ID REQUEST";
+        uut.deleteAll(filter, asyncCallbackMock);
+        assertNotEquals(filter.toLowerCase(), NotificationCategory.ALL.toString().toLowerCase());
+        assertEquals(filter.toLowerCase(), NotificationCategory.PERMANENTIDREQUEST.toString().toLowerCase());
```
psarando: I might be missing something with these tests.
It seems like a `filter` string is created, then it's just compared to the `NotificationCategory` enums as strings?
I see the `filter` is passed to a MessageServiceFacadeImpl#deleteAll call, but it doesn't look like any of the facade's functionality is tested.
Unless we're capturing or verifying calls into the `deServiceFacadeMock`, maybe we don't need these tests?

PR #86, opened 2016-02-10
Core 3484

dennisroberts71:


dennisroberts71:
This PR includes changes for CORE-3484 and CORE-3485.

databases/de-database-schema/src/main/constraints/46_genome_ref.sql
```diff
@@ -0,0 +1,15 @@
+SET search_path = public, pg_catalog;
+
+--
+-- Uniqueness constraint on reference genome name.
+--
+ALTER TABLE ONLY genome_reference
+ADD CONSTRAINT genome_ref_mame_unique
```
ianmcorvidae: It's just in the name of a constraint so it doesn't change functionality, but this says `mame`, not `name`

dennisroberts71: ah. thanks for catching that

services/apps/src/apps/metadata/reference_genomes.clj
```diff
@@ -83,18 +91,33 @@
     (sql/update genome_reference (set-fields update-values) (where {:id reference-genome-id}))
```
ianmcorvidae: I think that this function (`update-reference-genome`) should also validate that, if a new name or path is provided, it does not conflict with anything.

dennisroberts71: Good idea. I'll add that as well.

ianmcorvidae:
Aside from the comment about `update-reference-genome` this looks straightforward, so as far as I'm concerned once that's in, :shipit:

psarando:
Looks good :shipit:

PR #87, opened 2016-02-11
CORE-7386: remove support for the `public` query parmeter from GET /a…

dennisroberts71:
…dmin/apps/categories

For this issue, I decided to have terrain continue to ignore the `public` query parameter. We can make the endpoint definitions more formal when we migrate terrain to compojure-api. See the Jira issue for more details.


ianmcorvidae:
Seems simple enough. :ship:

dennisroberts71:
Cool, thanks for the review. Merging it now...

PR #88, opened 2016-02-11
CORE-7511 Add more user info to PermID Request details.

psarando:
Updated the responses of Permanent ID Request endpoints that return a request's details to include more user info in the 'requested_by' field, such as email and first/last name.
Since the requesting user's email is now available when notifications are sent to that user, notification emails can now be enabled.
Added the new ID to the "Completion" notification update in the "comments" field, which is now included as the body of the "Completion" notification email.

### DIFF
--- a/services/terrain/src/terrain/services/permanent_id_requests.clj
+++ b/services/terrain/src/terrain/services/permanent_id_requests.clj
@@ -10,6 +10,7 @@
             [terrain.clients.data-info :as data-info]
             [terrain.clients.data-info.raw :as data-info-client]
             [terrain.clients.ezid :as ezid]
+            [terrain.clients.iplant-groups :as groups]
             [terrain.clients.metadata.raw :as metadata]
             [terrain.clients.notifications :as notifications]
             [terrain.util.config :as config]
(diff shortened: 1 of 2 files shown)


services/terrain/src/terrain/services/permanent_id_requests.clj
```diff
@@ -263,6 +269,19 @@
       (dissoc :target_id :target_type :original_path)
       (assoc :folder (data-info/stat-by-uuid user (uuidify target_id)))))
 
+(defn- format-requested-by
+  [user {:keys [requested_by target_id] :as permanent-id-request}]
+  (let [user-info (groups/lookup-subject user requested_by)]
+    (if user-info
```
ianmcorvidae: This is, of course, minor, but may as well use `if-let` here.

services/terrain/src/terrain/services/permanent_id_requests.clj
```diff
@@ -302,25 +326,25 @@
...
+       (format-permanent-id-request-details (:shortUsername current-user))))
 
 (defn update-permanent-id-request
   [request-id params body]
   (let [response (->> (metadata/update-permanent-id-request request-id body)
                       parse-service-json
-                      (format-perm-id-req-response (:shortUsername current-user)))]
+                      (format-permanent-id-request-details (:shortUsername current-user)))]
     (send-update-notification response)
     response))
```
ianmcorvidae: This isn't something you're changing here, but am I missing something or is `params` not used in this function? If so, why is it included?

services/terrain/src/terrain/services/permanent_id_requests.clj
```diff
@@ -355,4 +379,5 @@
   (let [identifier (complete-permanent-id-request (:shortUsername current-user)
                                                   (admin-get-permanent-id-request request-id nil))]
     (update-permanent-id-request request-id nil (json/encode {:status       status-code-completion
+                                                              :comments     identifier
```
ianmcorvidae: This seems unrelated (though presumably fine) -- what's going on here adding the extra key with the same info?

ianmcorvidae: Oh, I see, this is what's going into the body of the email. Why not just use the pre-existing `:permanent_id` key, then?

services/terrain/src/terrain/services/permanent_id_requests.clj
psarando: Thanks, good catch.

services/terrain/src/terrain/services/permanent_id_requests.clj
psarando: No, `params` is not used. It was added at the beginning to all these functions when they were stubbed out in case they were needed, but the params never got used.

services/terrain/src/terrain/services/permanent_id_requests.clj
psarando: This status update is an automated `Completion` status update.

The `comments` field can be any string and is the comments of the status update, which is displayed in the request's status history and also used for the body of the status update's notification. So the comments also become the body of the notification's email.
Other status updates besides the `Completion` status can set any other arbitrary string for the comments, and those mainly come from the UI when an admin is updating the status of a request.

The `permanent_id` is the newly created ID. This field is specific to the `POST /admin/permanent-id-requests/{request-id}/status` metadata endpoint and is only sent once when the request is completed, then the metadata service saves the ID with the Permanent ID Request records. This endpoint will return an error if this field is populated when an ID is already saved for the Permanent ID Request.

ianmcorvidae: Ah, alright. Makes sense; I'd somehow missed other places setting this value.

ianmcorvidae:
Changes look good, I'd say. It wouldn't hurt to refactor to `if-let` or remove that unused function argument but it's also fine without that. :shipit:

PR #90, opened 2016-02-16
Template Metadata Indexing (ongoing)

ianmcorvidae:
This PR is up now in order to start getting feedback early, though the service does not yet support all of its features (nor is it yet set up to be configured or deployed or anything like that).

This PR adds a new service, which I've provisionally named `templeton`. This service performs tasks related to indexing of so-called template metadata, i.e., metadata which is stored in the 'metadata' database rather than as iRODS AVUs.

Ultimately, this service will run persistently in two instances running in two different modes: periodic, which listens for messages which trigger complete reindexing of the database (like Infosquito or monkey), and incremental, which listens for messages describing smaller changes and which update the index (like dewey). Right now, it only supports an option like Infosquito's `--once`, which crawls the database and updates the index once.

The first commit on this branch is a good place to review. The other two commits (at this time) just add vendored dependencies and make a minor documentation-only change to the UI code in anticipation of changes there in the future.


ianmcorvidae:
I'll leave comments noting what things I've largely pilfered from other services (jex-events and the new jobservices work) and what's new here.

ianmcorvidae:
The mapping definition for the type where this is indexing stuff is: 

```json
{
  "metadata" : {
    "properties" : {
      "id" : {
        "type"  : "string",
        "index" : "not_analyzed"
      },
      "metadata" : {
        "_comment"   : "this is an array",
        "type"       : "nested",
        "properties" : {
          "attribute" : {
            "type"     : "string",
            "analyzer" : "irods_entity"
          },
          "value" : {
            "type" : "string"
          },
          "unit" : {
            "type" :     "string",
            "analyzer" : "keyword"
          }
        }
      }
    }
  }
}
```

This is in a parallel branch in the private de-ansible repository where we define these entities. I've put it in the same index as other data, because that way I don't need to redefine the analyzers and such to keep these analyzed and indexed the same as iRODS AVUs are.

services/templeton/src/configurate/configurate_test.go
```diff
@@ -0,0 +1,87 @@
+package configurate
+
+import "testing"
+
```
ash3rz: I like that you have tests already!  Yay tests!

services/templeton/src/templeton/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,86 @@
...
+
+// NewElasticer returns a pointer to an Elasticer instance that has already tested its connection
+// by making a WaitForStatus call to the configured Elasticsearch cluster
+func NewElasticer(elasticsearchBase string, elasticsearchIndex string) (*Elasticer, error) {
+	c := elastigo.NewConn()
+
+	err := c.SetFromUrl(elasticsearchBase)
+	if err != nil {
+		return nil, err
+	}
```
ash3rz: (Knowing absolutely nothing about Go, so I'm asking out of curiosity) I see these same 3 lines for checking errors sprinkled in multiple times across multiple files, is there an easier way to handle this?

ianmcorvidae: As far as I can see, this is the idiomatic way to do it -- this exact sequence is literally just passing the same error up the chain. Go also lets you raise a panic (which is e.g. what the language itself does for indexes out of bounds) but that'd just change the piece inside the if, since I can't control how the libraries pass errors. I'm wondering if there's something nicer too, though, or at least more of an idea if I really need to be catching them all, what to do with them, etc. more than just passing them up the whole way and then logging them and dying. This particular one I think will just always be fatal, since it's not going to be able to do much with an elasticsearch base URL it can't parse, but I'm not sure that's generally true.

johnworth: Yeah, Go doesn't shy away from this sort of repetition. You could possibly shorten this chunk a bit like so:

```
    if err := c.SetFromURL(elasticsearchBase); err != nil {
        return nil, err
    }
```
The amount of space it saves is so little that it doesn't really matter which way you do it, IMO.

ianmcorvidae:
Another bit I'd love feedback on is ES library choice. I've used https://godoc.org/github.com/mattbaird/elastigo/lib because it looked the most idiomatic. Other options include https://godoc.org/github.com/belogik/goes which looks sort of anemic to me and https://godoc.org/github.com/olivere/elastic which seems very complete but operates by some sort of chained-method-call-for-configuration system where you create a client and then do something like `client.Ping().Pretty(true).Do()`, which seemed weird. elastigo also has a nice interface for the bulk indexing where you fire things off and it handles when to package them up for a bulk request in the background. But, I'm not sure exactly how it deals with errors or how much I should care about it.

ianmcorvidae:
Closing this to reopen against dev.

PR #91, opened 2016-02-17
Index template metadata periodically (CORE-7496)

ianmcorvidae:
This PR adds a new service, `templeton`, which indexes template metadata for search. It implements, for now, only a periodic reindexing setup, similar to `Infosquito`/`monkey`.

It is not currently set up to be deployed; the elasticsearch indexes will need to be fixed before indexing can begin, work which is pending. It is, however, ready to be reviewed in principle.

There is also an `index-template-metadata` branch in the private ansible repository, which contains the mappings which will be used in the updated indexes. The [`core-7499`](https://github.com/cyverse/DE/compare/dev...core-7499) branch has updates to the UI's search queries, which are not currently up for review due to the same need for elasticsearch index recreation.

See previous comments in https://github.com/cyverse/DE/pull/90 if interested.


ianmcorvidae:
Status update: at this point the `--full` indexing mode is done. Next steps are AMQP support/`--periodic` mode (which reuses a lot with `--full`) and updating the queries used by the DE to search the new type.

ianmcorvidae:
Update: --periodic mode is now done, bringing this to a state where it actually does something productive!

services/templeton/README.md
```diff
@@ -0,0 +1,20 @@
...
+Templeton is a service which indexes template metadata, as stored in the
+`metadata` database and interacted with ordinarily through the `metadata`
+service.
+
+Templeton operates in three modes, controlled by the `--mode` command-line flag:
+
+| Mode          | Operation |
+| ------------- | --------- |
+| `periodic`    | listens on a configured AMQP queue and recieves only `index.templates` and `index.all` messages to trigger complete reindexes of the metadata template information (similar to `Infosquito` and `monkey`) |
+| `incremental` | listens on a configured AMQP queue and recieves messages describing incremental changes to be indexed (similar to `dewey`) |
```
ianmcorvidae: (this mode is not implemented at this time)

services/templeton/src/logcabin/logcabin.go
```diff
@@ -0,0 +1,76 @@
+package logcabin
```
ianmcorvidae: This package has not been updated for several iterations of changes by @johnworth and @strootman. Since it works, for now I haven't updated it, but it's on my roadmap a bit after the elasticsearch index wrangling process.

services/templeton/src/messaging/amqp.go
```diff
@@ -0,0 +1,216 @@
...
+type MessageHandler func(amqp.Delivery)
+
+type aggregationMessage struct {
+	handler  MessageHandler
+	delivery *amqp.Delivery
+}
+
+type consumer struct {
+	exchange     string
+	exchangeType string
```
ianmcorvidae: Allowing the exchange type to be configured here is a modification in this service, since the `de` exchange used for the full-reindexing messages is a `direct` exchange, not a `topic` exchange.

services/templeton/src/templeton/database/database.go
```diff
@@ -0,0 +1,200 @@
...
+	if err != nil {
+		return nil, err
+	}
+	databaser := &Databaser{
+		db:         db,
+		ConnString: connString,
+	}
+	return databaser, nil
+}
+
```
ianmcorvidae: The stuff above here is pretty much copied from jobservices/jex-go, and the stuff below here is new in this service.

services/templeton/src/templeton/main.go
```diff
@@ -0,0 +1,177 @@
...
+
+		// Accept and handle messages sent out with the index.all and index.templates routing keys
+		client.AddConsumer(messaging.ReindexExchange, "direct", "templeton.reindexAll", messaging.ReindexAllKey, func(del amqp.Delivery) {
+			es.Reindex(d)
+			del.Ack(false)
+		})
+		client.AddConsumer(messaging.ReindexExchange, "direct", "templeton.reindexTemplates", messaging.ReindexTemplatesKey, func(del amqp.Delivery) {
+			es.Reindex(d)
+			del.Ack(false)
+		})
```
ianmcorvidae: A modification/improvement to this would presumably be to listen to both types of messages on one queue. I don't think this is required initially, but I could be convinced.

services/templeton/src/templeton/main.go
```diff
@@ -0,0 +1,177 @@
...
+
+		// spinner in order to keep the program running since client.Listen() is in a goroutine.
```
ianmcorvidae: This is copied out of how jobservices works in the one place it listens to messages but does not publish any.

services/templeton/src/templeton/main.go
```diff
@@ -0,0 +1,177 @@
...
+		}
+	}
+
+	if *mode == "incremental" {
+		logger.Println("Incremental indexing mode selected.")
+
+		// TODO: AMQP listener triggering incremental updates
+		return
+	}
+}
```
tedgin: I don't know about Go coding conventions, so I might be off base here.

I suggest breaking each mode out into its own function. That would make the largish main function easier to understand.

ianmcorvidae: Seems like a good idea, yeah. I'll do that.

services/templeton/src/configurate/configurate_test.go
```diff
@@ -0,0 +1,87 @@
...
+	if actual != expected {
+		t.Errorf("The elasticsearch.base was %s instead of %s", actual, expected)
+	}
+}
+
+func TestESIndex(t *testing.T) {
+	err := configurator()
+	if err != nil {
+		t.Error(err)
+		t.Fail()
```
johnworth: Something that was pointed out in my code is that following up a t.Error(err) with a t.Fail() is redundant. I was under the impression that t.Fail() caused a test case to fail immediately, which does not appear to be the case.

johnworth: Also, this should not prevent the code from being merged. Forgot to mention that.

ianmcorvidae: Easy fix though. Mostly in code copied from you anyway, heh.

services/templeton/src/templeton/database/database.go
```diff
@@ -0,0 +1,200 @@
...
+type objectCursor struct {
+	rows     *sql.Rows
+	lastRow  *model.AVURecord
+	moreRows bool
+	anyRows  bool
+}
+
+func newObjectCursor(rows *sql.Rows) *objectCursor {
+	return &objectCursor{rows: rows, lastRow: &model.AVURecord{TargetId: ""}, moreRows: true, anyRows: false}
```
johnworth: Might want to split the objectCursor initialization into multiple rows. It's a bit difficult to visually parse as it is.

johnworth: Again, not something that should prevent a merge.

PR #93, opened 2016-02-18
Core 7524

dennisroberts71:
This is a hotfix to production to ensure that new users are added to the de-users group.

### DIFF
--- a/services/apps/src/apps/service/workspace.clj
+++ b/services/apps/src/apps/service/workspace.clj
@@ -1,9 +1,11 @@
 (ns apps.service.workspace
   (:use [korma.db :only [transaction]])
-  (:require [apps.persistence.workspace :as wp]))
+  (:require [apps.clients.iplant-groups :as iplant-groups]
+            [apps.persistence.workspace :as wp]))
 
 (defn get-workspace
-  [{:keys [username]}]
+  [{short-username :shortUsername :keys [username]}]
+  (iplant-groups/add-de-user short-username)
   (if-let [workspace (wp/get-workspace username)]
     (assoc workspace :new_workspace false)
     (assoc (wp/create-workspace username) :new_workspace true)))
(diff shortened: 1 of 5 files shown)


services/apps/src/apps/service/workspace.clj
```diff
@@ -1,9 +1,11 @@
 (ns apps.service.workspace
   (:use [korma.db :only [transaction]])
-  (:require [apps.persistence.workspace :as wp]))
+  (:require [apps.clients.iplant-groups :as iplant-groups]
+            [apps.persistence.workspace :as wp]))
 
 (defn get-workspace
-  [{:keys [username]}]
+  [{short-username :shortUsername :keys [username]}]
+  (iplant-groups/add-de-user short-username)
```
psarando: Would this make more sense to have in the `bootstrap` endpoint rather than in the `get-workspace` function?
It makes sense to get-or-create the workspace in this function, but I wouldn't think to look here when trying to figure out where users are added to `iplant-groups`.

psarando: Ah, but this is the main function of the `workspaces` endpoint, and the `bootstrap` endpoint is in `terrain`.
Does `terrain` have an `iplant-groups` client? Or is that all contained in the `apps` service?

dennisroberts71: Yeah, that's a good point. I put it here because it seemed somewhat related to creating a workspace. (No apps will be listed if the user either doesn't have a workspace or isn't in the de-users group.) Terrain doesn't currently have an iplant-groups client, but one could be added. I could also create a new endpoint in apps that ensures that the user is in the appropriate group. I think I actually prefer the latter option because the apps service has to know the appropriate group name anyway.

psarando: OK, that makes sense.
We can leave it where it is for now, then add the new endpoint when we refactor the `bootstrap` endpoint.
So this PR looks good to me :shipit:

dennisroberts71: Cool. I'll create a Jira issue to ensure that this doesn't fall through the cracks.

PR #94, opened 2016-02-18
CORE-4345 & CORE-4832

ash3rz:
CORE-4345 In the Notifications window, now if you hit the "Delete All" button, it will only delete all the notifications within the category you are currently viewing (Data, Analysis, etc). If the "ALL" category is selected, then all notifications regardless of the category will get deleted.

CORE-4832 If you click either the "See All Notifications" or "See New Notifications" button in the Notification drop down menu, the Notification window will open and now the drop down will close itself instead of staying open.

### DIFF
--- a/ui/de-lib/src/main/java/org/iplantc/de/client/services/impl/MessageServiceFacadeImpl.java
+++ b/ui/de-lib/src/main/java/org/iplantc/de/client/services/impl/MessageServiceFacadeImpl.java
@@ -142,11 +142,11 @@ public class MessageServiceFacadeImpl implements MessageServiceFacade {
     }
 
     @Override
-    public void deleteAll(String filter, AsyncCallback<String> callback) {
+    public void deleteAll(NotificationCategory category, AsyncCallback<String> callback) {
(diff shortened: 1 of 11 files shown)


ui/de-lib/src/main/java/org/iplantc/de/client/services/impl/MessageServiceFacadeImpl.java
```diff
@@ -141,9 +142,13 @@ public void getMessageCounts(final AsyncCallback<Counts> callback) {
     }
 
     @Override
-    public void deleteAll(AsyncCallback<String> callback) {
+    public void deleteAll(NotificationCategory category, AsyncCallback<String> callback) {
         String address = deProperties.getMuleServiceBaseUrl() + "notifications/delete-all"; //$NON-NLS-1$
 
+        if (!category.equals(NotificationCategory.ALL)) {
```
psarando: I would invert this `equals` call, just in case `category` is `null`, then an NPE won't get thrown.
Although with an enum you could just use `category != NotificationCategory.ALL`, which is null-safe as well.

psarando:
Looks good! :shipit:

PR #96, opened 2016-02-22
Minor clj-icat-direct optimizations

ianmcorvidae:
All fairly minor stuff, @psarando work the other day hit the big issues. But, reducing these column counts and extra CTEs should make these queries run slightly faster due to reduced memory usage and less query planning (respectively), so I figured I'd put it up.

### DIFF
--- a/libs/clj-icat-direct/src/clj_icat_direct/queries.clj
+++ b/libs/clj-icat-direct/src/clj_icat_direct/queries.clj
@@ -517,21 +517,19 @@
 
 (def queries
   {:count-all-items-under-folder
-   "WITH user_groups AS ( SELECT g.*
+   "WITH user_groups AS ( SELECT g.group_user_id
                             FROM r_user_main u
                             JOIN r_user_group g ON g.user_id = u.user_id
                            WHERE u.user_name = ?
                              AND u.zone_name = ? ),
 
-         parent      AS ( SELECT * from r_coll_main
+         parent      AS ( SELECT DISTINCT coll_id, coll_name from r_coll_main
                            WHERE coll_name = ?
-                          UNION
-                          SELECT * from r_coll_main
-                           WHERE coll_name LIKE ? || '/%' ),
+                              OR coll_name LIKE ? || '/%' ),
 
-         data_objs   AS ( SELECT *
+         data_objs   AS ( SELECT data_id
                             FROM r_data_main
-                           WHERE coll_id = ANY(ARRAY( SELECT coll_id FROM parent )))
+                           WHERE coll_id = ANY(ARRAY( SELECT coll_id FROM parent )) )
 
     SELECT ((SELECT count(DISTINCT d.data_id) FROM r_objt_access a
                JOIN data_objs d ON a.object_id = d.data_id
@@ -545,13 +543,10 @@
                 AND c.coll_type != 'linkPoint')) AS total"
 
    :list-folders-in-folder
-   "WITH user_groups AS ( SELECT g.* FROM r_user_main u
+   "WITH user_groups AS ( SELECT g.group_user_id FROM r_user_main u
                             JOIN r_user_group g ON g.user_id = u.user_id
                            WHERE u.user_name = ?
                              AND u.zone_name = ? ),
-
-         parent      AS ( SELECT * from r_coll_main
-                           WHERE coll_name = ? )
 
     SELECT DISTINCT
            c.parent_coll_name                     as dir_name,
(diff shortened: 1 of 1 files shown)


libs/clj-icat-direct/src/clj_icat_direct/queries.clj
```diff
@@ -545,14 +543,11 @@
                 AND c.coll_type != 'linkPoint')) AS total"
 
    :list-folders-in-folder
-   "WITH user_groups AS ( SELECT g.* FROM r_user_main u
+   "WITH user_groups AS ( SELECT g.group_user_id FROM r_user_main u
                             JOIN r_user_group g ON g.user_id = u.user_id
                            WHERE u.user_name = ?
                              AND u.zone_name = ? ),
```
psarando: This trailing comma will cause a syntax error.

libs/clj-icat-direct/src/clj_icat_direct/queries.clj
```diff
@@ -565,28 +560,28 @@
...
 
-         parent      AS ( SELECT * from r_coll_main
+         parent      AS ( SELECT coll_id from r_coll_main
                            WHERE coll_name = ? ),
 
-         data_objs   AS ( SELECT *
+         data_objs   AS ( SELECT data_id
                             FROM r_data_main
-                           WHERE coll_id = ANY(ARRAY( SELECT coll_id FROM parent )))
+                           WHERE coll_id IN ( SELECT coll_id FROM parent ))
```
psarando: Why change `= ANY(ARRAY(...))` to `IN(...)` here but not above in the `count-all-items-under-folder` query?
Also, I was under the impression from @tedgin that `= ANY(ARRAY(...))` would be faster in these queries than `IN(...)`

ianmcorvidae: Ah, this is another oversight. I'd changed everything to `IN` originally for consistency, but I decided against that and changed one of them back (but I missed this one). As far as I know there's no performance difference between `IN (...)` and `= ANY(ARRAY(...))` (in this sort of a case they're both just producing a CTE scan and a filter) but I'd figured we could be consistent (and `IN` is better-known, hence the original choice, but we're more often using the array notation, hence changing my mind).

psarando:
Looks good! :shipit:

libs/clj-icat-direct/src/clj_icat_direct/queries.clj
tedgin: You wouldn't use `ANY(ARRAY())` in this case. There is only 1 parent folder.  `IN` in intent that `ANY(ARRAY())`.

libs/clj-icat-direct/src/clj_icat_direct/queries.clj
```diff
@@ -517,21 +517,19 @@
...
   {:count-all-items-under-folder
-   "WITH user_groups AS ( SELECT g.*
+   "WITH user_groups AS ( SELECT g.group_user_id
                             FROM r_user_main u
                             JOIN r_user_group g ON g.user_id = u.user_id
                            WHERE u.user_name = ?
                              AND u.zone_name = ? ),
 
-         parent      AS ( SELECT * from r_coll_main
+         parent      AS ( SELECT DISTINCT coll_id, coll_name from r_coll_main
```
tedgin: `coll_id` is unique in `r_coll_main`, so you don't need to use `DISTINCT`. It will only slow the query down as duplicates that don't exist will be searched for.

libs/clj-icat-direct/src/clj_icat_direct/queries.clj
ianmcorvidae: If it is, it's not enforced:

```
ICAT=# \d r_coll_main
                         Table "public.r_coll_main"
      Column      |          Type           |           Modifiers            
------------------+-------------------------+--------------------------------
 coll_id          | bigint                  | not null
 parent_coll_name | character varying(2700) | not null
 coll_name        | character varying(2700) | not null
 coll_owner_name  | character varying(250)  | not null
 coll_owner_zone  | character varying(250)  | not null
 coll_map_id      | bigint                  | default 0
 coll_inheritance | character varying(1000) | 
 coll_type        | character varying(250)  | default '0'::character varying
 coll_info1       | character varying(2700) | default '0'::character varying
 coll_info2       | character varying(2700) | default '0'::character varying
 coll_expiry_ts   | character varying(32)   | 
 r_comment        | character varying(1000) | 
 create_ts        | character varying(32)   | 
 modify_ts        | character varying(32)   | 
Indexes:
    "idx_coll_main2" UNIQUE, btree (parent_coll_name, coll_name)
    "idx_coll_main3" UNIQUE, btree (coll_name)
    "idx_coll_main1" btree (coll_id)
    "idx_coll_main_coll_type" btree (coll_type)
    "idx_coll_main_parent_coll_name" btree (parent_coll_name)

```

Only `coll_name` and `parent_coll_name, coll_name` are made unique.

libs/clj-icat-direct/src/clj_icat_direct/queries.clj
ianmcorvidae: If it is, it's not enforced:

```
ICAT=# \d r_coll_main
                         Table "public.r_coll_main"
      Column      |          Type           |           Modifiers            
------------------+-------------------------+--------------------------------
 coll_id          | bigint                  | not null
 parent_coll_name | character varying(2700) | not null
 coll_name        | character varying(2700) | not null
 coll_owner_name  | character varying(250)  | not null
 coll_owner_zone  | character varying(250)  | not null
 coll_map_id      | bigint                  | default 0
 coll_inheritance | character varying(1000) | 
 coll_type        | character varying(250)  | default '0'::character varying
 coll_info1       | character varying(2700) | default '0'::character varying
 coll_info2       | character varying(2700) | default '0'::character varying
 coll_expiry_ts   | character varying(32)   | 
 r_comment        | character varying(1000) | 
 create_ts        | character varying(32)   | 
 modify_ts        | character varying(32)   | 
Indexes:
    "idx_coll_main2" UNIQUE, btree (parent_coll_name, coll_name)
    "idx_coll_main3" UNIQUE, btree (coll_name)
    "idx_coll_main1" btree (coll_id)
    "idx_coll_main_coll_type" btree (coll_type)
    "idx_coll_main_parent_coll_name" btree (parent_coll_name)

```

Only `coll_name` and `parent_coll_name, coll_name` are made unique.

psarando: We should verify that. The same logic should have applied to the `SELECT count(DISTINCT c.coll_id)` in this query below, but when I tested it without the `DISTINCT` I got a higher count.

ianmcorvidae: @psarando it looks like in our actual databases, they are in fact unique. I think the query below needed it because of the join on `r_objt_access` which could result in multiple rows in the resultset.

It's still not enforced at a DB level, just higher up (presumably) in iRODS.

libs/clj-icat-direct/src/clj_icat_direct/queries.clj
```diff
@@ -517,21 +517,19 @@
...
                            WHERE coll_name = ?
-                          UNION
-                          SELECT * from r_coll_main
-                           WHERE coll_name LIKE ? || '/%' ),
+                              OR coll_name LIKE ? || '/%' ),
```
tedgin: I used a `UNION` instead of an `OR`, because the `UNION` appeared to be faster in timing tests.

ianmcorvidae: Hm, interesting. This could be another collation/indexing for `LIKE` queries issue. When it can use an index on `coll_name` for both conditions those two should be combinable, but I guess the `LIKE` currently can't use any indices.

tedgin: That makes sense! If we fix the indexing, we should use `OR`.  It's purpose is clearer.

PR #98, opened 2016-02-22
CORE-6341 Notifications Refactor

ash3rz:
The ticket specifically talks about theming, but I went ahead and did a full refactor for theming, dependency injection, and firing events.

I tried to separate out the commits into logical chunks, but some of the appearance pattern and dependency injection stuff were a bit too intertwined so there may be overlap in those commits.


psarando:
Looks good :+1: 

I'm just wondering about the minor `NotificationViewImpl.ui.xml` rename.
I see sometimes we use a `*ViewImpl.ui.xml` name, but in more places we're using `*View.ui.xml` names.
On the one hand, the `NotificationViewImpl.ui.xml` name just matches the class name where it's used.
On the other hand, the xml file isn't really an implementation of some other, abstract xml file, so the `*View.ui.xml` name makes more sense in that case.

I guess it could also be named depending on if the `ui.xml` is supposed to be specific to the `Impl` class, or is it just for the `View` interface and can be used by any `Impl` of that interface.

ash3rz:
I didn't think it through that far :stuck_out_tongue_closed_eyes:  I really only figured if I rename the file, the `Impl` doesn't have to specify the file name.

I've been thinking about it and going back and forth on whether the `ui.xml` is specific to the `Impl` or just for the `View`.  I think ultimately a change in the `ui.xml` (adding or deleting a UI element for example) would require a change in the `Impl` and vice versa, but not the `View`.  So I think it's specific to the `Impl`.  But then again, I could see how adding `Impl` to the name would sound like it's an implementation of some other xml file... so really, I could go either way.

psarando:
That sounds good to me, so I don't see a problem leaving it as-is: this `NotificationViewImpl.ui.xml` is specific to the `NotificationViewImpl` class.

If someone else has better arguments for the `View.ui.xml` name we can always change it later, so I think this is ready to merge. :shipit:

ash3rz:

PR #99, opened 2016-02-23
CORE-7522: require at least one path in the Paths schema.

ianmcorvidae:
This makes it so that `/secured/filesystem/anon-files` will properly validate -- QA has a test which passes a misspelled key, and terrain was pulling out the nonexistent "paths" key and passing it to data-info, which was then happily doing nothing and returning rather than erroring.

### DIFF
--- a/services/data-info/src/data_info/routes/domain/common.clj
+++ b/services/data-info/src/data_info/routes/domain/common.clj
@@ -11,18 +11,16 @@
 (def DataIdPathParam (describe UUID "The data item's UUID"))
 
 (s/defschema Paths
-  {:paths (describe [NonBlankString] "A list of iRODS paths")})
+  {:paths (describe [(s/one NonBlankString "path") NonBlankString] "A list of iRODS paths")})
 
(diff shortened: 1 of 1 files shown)


services/data-info/src/data_info/routes/domain/common.clj
```diff
@@ -11,18 +11,16 @@
 (def DataIdPathParam (describe UUID "The data item's UUID"))
 
 (s/defschema Paths
-  {:paths (describe [NonBlankString] "A list of iRODS paths")})
+  {:paths (describe [(s/one NonBlankString "path") NonBlankString] "A list of iRODS paths")})
 
 (s/defschema OptionalPaths
-  (-> Paths
-    (->optional-param :paths)))
+  {(s/optional-key :paths) (describe [NonBlankString] "A list of iRODS paths")})
```
ianmcorvidae: This way OptionalPaths will still allow passing an empty list, as well as a missing key.

tedgin:
Looks fine to me. :shipit:

PR #100, opened 2016-02-24
CORE-7445 display error when file upload fails when uploading a file …

sri2k1us:
…> 2G. Note: No specific error msg is displayed cos UI does not receive one.


ash3rz:
Looks straightforward to me!

PR #101, opened 2016-02-24
Rewrite of the JEX

johnworth:
This pull request encompasses all of the current work towards replacing the JEX with a set of services that communicate over an AMQP exchange and no longer depend on HTCondor's logs for event notifications.

This feature is not complete. I'm opening this pull request up so that people can looking over the code a bit more easily and start asking questions (which I'll hopefully be able to answer :sweat:).

Yes, this pull request is way too large. The work contained it in started before we began using pull requests for merges, plus the scope of the project expanded as time went on. The end result is a monster of pull request.

# Overview of changes

## 1. No more JEX

The code for the JEX has been moved into the services/old-jex directory and is no longer actively developed. It remains in the git repo as a reference.

## 2. No more jex-events

The jex-events service is no longer needed. It's functionality has been incorporated into this new project. 

## 3. No more condor-log-monitor

The condor-log-monitor service no longer exists. We aren't parsing the HTCondor event_log to extract event information any more, we're surfacing events from the jobs themselves over a RabbitMQ exchange. We may end up revisiting the condor-log-monitor logic to handle jobs that go into a HELD state, but that can be looked into later.

## 4. No more JEX database

The SQL files and conversions associated with the JEX database are still present so that they can be used as a reference, but the JEX database itself is no longer needed. It's been replaced with the job_status_updates table in the DE database, which is added as part of this pull request. This table contains all of the job status updates that are received by the job-status-recorder service over AMQP.

## 5. jobservices is the new JEX

The functionality that the JEX provided is now provided by the jobservices project. It is a collection of services oriented towards executing and tracking jobs. It is written in Go and built with gb. Check out the README in **services/jobservices** for more information.

#### 5.1.1 condor-launcher

**condor-launcher** connects to the **jobs.launches** topic on the **jobs** exchange and launches jobs on the HTCondor cluster based on any messages that it receives.

When a message is received, condor-launcher creates a directory with this pattern: **/tmp/$username/$job-uuid/**. Inside this directory it creates multiple files: **irods-config**, **config**, **job**, **iplant.cmd**.

**irods-config** contains the settings that porklock needs to transfer files into and out of iRODS.

**config** contains the road-runner and RabbitMQ settings road-runner needs to run jobs and post job status updates.

**job** contains the job JSON generated by the **apps** service.

**iplant.cmd** contains the submission file that HTCondor requires for jobs to be submitted.

The above files and the **road-runner** executable are transferred to the HTCondor node that will run the job. The directory that is created is also the directory that contains the HTCondor logs generated by the jobs.

#### 5.1.2 jex-adapter

**jex-adapter** allows the **apps** service to send messages out on the **jobs.launches** topic without having to be modified. It does this by implementing the REST API from the old JEX service. **apps** is configured to hit this service instead of the old JEX.

#### 5.1.3 job-status-recorder

**job-status-recorder** listens for job status updates sent out by **road-runner** instances on the **jobs.updates** topic and records them in the **job_status_updates** table of the DE database.

#### 5.1.4 job-status-to-apps-adapter

**job-status-to-apps-adapter** periodically queries the **job_status_updates** table for any jobs that have updates that have not been posted to the **apps** service. It then iterates through all of those jobs, attempting to post the unposted updates and recording the result of the attempt back to the DE database.

#### 5.1.5 road-runner

**road-runner** executes jobs based on a JSON blob serialized to a file. Each step of the job runs inside a Docker container. Job results are transferred back into iRODS with the porklock tool. Job status updates are posted to the **jobs.updates** topic in the **jobs** exchange.

#### 5.2.1 jobservices/src/configurate 

Provides common configuration functionality. It supports reading configuration settings from a YAML file. All of the job services are intended to read from the same configuration file.

#### 5.2.2 jobservices/src/logcabin

Provides common logging functionality that can be used to emit messages in the JSON format that we use for logstash/kibana.

#### 5.2.3 jobservices/src/messaging

Provides the logic and data structures that the services will need to communicate with each other over AMQP (as implemented by RabbitMQ).

#### 5.2.4 jobservices/src/model

Contains the logic and data structures for working with job definitions. This allows the services to work with a common data model, hopefully avoiding situations where services have diverging ideas about how to interact with job definitions.

#### 5.2.5 jobservices/src/test

Contains files and stubs that each of the services can use in their respective unit and integration tests.

#### 5.2.6 jobservices/vendor 

Contains the third-party code that the job services and packages use.

## 6. Changes to the facepalm packaging

Facepalm's build has been modified to build the database tarballs and copy them into the facepalm Docker image directly, rather than pull down the pre-built tarballs from Jenkins. This allows facepalm to reference changes in the database that are included in the branch it was built from, rather than only containing the tarballs from the dev branch. In other words, the discoenv/facepalm:jex-go Docker image should contain database tarballs from the jex-go git branch instead of tarballs built from the dev branch sources. Dealing with facepalm on the jex-go before these changes was...frustrating.

## 7. Changes to apps service

The apps service has been modified to remove the job synchronization logic that fired off periodically.

## 8. Changes to the notification_agent service

The notification-agent service has been modified to allow multiple 'Running' messages to flow through it.

## 9. docker-compose for fun and profit

The docker-compose file was modified to remove the container names and the old jex service. It now includes the new job services.

# Job Workflow

A user submits a job through the UI. The job submission is assembled into JSON in the **apps** service, which posts the JSON to the **jex-adapter** service. The **jex-adapter** service sends the JSON as a message out on the **jobs.launches** topic of the **jobs** exchange.

**condor-launcher** receives the message, assembles the job submission for HTCondor, and submits the job by calling out to **condor_submit**. The submitted job is configured to run **road-runner**, passing it configuration files and the JSON job definition inside of a file. More information on these files is provided below.

HTCondor schedules the job for a slot in the cluster. It fires up **road-runner**, which reads in the job JSON and executes the job. It sends out multiple job status updates to the **jobs.updates** topic of the **jobs** exchange.

Each job status update is received by **job-status-recorder**, which saves the update to the **job_status_updates** table of the DE database.

Periodically, **job-status-to-apps-adapter** spins up and queries the database for any jobs that have updates that have not been propagated up to the UI yet. It posts those updates to the **apps** service, recording the status of each attempt.

At that point, **apps** forwards the update to the **notification-agent**, which makes it available to the UI and forwards completion notifications to the **iplant-email** service.


databases/de-database-schema/src/main/tables/76_job_status_updates.sql
```diff
@@ -0,0 +1,48 @@
+SET search_path = public, pg_catalog;
+
+--
+-- Records job status updates
+--
+CREATE TABLE job_status_updates (
+    id uuid NOT NULL DEFAULT uuid_generate_v4(),
+
+    -- corresponds to the external_id field in the job_steps table
+    external_id character varying(64) NOT NULL,
```
ianmcorvidae: If this corresponds, can it be made into a foreign key to that table, or is there a reason it's not?

databases/de-database-schema/src/main/tables/76_job_status_updates.sql
```diff
@@ -0,0 +1,48 @@
...
+    -- The ip address that the update was sent from. This should be set to the
+    -- condor node ip address or to the agave callback source ip.
+    sent_from inet NOT NULL,
+
+    -- The hostname provided in the update message.
+    sent_from_hostname text NOT NULL,
+
+    -- Send date is the date that the update was actually sent, not the date
+    -- that the record was created. Should be in milliseconds since the epoch.
+    sent_on bigint NOT NULL,
```
ianmcorvidae: Though this is sent as milliseconds since epoch, it seems like as a timestamp thing it should probably be stored as such in the DB... I'm not sure I feel strongly about it, but mostly it seems like it'd be annoying to do comparisons with actual timestamps if it's stored this way. What I can find suggests you'd have to do `'1970-01-01 00:00:00 GMT'::timestamp + 
((sent_on::bigint)::text)::interval` to convert to a timestamp, which is... rather ugly to say the least.

databases/de-database-schema/src/main/tables/76_job_status_updates.sql
ianmcorvidae: Unrelated but it looks like this is something which we're joining against in queries, too, which means it'd probably be a good idea to add an index on it.

libs/kameleon/src/kameleon/entities.clj
```diff
@@ -298,3 +298,6 @@
 (defentity user-saved-searches
   (table :user_saved_searches)
   (belongs-to users {:fk :user_id}))
+
+(defentity job-status-updates
+  (table :job_status_updates))
```
ianmcorvidae: I haven't really used korma much but it seems like the link via the `external_id` column can also be added here in some sort of productive fashion.

ianmcorvidae:
As a fairly general comment, I think I'd prefer if we completely removed the `old-jex` directory (and if it's still there, clm stuff); it's always possible to find these old pieces of code in the history if they need to be recovered, so we may as well have the head of our branches represent stuff we actually use, as I see it.

databases/de-database-schema/src/main/tables/76_job_status_updates.sql
johnworth: This originally was a timestamp, but something in the Go stack was reducing the precision down to the second, which isn't good enough to sort entries on. I lost patience and just made it a bigint to maintain my sanity.

databases/de-database-schema/src/main/tables/76_job_status_updates.sql
johnworth: We might benefit from an index here, yes. I was waiting to see if we're actually going to need one before adding one however. Not sure how often we're actually going to be doing queries into this table.

The external_id in job_steps isn't unique or part of a primary key that I could find, so I didn't make this a foreign key.

psarando: Is `external_id` always populated in the `job_steps` table? I thought it only corresponded to Agave app IDs?

psarando: Nevermind, @dennisroberts71 just explained the purpose of this field offline:
It can either be a DE job step ID, or an Agave job step ID.

ianmcorvidae: Hm. It does look like the code that joins on this in apps is unused (vestigial bits of the periodic-sync code). However, job-status-to-apps-adapter does do a select `WHERE external_id = $1` which without an index will need to look at every row of the job_status_updates table; this is in code which happens every time it tries to propagate anything up (as you might expect). To start this will not likely be an issue since the table will start out fairly small, but I think it would be wise to add an index on this column now anyway. "Number of job updates since the new jex was shipped" shouldn't be a linear performance problem.

services/jobservices/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+    DOCKER_USER=johnworth ./docker-build.sh
+
+## Running Tests
+
+**gb** is used to run the unit tests.
+
+    gb test
+
+To run integration tests, use the **test.sh** script. It uses **docker-compose** and the **test.yml** file to spin up containers running RabbitMQ, PostgreSQL, and Docker-in-Docker.
```
strootman: Needs instructions for troubleshooting missing docker network `jobservices`

services/jobservices/src/condor-launcher/condor.go
```diff
@@ -0,0 +1,425 @@
...
+	builtby string
+)
+
+func init() {
+	flag.Parse()
+}
+
+// GenerateCondorSubmit returns a string (or error) containing the contents
+// of what should go into an HTCondor submission file.
+func GenerateCondorSubmit(submission *model.Job) (string, error) {
```
strootman: Is it necessary to export this function? Same goes for `GenerateJobConfig`, `GenerateIRODSConfig`, `CreateSubmissionDirectory`, and `CreateSubmissionFiles`.

services/jobservices/src/condor-launcher/condor.go
```diff
@@ -0,0 +1,425 @@
...
+	output, err := cmd.CombinedOutput()
+	log.Printf("condor_rm output for job %s:\n%s\n", s.CondorID, string(output))
+	if err != nil {
+		return "", err
+	}
+	return string(output), err
+}
+
+// AppVersion prints version information to stdout
+func AppVersion() {
```
strootman: This functionality is duplicated a few times. Very minor, but it could be extracted. It seems that we may be having more `golang`-based projects in the future, so having a common lib which allows us to print app version statements in a consistent manner may be a good idea.

Not necessary for this PR. Would you like me to create an issue?

services/jobservices/src/condor-launcher/condor.go
```diff
@@ -0,0 +1,425 @@
...
+	id, err := submit(cmd, s)
+	if err != nil {
+		log.Printf("Error submitting job:\n%s", err)
+		return "", err
+	}
+	log.Printf("Condor job id is %s\n", id)
+	return id, err
+}
+
+func stop(s *model.Job) (string, error) {
```
strootman: This function is only referenced from a test. 
Is it going to be used?

johnworth: Yes, it's going to be used. I still have to implement stopping jobs.

services/jobservices/src/condor-launcher/condor.go
ianmcorvidae: I've also copied this functionality for my new service, so it may make sense to split it out, yeah. The caveat on this, though, is that the way these things tie together with the go vendoring we either need to have multiple copies of stuff (maybe symlinks will work? I'm not sure...) or we need to put all our go projects in one folder, meaning all their dependencies get upgraded together. So we may need to consider our structure as we grow into more go projects if we want to avoid code duplication, or accept that some duplication will happen and not worry about it too much.

PR #102, opened 2016-02-24
CORE-7519 fix error message details display.

sri2k1us:


ui/de-lib/src/main/java/org/iplantc/de/commons/client/ErrorHandler.java
```diff
@@ -162,11 +162,11 @@ private static String parseExceptionJson(Throwable caught) {
         }
 
         if (jsonError != null) {
-            String name = JsonUtil.getInstance().getString(jsonError, "name"); //$NON-NLS-1$
-            String message = JsonUtil.getInstance().getString(jsonError, "message"); //$NON-NLS-1$
+            String error_code = JsonUtil.getInstance().getString(jsonError, "error_code"); //$NON-NLS-1$
+            String message = JsonUtil.getInstance().getString(jsonError, "reason"); //$NON-NLS-1$
```
psarando: You could go the extra mile here and decode an AutoBean instead of JsonObjects, or that could be done later in another PR.

psarando:
Looks good :+1:

ui/de-lib/src/main/java/org/iplantc/de/commons/client/ErrorHandler.java
sri2k1us: Looked into it. That change includes a lot more refactoring. So its going to a separate PR.

PR #103, opened 2016-02-25
CORE-3358

dennisroberts71:
This change adds duplicate app name checking to the apps services. Two apps that are in the same category and are not marked as deleted should not be allowed to have the same name in order to avoid confusion. The `Beta` category is treated as an exception because it's intended to be a staging area for newly published apps. It is left to the app administrator to determine if an app in the `Beta` category needs to be renamed before it can be placed in its destination category.

### DIFF
--- a/services/apps/src/apps/persistence/app_metadata.clj
+++ b/services/apps/src/apps/persistence/app_metadata.clj
@@ -712,12 +712,25 @@
                        {:aca.app_id (uuidify app-id)
                         :l.id       [not= (user-favorite-subselect :w.root_category_id faves-idx)]})))))
 
-(defn list-duplicate-apps
-  "List apps with the same name that exist in the same category as the new app."
-  [app-name category-ids]
+(defn- list-duplicate-apps*
+  [app-name category-id-set]
   (select [:apps :a]
           (fields :a.id :a.name :a.description)
           (join [:app_category_app :aca] {:a.id :aca.app_id})
-          (where {:aca.app_category_id           [in category-ids]
-                  (raw "trim(both from a.name)") (string/trim app-name)
-                  :a.deleted                     false})))
+          (where {(raw "trim(both from a.name)") (string/trim app-name)
+                  :a.deleted                     false
+                  :aca.app_category_id           [in category-id-set]})))
+
+(defn- app-category-id-subselect
+  [app-id beta-app-category-id]
+  (subselect :app_category_app
+             (fields :app_category_id)
+             (where {:app_id          app-id
+                     :app_category_id [not= beta-app-category-id]})))
+
+(defn list-duplicate-apps
+  "List apps with the same name that exist in the same category as the new app."
+  ([app-name category-ids]
+     (list-duplicate-apps* app-name category-ids))
+  ([app-name app-id beta-app-category-id]
+     (list-duplicate-apps* app-name (app-category-id-subselect app-id beta-app-category-id))))
--- a/services/apps/src/apps/persistence/app_metadata.clj
+++ b/services/apps/src/apps/persistence/app_metadata.clj
@@ -1,6 +1,7 @@
 (ns apps.persistence.app-metadata
   "Persistence layer for app metadata."
   (:use [kameleon.entities]
+        [kameleon.util :only [normalize-string]]
         [kameleon.uuids :only [uuidify]]
         [korma.core :exclude [update]]
         [korma.db :only [transaction]]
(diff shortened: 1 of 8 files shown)


services/apps/src/apps/service/apps/de/admin.clj
```diff
@@ -91,8 +91,10 @@
 (defn update-app
   "This service updates high-level details and labels in an App, and can mark or unmark the app as
    deleted or disabled in the database."
-  [{app-id :id :as app}]
+  [{app-name :name app-id :id :as app}]
   (validate-app-existence app-id)
+  (when-not (nil? app-name)
+    (av/validate-app-name app-name app-id (workspace-beta-app-category-id)))
```
ianmcorvidae: Should this be inside the transaction? I'm not sure how the transaction macro deals with non-SQL errors, so I don't know how the error-throwing would interact there, but if it can be that could help the potential (very rare) race condition of one being inserted between this and the transaction updating the app.

services/apps/src/apps/persistence/app_metadata.clj
```diff
@@ -710,3 +719,28 @@
                            {:u.username username})
                        {:aca.app_id (uuidify app-id)
                         :l.id       [not= (user-favorite-subselect :w.root_category_id faves-idx)]})))))
+
+(defn- list-duplicate-apps*
+  [app-name app-id category-id-set]
+  (select [:apps :a]
+          (fields :a.id :a.name :a.description)
+          (join [:app_category_app :aca] {:a.id :aca.app_id})
+          (where {(raw "trim(both from a.name)") (string/trim app-name)
```
ianmcorvidae: The default behavior of `trim` is to trim both sides with whitespace, so this can just be `(sqlfn trim :a.name)`. For extra equivalentness you could have the DB do the trimming of both sides of the equality but I'm not sure that's worth it.

ianmcorvidae: ... an extra bit of possible improvement here could also be to check the equivalence more ways -- at least `lower(trim(name))` could be nice, and (though I don't know if this is possible in our DB) collapsing whitespace to one space, so something like `lower(regexp_replace(trim(name), '\s+', ' '))` or so.

Sadly we can't add an index to enforce this since it's multi-table and triggers are a pain.

ianmcorvidae:
My comments are all quite minor, so :shipit: when ready.

services/apps/src/apps/service/apps/de/admin.clj
dennisroberts71: Yeah, that's a good point. Any exception that occurs inside the transaction macro will cause a rollback, so including the check in the transaction would help. Good catch.

services/apps/src/apps/persistence/app_metadata.clj
dennisroberts71: Ah, thanks! I noticed that the default behavior was to trim whitespace, but I didn't notice that it was to trim whitespace from both ends of the string.

dennisroberts71: I was on the fence about the other improvements, so I asked Sriram about it. We agreed that they sounded helpful, so I'll add those too. Thanks!

psarando: That's a good point about collapsing the extra whitespace down to one space, especially since multiple spaces will only be displayed as 1 space to the user in the browser.

dennisroberts71: Ah, yes, I hadn't considered that.

services/apps/src/apps/service/apps/de/edit.clj
```diff
@@ -347,8 +347,9 @@
 (defn update-app
   "This service will update a single-step App, including the information at its top level and the
    tool used by its single task, as long as the App has not been submitted for public use."
-  [user {app-id :id :keys [references groups] :as app}]
+  [user {app-id :id app-name :name :keys [references groups] :as app}]
   (verify-app-editable user (persistence/get-app app-id))
+  (validate-app-name app-name app-id (workspace-beta-app-category-id))
```
psarando: Should this validation also be inside the transaction, for the same reasons @ianmcorvidae gave?

ianmcorvidae: Probably, and also the one for relabeling below (I believe)

dennisroberts71: :confounded: I should have checked the other updates.

psarando:
Looks good! :shipit: :+1:

PR #105, opened 2016-02-27
CORE-7180

dennisroberts71:
This change is a slight optimization to use the app search feature in Agave when obtaining app status information for analysis listings.


psarando:
Looks good to me :shipit:

PR #106, opened 2016-02-29
CORE-3599: updated the notification agent documentation in a couple o…

dennisroberts71:
…f places

This is just a minor documentation change. The example was already updated, but the description still needed to be updated. There was one other case where the documentation needed to be updated. I didn't look over the system messages endpoints.


ianmcorvidae:
Straightforward. :shipit:

dennisroberts71:
Cool, thanks for the review.

PR #107, opened 2016-03-01
CORE-4531: added the debug flag to the response body the GET /analyse…

dennisroberts71:
…s/:id/relaunch-info endpoint in apps


ianmcorvidae:
After reading the ticket associated with this, looks pretty straightforward as well. :ship:

PR #110, opened 2016-03-03
Add lb/qos calls to dewey/info-typer; standardize logging some.

ianmcorvidae:
http://reference.clojurerabbitmq.info/langohr.basic.html#var-qos

basic.qos configures how many unacknowledged messages may be sent on the channel unacknowledged; by default, everything's pretty much allocated immediately, is my understanding from documentation; I'd really prefer if it was a little more divided-out. For some reason, info-typer is not getting everything at once; I'm not sure if there's some difference in the implementations such that dewey is getting them all and info-typer isn't. But in any case, this makes it explicitly configured.

I've also added a line to dewey to make it log messages in basically the same format info-typer does, i.e. `[call function] [routing-key] [message]` at the INFO level. dewey basically only logs errors without this, which makes it hard to determine how much or what it's actually doing at any given time.


tedgin:
Please make the QOS a configuration value.  Other than that, the changes look good.

PR #111, opened 2016-03-03
CORE-7542: modified sharkbait to add analysis permissions to Grouper

dennisroberts71:


ianmcorvidae:
It's going to be nice and slow, as we've already found out, but that's no fault of this code, which looks straightforward to me. :shipit:

dennisroberts71:
Cool, thanks for the review.

PR #112, opened 2016-03-04
CORE-7546: register newly submitted analyses in Grouper and give the subm…

dennisroberts71:
…itter ownership permission


ianmcorvidae:
Straightforward as usual, seems to work the same way as apps :ship:

psarando:
Ah, I was confused, the commit and this PR description say `register newly submitted apps`, but isn't this change for analyses?

ianmcorvidae:
Oh, that's a good point. It might be nice to fix up the commit message to say analyses as well (given that's where the PR description came from)

psarando:
Other than fixing the commit message, the changes look good :+1:

dennisroberts71:
Heh, hanks for catching the broken commit message. That's fixed now. Thanks for the reviews.

PR #114, opened 2016-03-04
Make dewey's queue declaration not auto-delete. 

ianmcorvidae:
Durability means very little without this, since it would only be durable in the case of being created and the broker restarted before anything connected to it. This is because `durable` makes it stay on server restarts, but `auto-delete` makes it get deleted any time the number of consumers goes to 0 from a larger number. So, durable + auto-delete means it'd only stay around if for some reason it was created but not connected to before a server restart, which isn't exactly useful.


tedgin:
Looks good! :shipit: .

PR #115, opened 2016-03-08
CORE-7552: validate UUIDs refer to extant objects in stat-gatherer endpoint.

ianmcorvidae:
Just a small bugfix. Previously it was throwing an NPE after the `(partial uuid/get-path cm)` a line below returned nil, in the case of the UUID referring to a nonexistent object.


psarando:

PR #116, opened 2016-03-08
Analysis sharing

dennisroberts71:


services/apps/src/apps/clients/iplant_groups.clj
```diff
@@ -18,6 +18,30 @@
 (def grouper-analysis-permission-def-fmt "iplant:de:%s:analyses:analysis-permission-def")
 (def grouper-analysis-resource-name-fmt "iplant:de:%s:analyses:%s")
 
+(def ^:private permission-precedence
+  (into {} (map-indexed (fn [i v] (vector v i)) ["own" "write" "read"])))
```
psarando: I like this solution, especially if you needed to build this map dynamically with values read from a db, but since they're static in the code, would it be more readable if the final map was just static?
```clojure
{"own"   0
 "write" 1
 "read"  2}
```

dennisroberts71: Yeah, I guess it might be a little more readable that way. I liked the compact representation when I wrote this line of code, but making it explicit sounds fine to me too.

dennisroberts71: I'll rebase dev at the same time.

ianmcorvidae:
I guess this PR just adds the `POST /analyses/permission-lister` endpoint? I don't see anything to do actual sharing of analyses, but I could be missing something.

dennisroberts71:
Correct. This pull request doesn't actually add an option to share the repositories.

ianmcorvidae:
Cool. Seems sensible as far as that, so I'd say :ship:

dennisroberts71:
:+1:

PR #117, opened 2016-03-08
Make the name of the "data" index configurable (in ansible, Infosquito, and terrain)

ianmcorvidae:
In order to update the mappings for template metadata, I'm planning to migrate us to having the 'data' index be an index alias, pointing to an underlying index with some other name. The private ansible repo's `index-template-metadata` has changes to the role there which initializes our elasticsearch cluster with the new mappings, as well as using the newly-configurable index name. I've also made it so Infosquito and terrain use this configurable name, primarily because it was easy. dewey I've left alone, its config is more annoying to pass around.

### DIFF
--- a/ansible/roles/util-cfg-service/templates/terrain.properties.j2
+++ b/ansible/roles/util-cfg-service/templates/terrain.properties.j2
@@ -58,8 +58,9 @@ terrain.jex.base-url = {{ jex.base }}/
 # Tree viewer settings
 terrain.tree-viewer.base-url = {{ tree_parser_base }}
 
-# Elastic Search settings
-terrain.infosquito.es-url = {{ elasticsearch.base }}
+# Elasticsearch settings
+terrain.infosquito.es-url   = {{ elasticsearch.base }}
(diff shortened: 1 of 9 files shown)


ansible/roles/util-cfg-service/templates/terrain.properties.j2
```diff
@@ -58,8 +58,9 @@ terrain.jex.base-url = {{ jex.base }}/
 # Tree viewer settings
 terrain.tree-viewer.base-url = {{ tree_parser_base }}
 
-# Elastic Search settings
-terrain.infosquito.es-url = {{ elasticsearch.base }}
+# Elasticsearch settings
+terrain.infosquito.es-url   = {{ elasticsearch.base }}
+terrain.infosquito.es-index = {{ elasticsearch.data_index }}
```
ianmcorvidae: It's a bit weird that these are prefixed "infosquito" but I'm not really concerned about it so I kept it for consistency.

ianmcorvidae:
(I changed my mind and will only commit the first commit of this, which just adds the property for the private ansible to use, directly to dev)

PR #118, opened 2016-03-09
CORE-7499: Add has_child queries for file and folder (template) metadata

ianmcorvidae:
This implements the UI changes necessary to take advantage of the new template metadata indexing (#91). Instead of a single `nested` query which queries the `metadata` field of the main `file` or `folder` document, this changes it so that it combines this with two other queries in a `bool` via `should`: two `has_child` queries which query the `file_metadata` and `folder_metadata` types, respectively, with their own `nested` queries against the identically-structured `metadata` field of those types. I've set these to use `score_mode` of `max`, which shouldn't really matter much because our structure only has one child document, meaning various averages, maximums, sums, etc. pretty much work the same, but I thought I may as well be explicit.

This requires the reindexing to have already been done; these queries will error when run against an index without the new mappings.

### DIFF
--- a/ui/de-lib/src/main/java/org/iplantc/de/client/services/impl/DataSearchQueryBuilder.java
+++ b/ui/de-lib/src/main/java/org/iplantc/de/client/services/impl/DataSearchQueryBuilder.java
@@ -43,6 +43,7 @@ public class DataSearchQueryBuilder {
     public static final String PATH = "path";
     public static final String METADATA2 = "metadata";
     public static final String NESTED2 = "nested";
+    public static final String HAS_CHILD = "has_child";
     public static final String FILE_SIZE = "fileSize";
     public static final String LABEL = "label";
     public static final String DATE_CREATED = "dateCreated";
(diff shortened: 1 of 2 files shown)


ianmcorvidae:
note: if you'd like to test this, there's some data in the de-2 ES cluster. The query that this generates looks like (other than I've added `inner_hits: {}` to the child queries, since for debugging it's nicer since it'll show the matching child document):

```json
{
  "query": {
    "bool": {
      "should": [
        {
          "nested": {
            "path": "metadata",
            "query": {
              "query_string": {
                "query": "datacite.*",
                "fields": ["metadata.attribute"]
              }
            }
          }
        },
        {
          "has_child": {
            "type": "folder_metadata",
            "score_mode": "max",
            "inner_hits": {},
            "query": {
              "nested": {
                "path": "metadata",
                "query": {
                  "query_string": {
                    "query": "datacite.*",
                    "fields": ["metadata.attribute"]
                  }
                }
              }
            }
          }
        },
        {
          "has_child": {
            "type": "file_metadata",
            "inner_hits": {},
            "score_mode": "max",
            "query": {
              "nested": {
                "path": "metadata",
                "query": {
                  "query_string": {
                    "query": "datacite.*",
                    "fields": ["metadata.attribute"]
                  }
                }
              }
            }
          }
        }
      ]
    }
  }
}```

This can be put into the head plugin interface on the "Any Request" tab, with a POST to `data/file,folder/_search`. With the data currently there, you should see one request which matches in `file_metadata`, one in `folder_metadata`, and some in the normal `file` and `folder` indexes, which should validate this searches in all relevant places.

ash3rz:
I don't see any obvious issues here.  :shipit:

PR #119, opened 2016-03-09
Static IDs - CORE-7179 Epic

ash3rz:
This PR is an attempt to close ~half of the "Facilitate Robust UI Selenium Testing" epic.  These changes will basically add an "id=xyz" attribute to the UI's HTML so QA can more easily run their tests.

This looks like a huge PR, but a lot of the changes are small and repetitive across different commits.  I'd recommend looking at this commit by commit instead of file by file to make sense of the changes involved.

I should point out in order to add an ID to the actual input field of `TextField` UI elements, for example, instead of the overarching combination of the label plus the field, I had to use `setId()` instead of `ensureDebugId()`.  So that explains why some elements are using different methods to set the IDs.


strootman:
I verified that all the debug ids are in place. I glanced at the code, and nothing sticks out as unexpected. 
 I re-opened some of the tickets and left comments. Once those are addressed, I think this PR will be ready.

Aside from the comments I made in some of the jira tickets I re-opened, did you want me to look at anything else in particular?

ash3rz:
Thanks for the review!  No, nothing in particular I think I need you to look at as long as the Ids are where they should be.  I went ahead and pushed changes for CORE-5906 and CORE-6644 to address your comments in the tickets.

PR #121, opened 2016-03-10
upgrade dewey/infosquito/monkey to langohr 3.5.1

ianmcorvidae:
I did the upgrade for info-typer separately since it wasn't changing major versions, but these ones are. This upgrade primarily serves to upgrade dependencies of langohr itself, including lots of bugfixes in the java rabbitmq client. I'm hoping that upgrading this will help dewey the most (as the thing handling the most messages) but it's mostly just intended as keeping on top of things.


ash3rz:
Looks like straight-forward changes to me, but perhaps another set of eyes could be good since I'm still new to Clojure.

psarando:
Looks straight-forward :shipit:

PR #122, opened 2016-03-10
Modernize dewey's config to use defprop etc.

ianmcorvidae:
This makes it so the config is actually validated, and it'll make it easier to add things and use things from places outside of `dewey.core` if we want to (which I think we might for configuring an ES index, for example).

### DIFF
--- a/services/dewey/src/dewey/core.clj
+++ b/services/dewey/src/dewey/core.clj
@@ -9,6 +9,7 @@
             [dewey.amq :as amq]
             [dewey.curation :as curation]
             [dewey.status :as status]
+            [dewey.config :as cfg]
             [common-cli.core :as ccli]
             [me.raynes.fs :as fs]
             [service-logging.thread-context :as tc])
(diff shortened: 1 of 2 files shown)


services/dewey/src/dewey/core.clj
```diff
@@ -32,72 +33,55 @@
...
+                                            (cfg/amqp-qos)
                                             (partial curation/consume-msg irods-cfg es)
                                             "data-object.#"
                                             "collection.#")
                     (log/info "Attached to the AMQP broker.")
                     true
                     (catch Exception e
-                      (log/debug (str e))
-                      (log/info "Failed to attach to the AMQP broker. Retrying...")
+                      (log/info e "Failed to attach to the AMQP broker. Retrying...")
```
ianmcorvidae: This is the only non-config-related change; logging the error this way means the stacktrace will get logged.

services/dewey/src/dewey/core.clj
```diff
@@ -122,7 +106,10 @@
     (try+
      (let [{:keys [options arguments errors summary]} (ccli/handle-args svc-info args cli-options)]
        (when-not (fs/exists? (:config options))
-         (ccli/exit 1 (str "The config file does not exist.")))
-       (run (partial config/load-config-from-file (:config options))))
+         (ccli/exit 1 "The config file does not exist."))
+       (when-not (fs/readable? (:config options))
+         (ccli/exit 1 "The config file is not readable."))
```
ianmcorvidae: I added this check on readability since it seemed like that would also be useful, not that it'd probably work if it went further anyway.

ash3rz:
Looks good to me, but again somebody else should probably take a look too.

psarando:

PR #123, opened 2016-03-11
CORE-7558: added endpoints to grant and revoke access to a data item

dennisroberts71:


services/data-info/src/data_info/routes/permissions.clj
```diff
@@ -28,11 +28,34 @@
...
+
+      (PUT* "/:share-with/:permission" [:as {uri :uri}]
+        :path-params [share-with :- (describe NonBlankString "The user to grant permissions to.")
+                      permission :- (describe PermissionEnum "The permission level to grant.")]
+        :query [params StandardUserQueryParams]
+        :return DataItemPermissionsResponse
+        :summary "Grant Data Item Permissions"
+        :description (str
+"Grants access to a data item."
+(get-error-code-block "ERR_DOES_NOT_EXIST, ERR_NOT_READABLE, ERR_NOT_A_USER", "ERR_NOT_OWNER"))
```
ianmcorvidae: I suspect the last comma of this list was intended to be inside the string? (same below)

services/data-info/src/data_info/services/sharing.clj
```diff
@@ -89,6 +93,52 @@
...
+    (remove-permissions cm unshare-with fpath)
+
+    (when (is-dir? cm fpath)
+      (log/warn "Unsharing directory" fpath "from" unshare-with "by" user)
+      (unshare-dir cm user unshare-with fpath))
+
+    (log/warn "Removing read perms on parents of" fpath "from" unshare-with "by" user)
+    (process-parent-dirs
+      (partial set-readable cm unshare-with false)
+      #(and (not (base-dirs %)) (not (contains-accessible-obj? cm unshare-with %)))
```
ianmcorvidae: I'm not sure I'm seeing where `contains-accessible-obj?` is coming into this namespace from, could be  missing something though. Unless it's from `clj-jargon.permissions`?

services/data-info/src/data_info/services/sharing.clj
```diff
@@ -89,6 +93,52 @@
...
+       1. Remove the access permissions for the user.  This is done recursively in case the path
+          being unshared is a directory.
+
+       2. If the item being unshared is a directory, perform any directory-specific unsharing
+          steps that are required.
+
+       3. Remove the user's read permissions for parent directories in which the user no longer has
+          access to any other files or subdirectories."
+  [cm user unshare-with fpath]
+  (let [base-dirs #{(ft/rm-last-slash (paths/user-home-dir user)) (trash-base-dir (:zone cm) user)}]
```
ianmcorvidae: It might be worth adding a comment that says why this one is the requesting user's home directory rather than the containing home directory like with sharing (which I guess is just that it's fine to revoke read permissions on someone else's home dir).

services/data-info/src/data_info/services/permissions.clj
```diff
@@ -16,16 +17,52 @@
...
 
 (with-post-hook! #'list-permissions (dul/log-func "list-permissions"))
+
+(defn add-permission
+  [{:keys [user]} data-id share-with permission]
+  (with-jargon (cfg/jargon-cfg) [cm]
+    (let [path (ft/rm-last-slash (:path (uuids/path-for-uuid user data-id)))]
+      (validators/user-exists cm user)
+      (validators/user-owns-path cm user path)
+      (validators/user-exists cm share-with)
```
ianmcorvidae: I think I don't see where an `ERR_NOT_READABLE` can happen here (which makes some sense, since there's not much reason to check ownership and readability both). Can probably be removed from the swagger documentation (same for remove-permission).

ianmcorvidae:
Nitpicking aside this looks good to me! :ship:

services/data-info/src/data_info/routes/permissions.clj
dennisroberts71: Yes, thanks for catching that.

services/data-info/src/data_info/services/sharing.clj
dennisroberts71: It does come from `clj-jargon.permissions`. I could put `:only` in the use clause, but I was a little afraid of masking something else that was imported from that namespace.

services/data-info/src/data_info/services/permissions.clj
dennisroberts71: I updated this too, thanks.

services/data-info/src/data_info/services/sharing.clj
dennisroberts71: This is actually trying to determine the base case for recursion into the parent directories. This assumes that the file being unshared is in the requesting user's home directory, however, which is not necessarily the case. I'll modify this to build the base directory from the file path instead. Good catch.

ianmcorvidae: Ah, okay. There's already a function used by the share code which calculates the nearest containing home directory, which you should be able to reuse I think.

dennisroberts71: Cool; I'll take a look at it.

dennisroberts71:
The code review comments have all been addressed. Merging the PR now.

PR #124, opened 2016-03-11
CORE-7550 Update terrain's Perm-ID Request endpoints to return null folders instead of errors when folder does not exist.

psarando:
Also updated Belphegor to display "Not Found" messages for null folders in the listing grid and when the "View Metadata" action is selected.
The Perm-ID Request "View Metadata" error message also includes the request's original path.

### DIFF
--- a/ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/permIdRequest/views/PermanentIdRequestViewImpl.java
+++ b/ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/permIdRequest/views/PermanentIdRequestViewImpl.java
@@ -154,7 +154,7 @@ public class PermanentIdRequestViewImpl extends Composite implements PermanentId
         ColumnConfig<PermanentIdRequest, String> nameCol = new ColumnConfig<>(pr_props.requestedBy(),
                                                                               appearance.nameColumnWidth(),
                                                                               appearance.nameColumnLabel());
-        ColumnConfig<PermanentIdRequest, String> pathCol = new ColumnConfig<>(pr_props.path(),
+        ColumnConfig<PermanentIdRequest, String> pathCol = new ColumnConfig<>(new FolderPathProvider(appearance),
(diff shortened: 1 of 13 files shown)


ianmcorvidae:
Looks reasonable as far as I can tell; some extra review on the UI parts at least might be good though, since I don't know that part very well.

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/permIdRequest/views/PermanentIdRequestViewImpl.java
```diff
@@ -154,7 +153,7 @@ public void onDialogHide(DialogHideEvent event) {
         ColumnConfig<PermanentIdRequest, String> nameCol = new ColumnConfig<>(pr_props.requestedBy(),
                                                                               appearance.nameColumnWidth(),
                                                                               appearance.nameColumnLabel());
-        ColumnConfig<PermanentIdRequest, String> pathCol = new ColumnConfig<>(pr_props.path(),
+        ColumnConfig<PermanentIdRequest, String> pathCol = new ColumnConfig<>(new PermanentIdRequestPathProvider(appearance),
```
ash3rz: I think you can still inject`PermanentIdRequestPathProvider` in `PermanentIdRequestViewImpl`'s constructor like you have with the `appearance` and 'factory', but it's not critical.

ash3rz:
The UI aspects look fine to me! :ship:

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/permIdRequest/views/PermanentIdRequestViewImpl.java
psarando: Good point.
I don't think this is required for the tests, and since we don't need the ability to swap this `ValueProvider` for others via Injection, I think I'll leave it for now.

PR #125, opened 2016-03-12
Add the POST /analyses/sharing endpoint

dennisroberts71:
...and the POST /analyses/unsharing endpoint.

### DIFF
--- a/services/apps/src/apps/routes/analyses.clj
+++ b/services/apps/src/apps/routes/analyses.clj
@@ -45,11 +45,26 @@
          :return perms/AnalysisPermissionListing
          :summary "List App Permissions"
          :description "This endpoint allows the caller to list the permissions for one or more analyses.
-         The authenticated user must have ownership permission on every analysis in the request body for
-         this endpoint to succeed."
+         The authenticated user must have read permission on every analysis in the request body for this
+         endpoint to succeed."
          (ok (apps/list-job-permissions current-user (:analyses body))))
 
-  (PATCH* "/:analysis-id" []
+(POST* "/sharing" []
+         :query [params SecuredQueryParams]
+         :body [body (describe perms/AnalysisSharingRequest "The app sharing request.")]
+         :return perms/AnalysisSharingResponse
+         :summary "Add App Permissions"
+         :description "This endpoint allows the caller to share multiple analyses with multiple users. The
+         authenticated user must have ownership permission to every analysis in the request body for this
+         endpoint to fully succeed. Note: this is a potentially slow operation and the response is returned
+         synchronously. The DE UI handles this by allowing the user to continue working while the request is
+         being processed. When calling this endpoint, please be sure that the response timeout is long
+         enough. Using a response timeout that is too short will result in an exception on the client side.
+         On the server side, the result of the sharing operation when a connection is lost is undefined. It
+         may be worthwhile to repeat failed or timed out calls to this endpoint."
+         (ok (apps/share-jobs current-user (:sharing body))))
+
+(PATCH* "/:analysis-id" []
           :path-params [analysis-id :- AnalysisIdPathParam]
           :query       [params SecuredQueryParams]
           :body        [body AnalysisUpdate]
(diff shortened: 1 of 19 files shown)


services/apps/src/apps/routes/analyses.clj
```diff
@@ -45,11 +45,26 @@
...
+         :summary "Add App Permissions"
+         :description "This endpoint allows the caller to share multiple analyses with multiple users. The
+         authenticated user must have ownership permission to every analysis in the request body for this
+         endpoint to fully succeed. Note: this is a potentially slow operation and the response is returned
+         synchronously. The DE UI handles this by allowing the user to continue working while the request is
+         being processed. When calling this endpoint, please be sure that the response timeout is long
+         enough. Using a response timeout that is too short will result in an exception on the client side.
+         On the server side, the result of the sharing operation when a connection is lost is undefined. It
+         may be worthwhile to repeat failed or timed out calls to this endpoint."
+         (ok (apps/share-jobs current-user (:sharing body))))
```
dennisroberts71: I found the instances of `app` in this documentation and fixed it. I'll submit the updates to this in the next pull request.

services/apps/src/apps/service/apps.clj
```diff
@@ -360,3 +360,11 @@
...
   [user job-ids]
   (jobs/list-job-permissions (get-apps-client user) user job-ids))
+
+(defn share-jobs
+  [user sharing-requests]
+  {:sharing (jobs/share-jobs (get-apps-client user) user sharing-requests)})
+
+(defn can-access-app
+  [user username app-id required-level]
+  (.hasAppPermission (get-apps-client user) username app-id required-level))
```
dennisroberts71: This function will be removed in the next pull request.

dennisroberts71:
The POST /analyses/unsharing endpoint is ready, so I just merged the changes related to that endpoint into this branch. The changes are small enough not to require a separate pull request.

services/apps/src/apps/service/apps/jobs/sharing.clj
```diff
@@ -0,0 +1,170 @@
...
+      (share-output-folder sharer sharee job)
+      (iplant-groups/share-analysis job-id sharee level)))
+
+(defn- share-accessible-job
+  [apps-client sharer sharee job-id job level]
+  (if-let [failure-reason (do-job-sharing-steps apps-client sharer sharee job-id job level)]
+    (job-sharing-failure job-id job level failure-reason)
+    (job-sharing-success job-id job level)))
+
+(defn- share-extant-job
```
ianmcorvidae: (on a random function)

this seems like a really verbose way to do this, with a bunch of functions with their own parameter lists etc. It doesn't change the functionality any so it shouldn't block things, but is there a reason you chose this over putting them into one function for `do-job-sharing-steps`, `share-accessible-job` and `share-extant-job` plus one for the catch (perhaps just `share-job` and `share-job*` as you've done with other stuff)?

dennisroberts71: Yes, I chose to put them into separate functions to keep the nesting levels within functions shallow. Admittedly, the additional functions do add some verbosity. On the other hand, they also tend to keep the functions from becoming syntax salad.

ianmcorvidae: Ah, right, I see they're all a lot of ifs and ors and such which would do that. To some degree the bunches of defn and parameter lists are looking pretty leafy and green too but I do see why this is better.

ianmcorvidae:
I think this looks ready to go so downstream testing can get started, so :shipit:. I have at least one more annoying nitpicking question but I know people are waiting and it looks functionally ready!

services/apps/src/apps/service/apps/jobs/sharing.clj
dennisroberts71: Having `share-job*` and `share-job*` would work, but I wanted a finer level of granularity here along with function names that provided a little bit of documentation.

PR #126, opened 2016-03-14
Analysis sharing

dennisroberts71:
This PR consists of a few different changes. The primary change was to add a Boolean flag to the analysis listing service indicating whether or not the analysis could be shared. An analysis can be shared if it does not have a parent analysis (that is, if it's not a member of an HT batch) and if all of its execution systems support analysis sharing. The rest of the changes were mostly related to validation. The only exception was the notification type used for analysis sharing notifications, which was changed from `analyses` to `analysis`.


ianmcorvidae:
LGTM :shipit:

PR #128, opened 2016-03-15
Analysis sharing

dennisroberts71:
This change adds a way to filter analyses by submitter. There are three options: list analyses submitted by the authenticated user, list analyses _not_ listed by the authenticated user, and list all analyses. The default is to list all analyses. There were also some problems with the way that HT batches were being shared; these problems have been fixed.


dennisroberts71:
I'm merging this PR now, but if I could get a post-merge review, I'd appreciate it.

psarando:
I didn't find any issues with this PR :+1:

dennisroberts71:
Cool, thanks for the review.

PR #130, opened 2016-03-16
CORE-7566: also unshare input files when unsharing an analysis

dennisroberts71:
This PR includes unsharing input files when an analysis is unshared and the problems with the parameter values and relaunch info endpoints.

### DIFF
--- a/services/apps/src/apps/service/apps/jobs/sharing.clj
+++ b/services/apps/src/apps/service/apps/jobs/sharing.clj
@@ -173,16 +173,28 @@
    (catch ce/clj-http-error? {:keys [body]}
      (str "unable to unshare result folder: " (:error_code (service/parse-json body))))))
 
-;; The apps client isn't used at this time, but it will be once we extend analysis sharing
-;; to HPC apps.
+(defn- unshare-input-file
+  [sharer sharee path]
+  (try+
+   (data-info/unshare-path sharer path sharee)
+   nil
+   (catch ce/clj-http-error? {:keys [body]}
+     (str "unable to unshare input file: " (:error_code (service/parse-json body))))))
+
+(defn- unshare-child-job
+  [apps-client sharer sharee job]
+  (or (process-job-inputs (partial unshare-input-file sharer sharee) apps-client job)
+      (iplant-groups/unshare-analysis (:id job) sharee)))
+
 (defn- unshare-job*
   [apps-client sharer sharee job-id job]
   (or (verify-not-subjob job)
       (verify-accessible sharer job-id)
       (verify-support apps-client job-id)
       (unshare-output-folder sharer sharee job)
+      (process-job-inputs (partial unshare-input-file sharer sharee) apps-client job)
       (iplant-groups/unshare-analysis job-id sharee)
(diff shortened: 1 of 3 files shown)


services/apps/src/apps/service/apps/jobs/sharing.clj
```diff
@@ -173,16 +173,28 @@
    (catch ce/clj-http-error? {:keys [body]}
      (str "unable to unshare result folder: " (:error_code (service/parse-json body))))))
 
-;; The apps client isn't used at this time, but it will be once we extend analysis sharing
-;; to HPC apps.
+(defn- unshare-input-file
+  [sharer sharee path]
+  (try+
+   (data-info/unshare-path sharer path sharee)
```
ianmcorvidae: I'm not completely sure how this would play out so I'm commenting to ensure you've thought about it: what happens if an input file is a file which the analysis sharee had access to before the analysis was shared? I think this is especially important in the case where the sharee might even be the input file's owner -- say, a professor has some data which they share with a student, who runs an analysis on that data and then shares the analysis back to the professor, or, in the case where the input file is a public file.

I guess that this call will fail if the sharer isn't the owner of the input file, which will cover most of the cases but might result in the unsharing failing, when it probably should continue in that case.

Thoughts?

ianmcorvidae:
The rest of the PR looks good to me, other than the one functional concern I commented on.

services/apps/src/apps/service/apps/jobs/sharing.clj
dennisroberts71: Yeah, that's a good question. It would cause the unsharing operation to fail in most cases, which is not really correct. Unfortunately, there's no way to determine whether or not the sharee had access to the file before the analysis was originally shared. Possibly a better option would be only to attempt to unshare the file if the sharer is the owner of the file. During a sharing operation, this is handled by the file sharing service, which doesn't attempt to share the path if the sharee already has access to it.

I think I'm going to leave it as it is for the time being and create a Jira issue to modify the behavior of the unsharing endpoint in the next release.

ianmcorvidae: Sounds good. :shipit:

As a sort of aside, I don't know if it would make sense to consider the permission the file currently has as a marker of whether unsharing should happen -- for example, if the sharee has write permission but analysis sharing would only have given them read, it could be argued they shouldn't have their permissions removed. But this could get confusing for other reasons and I'm not sure what variant is more confusing.

We could also do something like have analysis sharing add an administrative AVU when the user already had permissions before the sharing, so that the unsharing endpoint can restore those permissions exactly. But that's definitely a task for later if we want to do it!

dennisroberts71: The administrative AVU idea does sound interesting, although we'd have to take care to handle the case where multiple analyses that use the same input file are shared with the same user. We'd have to keep a counter in that AVU indicating how many analyses are currently shared with the user. And increment or decrement that count each time, only restoring the original permissions when the counter reaches zero. And we also wouldn't want to reduce permissions on the file if the ownership already has access, which would seem to indicate that the current permission should be used as an indicator. There's a lot of edge cases here.

PR #131, opened 2016-03-17
Sharing

sri2k1us:


ui/de-lib/src/main/java/org/iplantc/de/analysis/client/presenter/AnalysesPresenterImpl.java
```diff
@@ -353,8 +351,13 @@ public void onShareSelected(List<Analysis> selected) {
 
     @Override
     public void setCurrentFilter(AnalysisFilter filter) {
-       this.currentFilter = filter;
-       loadAnalyses(currentFilter);
+        if(filter == null) {
+            currentFilter =filter;
+            return;
```
ianmcorvidae: Not knowing the code that well so bear with me, but why do we not call `loadAnalyses` in the case of setting the filter to null (which I assume is intended to un-set the filter)?

ianmcorvidae:
Assuming my one comment isn't an issue and is just me misunderstanding the code, the rest of the changes look straightforward!

ui/de-lib/src/main/java/org/iplantc/de/analysis/client/presenter/AnalysesPresenterImpl.java
sri2k1us: LoadAnalyses method has been refactored to construct the query that should be sent to service while currentFilter is just a filed that track what UI is displaying in the drop down (later used for session restore). When the drop down value changes, i am updating the currentFilter value  and calling loadAnalyses to update the view. Not sure if this answers your questions but thats my thinking here...

dennisroberts71: Does this mean that the analysis listing won't be refreshed if the user removes the current filter?

sri2k1us: User cannot clear the filter. He can only change it one value to other. The code programmatically clears the filter when user is drilling down a HT analysis or doing a search.

dennisroberts71: Ah, okay, so the listing is refreshed elsewhere in those cases.

PR #132, opened 2016-03-18
CORE-7356 Fix data window updates on folder moves.

psarando:
Reverted terrain's [POST /filesystem/move endpoint](http://cyverse.github.io/DE/api/endpoints/filesystem/move.html#moving-files-andor-directories) response.
Updated the DiskResourceServiceFacadeImpl logic that updates its folder cache when folders are moved, since we can no longer lookup folders in the cache by path.
Instead, the source and destination folders are refreshed when folders have been moved to and from them.
Refreshing instead of preserving folder caches allows some DiskResourcesMovedEvent handler logic to be cleaned up in the NavigationPresenterImpl.

### DIFF
--- a/services/terrain/src/terrain/clients/data_info.clj
+++ b/services/terrain/src/terrain/clients/data_info.clj
@@ -170,12 +170,10 @@
     (raw/move-single user path-uuid dest)))
 
 (defn move
-  "Uses the data-info single and bulk mover endpoints to move an item or many items into a new directory."
+  "Uses the data-info bulk mover endpoint to move items into a new directory."
   [{:keys [user]} {:keys [sources dest]}]
   (validators/not-superuser user)
-  (if (= 1 (count sources))
-    (move-single user (first sources) dest)
-    (raw/move-multi user sources dest)))
+  (raw/move-multi user sources dest))
 
 (defn move-contents
   "Uses the data-info set-children-directory-name endpoint to move the contents of one directory
--- a/ui/de-lib/src/main/java/org/iplantc/de/client/services/impl/DiskResourceServiceFacadeImpl.java
+++ b/ui/de-lib/src/main/java/org/iplantc/de/client/services/impl/DiskResourceServiceFacadeImpl.java
@@ -374,9 +374,10 @@ public class DiskResourceServiceFacadeImpl extends TreeStore<Folder> implements
     }
 
     @Override
-    public void moveDiskResources(final List<DiskResource> diskResources,
+    public void moveDiskResources(final Folder sourceFolder,
                                   final Folder destFolder,
-                                  AsyncCallback<DiskResourceMove> callback) {
+                                  final List<DiskResource> diskResources,
+                                  final AsyncCallback<DiskResourceMove> callback) {
 
         String address = deProperties.getDataMgmtBaseUrl() + "move"; //$NON-NLS-1$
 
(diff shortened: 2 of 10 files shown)


dennisroberts71:
This looks good to me, although someone who's more familiar with the UI code should probably take a look at it.

ash3rz:
This looks good to me :+1:

services/terrain/src/terrain/clients/data_info.clj
```diff
@@ -170,12 +170,10 @@
...
 
 (defn move
-  "Uses the data-info single and bulk mover endpoints to move an item or many items into a new directory."
+  "Uses the data-info bulk mover endpoint to move items into a new directory."
   [{:keys [user]} {:keys [sources dest]}]
   (validators/not-superuser user)
-  (if (= 1 (count sources))
-    (move-single user (first sources) dest)
-    (raw/move-multi user sources dest)))
+  (raw/move-multi user sources dest))
```
ianmcorvidae: I don't see anything wrong with doing this, but I'm curious why it was needed. I guess just to make the response consistent?

ianmcorvidae: (this should not block merging, just curious. I don't have any complaints about the code)

psarando: Yes, the different response for 1 file/folder caused the UI to throw an exception when attempting to parse the new response, since it was always expecting the `sources` array in the response from the [POST /filesystem/move endpoint](https://github.com/cyverse/DE/blob/gh-pages-beta/api/endpoints/filesystem/move.md#moving-files-andor-directories).

ui/de-lib/src/main/java/org/iplantc/de/client/services/impl/DiskResourceServiceFacadeImpl.java
```diff
@@ -391,44 +397,66 @@ public void moveDiskResources(final List<DiskResource> diskResources,
...
+                                final DiskResourcesMovedEvent movedEvent) {
+        if (!diskResourceUtil.isDescendantOfFolder(destFolder, sourceFolder)) {
+            // The source folder is not under the destination, so it needs to be refreshed.
+            refreshFolderOnChildrenMoved(sourceFolder, resourcesMoved, callback, movedEvent);
         }
+        if (!diskResourceUtil.isDescendantOfFolder(sourceFolder, destFolder)) {
+            // The destination is not under the source folder, so it needs to be refreshed.
+            refreshFolderOnChildrenMoved(destFolder, resourcesMoved, callback, movedEvent);
+        }
+    }
```
psarando: One issue I haven't solved, introduced by this new `onFoldersMoved` method, is now the UI will display 2 "success" notifications when one or more folders are moved from a source directory to a destination that isn't nested under the source, and vice versa. Since this method will call the callback's `onSuccess` method once for each `refreshFolderOnChildrenMoved` call.

For example, if the folder `/a/b/c/d/` is moved into the folder `/a/b/`, then only `/a/b/` will be refreshed and the callback's `onSuccess` method will only be called once after that refresh.
If the folder `/a/b/c/d/` is moved into the folder `/a/e/f/`, then both `/a/b/c/` and `/a/e/f/` will be refreshed and the callback's `onSuccess` will be called after each refresh.

PR #133, opened 2016-03-18
CORE-7579: fixed a problem with the way administrative group membersh…

dennisroberts71:
…ip was being validated in belphegor

### DIFF
--- a/ui/de-webapp/src/main/java/org/iplantc/de/conf/AdminWebSecurityConfig.java
+++ b/ui/de-webapp/src/main/java/org/iplantc/de/conf/AdminWebSecurityConfig.java
@@ -145,7 +145,7 @@ public class AdminWebSecurityConfig extends WebSecurityConfigurerAdapter {
             .authorizeRequests()
             .antMatchers("/**/logout", "/**/logged-out", "/*.css", "/*.png").permitAll()
             .anyRequest().authenticated()
-            .anyRequest().hasAnyRole(authorizedGroups)
+            .anyRequest().hasAnyAuthority(authorizedGroups.split("\\s*,\\s*"))
             .and().exceptionHandling().authenticationEntryPoint(adminCasAuthenticationEntryPoint()).and()
             .exceptionHandling().accessDeniedPage("/access-denied");
         /*


psarando:
I can't tell if this change will fix the bug since I don't know this API, and I can't tell the difference between `hasAnyRole` and `hasAnyAuthority` (and if passing an array vs. a comma-delimited string will work), but the `split` regex looks good to me :+1:

dennisroberts71:
Heh, I guess I should have included some more details. The problem that we were encountering was that `hasAnyRole` prepends `ROLE_` to each of its arguments. In our case, the argument was a comma-delimited string containing a bunch of group names. Apparently `hasAnyRole` does allow the argument to be a comma-delimited string, but it still appends `ROLE_` to the entire string. So, for example, if `authorizedGroups` was `foo,bar,baz` then it was treated as if it were `ROLE_foo,bar,baz`. Apparently, Spring Security still supports the comma-delimited format, however, so any user in the groups `ROLE_foo`, `bar` or `baz` could log in. But since the group `ROLE_foo` doesn't exist, users who are in group `foo` and not in any of the other groups can't log in.

I checked the source code for a possible alternative:

```
        /**
         * Shortcut for specifying URLs require any of a number of roles. If you
         * do not want to have "ROLE_" automatically inserted see
         * {@link #hasAnyAuthority(String...)}
         *
         * @param roles
         *            the roles to require (i.e. USER, ADMIN, etc). Note, it
         *            should not start with "ROLE_" as this is automatically
         *            inserted.
         * @return the {@link ExpressionUrlAuthorizationConfigurer} for further
         *         customization
         */
        public ExpressionInterceptUrlRegistry hasAnyRole(String... roles) {
            return access(ExpressionUrlAuthorizationConfigurer.hasAnyRole(roles));
        }
```

This led me to try `hasAnyAuthority`. Since the argument type is `String...`, I thought it would be better for us to split the string ourselves rather than rely on Spring to do it for us.

I tested the change with `authorizedGroups` set to `dev`, `tito-admins,dev` and `dev,tito-admins`.

PR #134, opened 2016-03-21
Unit Tests (for automatic test running)

ianmcorvidae:
This should be fairly straightforward. I've already updated the relevant Jenkins jobs where changes needed to be made. I've also updated the terrain Jenkins job, which was already running tests, to also parse the junit XML output.

I've added tests as `clojure.test` ones where I was adding new tests; this is the kind that can get parsed by Jenkins rather than midje, but I don't really have terribly strong feelings about it as long as we can read the output, so it shouldn't be considered an imposition of policy, just a vague preference on my part.


services/iplant-email/test/iplant_email/core_test.clj
```diff
@@ -0,0 +1,7 @@
+(ns iplant-email.core-test
```
ianmcorvidae: this file move is because this seems to be the more common version now; the old test didn't have any functional components anyway

dennisroberts71:
Looks good to me. :shipit:

PR #135, opened 2016-03-23
CORE-7573 Fix Texta Font 404 errors

ash3rz:
Relative paths in the CSS url field were causing 404 errors.  Switched to using full path location in the .java files which the CSS files then use.

### DIFF
--- a/ui/de-lib/src/main/java/org/iplantc/de/theme/base/client/desktop/BaseDesktopAppearance.java
+++ b/ui/de-lib/src/main/java/org/iplantc/de/theme/base/client/desktop/BaseDesktopAppearance.java
@@ -7,6 +7,7 @@ import org.iplantc.de.resources.client.messages.IplantDisplayStrings;
 
 import com.google.gwt.core.client.GWT;
 import com.google.gwt.resources.client.ClientBundle;
+import com.google.gwt.resources.client.DataResource;
 import com.google.gwt.resources.client.ImageResource;
 
 import com.sencha.gxt.widget.core.client.button.IconButton.IconConfig;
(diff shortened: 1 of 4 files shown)


ui/de-lib/src/main/java/org/iplantc/de/theme/base/client/desktop/BaseDesktopAppearance.java
```diff
@@ -63,6 +64,10 @@
 
         @Source("org/iplantc/de/theme/base/client/desktop/user.png")
         ImageResource userPrefImg();
+
+        @DataResource.MimeType("font/opentype")
+        @Source("org/iplantc/de/theme/base/client/desktop/Texta_Font/Texta-Bold.otf")
+        DataResource textaBold();
```
ianmcorvidae: Operating as usual off of no experience, why does this need to be here when both of the font variants are in `IplantBaseWindowAppearance.java` below? I recall one of the other times I asked this sort of thing it was a Belphegor vs. DE difference but it doesn't look like it here.

ianmcorvidae:
Generally looks pretty good. At some point we might move the font format to be WOFF as well, since that seems to be the preferred direction things are moving (better compression, faster load times, somewhat better information embedded in the file), but that's very unrelated to this PR. :shipit:

psarando:

ui/de-lib/src/main/java/org/iplantc/de/theme/base/client/desktop/BaseDesktopAppearance.java
ash3rz: Good question.  Well, when I remove the definition from `BaseDesktopAppearance` the fonts still look correct and I can switch from Bold to Regular with no issue, so it would seem that it's not necessary to have the fonts defined in both places.  But since I don't fully understand the intricacies of how that all works or if there are random cases where it would stop working without the definition, I'd like to leave it here for now.  I will look into how that all works later though.

ash3rz:
Thanks for the review!

PR #136, opened 2016-03-23
CORE-7497: Templeton: incremental mode

ianmcorvidae:
These two commits add the handling for incremental update messages (sent on the `de` exchange with routing key `metadata.update` and a JSON body with 'entity' and 'author' keys) as well as adds the code to the metadata service to publish these messages.

### DIFF
--- a/services/metadata/src/metadata/services/avus.clj
+++ b/services/metadata/src/metadata/services/avus.clj
@@ -23,7 +23,7 @@
 
 (defn update-metadata-template-avus
   "Adds or Updates AVUs associated with a Metadata Template for the given user's data item."
-  [user-id data-id data-type template-id {avus :avus}]
+  [user-id data-id data-type template-id {avus :avus} & {:keys [publish-update] :or {publish-update true}}]
   (templates/validate-template-exists template-id)
   (let [avus (map (partial find-existing-metadata-template-avu data-id) avus)
         existing-avus (filter :id avus)
@@ -35,7 +35,7 @@
       (when (not-empty new-avus)
         (persistence/add-metadata-template-avus user-id new-avus data-type))
       (persistence/set-template-instances data-id template-id (map :id avus)))
-    (amqp/publish-metadata-update user-id data-id)
+    (when publish-update (amqp/publish-metadata-update user-id data-id))
     (persistence/metadata-template-avu-list data-id template-id)))
 
 (defn remove-metadata-template-avu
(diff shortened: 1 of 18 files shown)


services/metadata/src/metadata/services/avus.clj
```diff
@@ -34,6 +35,7 @@
       (when (not-empty new-avus)
         (persistence/add-metadata-template-avus user-id new-avus data-type))
       (persistence/set-template-instances data-id template-id (map :id avus)))
+    (amqp/publish-metadata-update user-id data-id)
```
ianmcorvidae: These publishings have to happen after the transaction has committed. I don't know of any functionality which would let me put this inside the transaction and have it sent when the transaction commits, which is what would be preferable. But, if this ends up sending extra messages, that's not too much of a concern either; the indexing is fast and always updates it to the current state, whatever that state is.

services/metadata/src/metadata/services/avus.clj
```diff
@@ -135,5 +140,7 @@
     (let [avu-ids (set (map :id (:avus (when template-id
                                          (update-metadata-template-avus
                                            user-id data-id data-type template-id metadata)))))]
-      (persistence/remove-orphaned-avus data-id avu-ids))
-    (list-metadata-template-avus data-id)))
+      (persistence/remove-orphaned-avus data-id avu-ids)))
+  ;; this one will duplicate the one sent by update-metadata-template-avus but we want to catch any orphaned AVUs that were removed and make sure it's outside of a transaction
+  (amqp/publish-metadata-update user-id data-id)
```
ianmcorvidae: This will usually be a second message sent -- if we start having issues with too much reindexing we'll want to make it possible to disable sending a message within `update-metadata-template-avus` in this case. I couldn't find anything that let me know if I was already in a transaction, which might be nicest -- if it's inside a DB transaction, always skip sending the message and wait for the call which is outside of the transaction.

Of course, having something that'd let me just run a hook when the outermost transaction commits would still be nicest :)

ianmcorvidae:
I haven't added deployment stuff to this yet. I'll probably push that out either later today or sometime tomorrow; it'll span repositories a bit more and I thought I could get this up for review with just the service changes themselves first.

ianmcorvidae:
There are also deployment changes in the `templeton-incremental` branch of the private de-ansible, for private inventories and docker-compose file updates.

tedgin:
I think I understood the code changes well enough.  I didn't see any problems with it.  It looks good to me.  :shipit:

ianmcorvidae:
(I'll merge this just before a 2.6->dev merge, to synchronize with the merge that needs to happen in the private de-ansible repo)

PR #137, opened 2016-03-24
Dewey's irods port as passed to irods/init needs to be a string

ianmcorvidae:
Otherwise it throws a bunch of errors about how it can't cast a string to an integer when it tries to call `Integer/parseInt` on the value it's passed.


dennisroberts71:
Looks good to me.

PR #138, opened 2016-03-25
Core 7583

dennisroberts71:
These configuration files are somewhat specific to our deployment, but they do provide a decent reference configuration, which is important for something as extensively configurable as Grouper.


ianmcorvidae:
Looks straightforward :shipit:

PR #139, opened 2016-03-25
Don't create temporary passwords for a fake user to create tickets

ianmcorvidae:
(at least, not when we already have an IRODSAccount to be used)

Assuming this doesn't break everything, it fixes the one place we make temporary passwords. I think this could be done better, probably? There seems to be a lot of vestigial code in this area of the codebase.

I'll work on testing that this actually works tomorrow morning, anyway.


dennisroberts71:
The code changes look fine to me. Assuming the testing goes well, :shipit:.

PR #140, opened 2016-03-25
CORE-7521: added the new Dockerfile for the public Grouper image

dennisroberts71:
This PR adds the Dockerfile used to create a Grouper image to the public code repository. There wasn't any sensitive information in the original Grouper image, so there's no reason not to make it public. I also made some minor changes to this version of the Dockerfile in an attempt to make the image smaller.


docker/grouper/Dockerfile
```diff
@@ -0,0 +1,132 @@
+FROM jeanblanchard/java:jdk-7
+MAINTAINER Dennis Roberts <dennis@cyverse.org>
+ARG git_commit=unknown
+LABEL org.iplantc.de.javabase.git-ref="$git_commit"
```
ianmcorvidae: It doesn't make much sense to overload this label for the grouper image as well as the javabase image. If we do want to add an extra label to track the git-ref when this image was built, we should give it a new name.

docker/grouper/Dockerfile
```diff
@@ -0,0 +1,132 @@
+FROM jeanblanchard/java:jdk-7
```
ianmcorvidae: I'm guessing the answer is just "because it doesn't work with Java 8" but in case it isn't: is there a reason we're not just using our existing `discoenv/javabase` image which already has JCE etc. and should also have a JDK?

dennisroberts71: We need to have a JDK to install Grouper and Grouper does complain (although it seems to work fine) if it runs against Java 8.

docker/grouper/Dockerfile
```diff
@@ -0,0 +1,132 @@
...
+        "$GROUPER_TEMP" \
+    \
+    && cd "$GROUPER_BASE" \
+    && tar -xzvf /tmp/tarballs/grouper-api.tar.gz \
+    && mv "grouper.apiBinary-$GROUPER_VERSION" "$GROUPER_HOME" \
+    && mv /tmp/configs/api/grouper.properties "$GROUPER_HOME/conf/" \
+    && mv /tmp/configs/api/grouper.hibernate.properties "$GROUPER_HOME/conf/" \
+    && mv /tmp/configs/api/grouper-loader.properties "$GROUPER_HOME/conf/" \
+    && mv /tmp/configs/api/subject.properties "$GROUPER_HOME/conf" \
+    && mv /tmp/configs/api/log4j.properties "$GROUPER_HOME/conf" \
```
ianmcorvidae: These files are pretty small so it probably doesn't matter, but the way this is done now they'll appear in multiple layers and thus their size will be (I believe, in this case) doubled. It looks like you're doing it this way because it's impractical to create `$GROUPER_HOME` early and just add them there in the first place, though, which seems fine.

docker/grouper/Dockerfile
dennisroberts71: Good catch; I'll change the name.

docker/grouper/Dockerfile
```diff
@@ -0,0 +1,132 @@
...
+# Copy the configuration files to temporary locations.
+RUN mkdir -p /tmp/configs/api /tmp/configs/ui /tmp/configs/ws
+COPY api/grouper.properties /tmp/configs/api
+COPY api/grouper.hibernate.properties /tmp/configs/api
+COPY api/grouper-loader.properties /tmp/configs/api
+COPY api/subject.properties /tmp/configs/api
+COPY api/log4j.properties /tmp/configs/api
+COPY ui/build.properties /tmp/configs/ui
+COPY ws/build.properties /tmp/configs/ws
+COPY ws/grouper-ws.properties /tmp/configs/ws
```
ianmcorvidae: Depending how much capability for reuse of individual files you foresee, it might be more efficient to just put `api`, `ui`, and `ws` into a folder and just `COPY` that one folder in. These layers are small enough that the reuse may not matter much -- usually the cache is more useful for things like installing packages anyway, which take some time and network traffic.

ianmcorvidae:
Other than my comments which might make the docker image mildly nicer but can also happily be ignored in all likelihood, there's not any actual issues I see here, so :shipit: as soon as you're ready.

docker/grouper/Dockerfile
dennisroberts71: Yeah, the configuration files need to be present when the compilation is done, and I wanted to keep the retrieval of the tarballs, the compilation and the cleanup all in one step in order to keep the size of that layer down.

PR #141, opened 2016-03-25
Reorganize code so clj-jargon.tickets/ticket-admin-service can be private.

ianmcorvidae:
Minor thing, but it seemed weird that we had one utility function for tickets in a kifshare namespace, used only once, and that was the only reason `ticket-admin-service`, which seemed intended to be private by its docstring, wasn't private.

So I moved it. ¯\\_(ツ)_/¯


psarando:
Seems fine to me :shipit:

PR #142, opened 2016-03-28
Core 7583

dennisroberts71:
Some inline comments will be included in the diff.


ansible/inventories/group_vars/all
```diff
@@ -282,11 +282,8 @@ gpg_home_dir:
 group_config:
 
 grouper:
-  service_name: iplant-grouper.service
-  service_name_short: iplant-grouper
-  compose_service: iplant_grouper
-  service_description: Grouper UI and Web Services
-  image_name: iplant-grouper
+  compose_service: grouper
+  image_repo: "{{ docker.user }}/grouper:2.2.2"
```
dennisroberts71: Most of the settings that were removed here were relics from the systemd deployment that we were using for Grouper. I changed the `image_name` member to `image_repo` to indicate that this is the full name of the repository we'll be using to retrieve the docker image.

ansible/inventories/group_vars/all
```diff
@@ -302,11 +299,8 @@ grouper:
     user:
     pass:
   http_server:
-    service_name: grouper-nginx.service
-    service_name_short: grouper-nginx
     compose_service: grouper_nginx
-    service_description: Grouper nginx
-    image_name: "{{ de.http_server.image_name }}"
+    image_repo: "{{ de.http_server.image_name }}"
```
dennisroberts71: These changes were made for the same reasons the previous changes.

docker/grouper/Dockerfile
```diff
@@ -15,10 +15,11 @@ ENV ANT_URL $ANT_BASE/apache-ant-$ANT_VERSION-bin.tar.gz
 
 # Environment variables for tomcat.
 ENV CATALINA_HOME /opt/tomcat
+ENV TOMCAT_CONF /etc/tomcat
 ENV PATH $CATALINA_HOME/bin:$PATH
 ENV TOMCAT_MAJOR 7
-ENV TOMCAT_VERSION 7.0.68
-ENV TOMCAT_BASE https://www.apache.org/dist/tomcat
+ENV TOMCAT_VERSION 7.0.65
+ENV TOMCAT_BASE https://archive.apache.org/dist/tomcat
```
dennisroberts71: This was done to work around an apparent conflict between the CSRF detection used by Grouper itself and the new `RestCsrfPreventionFilter` included in Tomcat 7.0.66.

docker/grouper/Dockerfile
```diff
@@ -76,6 +77,14 @@ RUN apk add --update curl \
...
     && cd "$CATALINA_HOME" \
     && rm bin/*.bat \
+    && mkdir -p "$TOMCAT_CONF" \
+    && mv "$CATALINA_HOME/conf/server.xml" "$TOMCAT_CONF/server.xml" \
+    && ln -s "$TOMCAT_CONF/server.xml" "$CATALINA_HOME/conf/server.xml" \
+    && mv "$CATALINA_HOME/conf/tomcat-users.xml" "$TOMCAT_CONF/tomcat-users.xml" \
+    && ln -s "$TOMCAT_CONF/tomcat-users.xml" "$CATALINA_HOME/conf/tomcat-users.xml" \
+    && rm -f "$CATALINA_HOME/bin/setenv.sh" \
+    && touch "$TOMCAT_CONF/setenv.sh" \
+    && ln -s "$TOMCAT_CONF/setenv.sh" "$CATALINA_HOME/bin/setenv.sh" \
```
dennisroberts71: These changes were made to allow us to continue to have our Tomcat configuration files in `/etc/tomcat` in the configs container.

psarando: So if the configs container isn't mounted, then these default files will be used, but if the container is mounted, then the configs volume overrides these defaults?

ianmcorvidae: Yup, this is how data volumes work in general. https://docs.docker.com/engine/userguide/containers/dockervolumes/#creating-and-mounting-a-data-volume-container mentions it a few paragraphs down:

> In this case, if the postgres image contained a directory called /dbdata then mounting the volumes from the dbstore container hides the /dbdata files from the postgres image. The result is only the files from the dbstore container are visible.

dennisroberts71: Correct. I guess I could have included a default Grouper user in `tomcat-users.xml`, but I figured that these files will be overridden most of the time.

psarando:
Looks good to me :shipit:

dennisroberts71:
Cool, thanks for the review.

PR #143, opened 2016-03-30
CORE-7588: time out requests made as part of terrain's /admin/status

ianmcorvidae:
This way, nonresponsive services are marked as non-functional in a timely fashion.

### DIFF
--- a/services/terrain/src/terrain/services/admin.clj
+++ b/services/terrain/src/terrain/services/admin.clj
@@ -40,11 +40,14 @@
   [url-to-scrub]
   (str (url/url url-to-scrub :path "/")))
 
+(defn get-with-timeout
+  [url]
+  (client/get url {:socket-timeout 1000 :conn-timeout 1000}))
 
 (defn perform-jex-check
   []
   (try
-    (let [s (:status (client/get (config/jex-base-url)))]
+    (let [s (:status (get-with-timeout (config/jex-base-url)))]
       (log/info "HTTP Status from JEX: " s)
       (<= 200 s 299))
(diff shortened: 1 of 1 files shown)


services/terrain/src/terrain/services/admin.clj
```diff
@@ -40,11 +40,14 @@
   [url-to-scrub]
   (str (url/url url-to-scrub :path "/")))
 
+(defn get-with-timeout
+  [url]
+  (client/get url {:socket-timeout 1000 :conn-timeout 1000}))
```
ianmcorvidae: I'm open to opinions about whether 1 second is a long enough timeout or not.

psarando: /shrug My gut says 5~10 seconds would be better, in case of network lag and not because the JEX is down?

psarando: Also, if this times out, will it throw an exception? Or will the status value contain a timeout message for this key?

ianmcorvidae: Yeah, may as well add an extra 0. What we're trying to prevent is just it running functionally forever/six-hour returns/that sort of thing, it's not like 10 seconds is really an issue even if every request we do as part of this takes that long.

ianmcorvidae: It throws an exception, at least according to clj-http documentation. We already handle exceptions to return a false value and all that.

psarando:
:shipit: Looks good :+1:

PR #144, opened 2016-04-04
iRODS AMQP broker can now be on separate host

tedgin:
This is a pull request against the master branch to bring the AMQP broker configuration specified in ansible scripts inline with what's currently in prod.  There are corresponding changes in the `irods_amqp_cfg` branch of the `de-ansible` private repo.

### DIFF
--- a/ansible/inventories/group_vars/all
+++ b/ansible/inventories/group_vars/all
@@ -394,6 +394,10 @@ irods:
   admins: rodsadmin
   data_curators_group: data-curators
   bad_chars: \u0060\u0027\u000A\u0009
+  amqp_broker:
+    host: "{{ groups['amqp-brokers'][0] }}"
+    port:
+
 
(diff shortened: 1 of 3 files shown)


ianmcorvidae:
Looks straightforward to me. :shipit:

ianmcorvidae:
actually, one question -- do we want to potentially handle the case of having a different username/password, too? this just makes it work with a different host/port, but requires the users/passwords to match.

tedgin:
I think we do want to handle that case, but I am trying to do the minimal fix against the master branch.  Once the changes are pushed down into the dev branch, I was planning to make the rest of them there.  I am expecting there to be more changes to add, but I haven't given any thought to what those could be.

Does this seem ok, or would it be more clear to have the username and password changes be part of this merge?

ianmcorvidae:
That seems fine, makes sense.

PR #145, opened 2016-04-04
Move all of the Go projects into a single top-level directory

johnworth:
This pull request includes the changes needed to create a single top-level directory containing all of the code written in Go. This simplifies sharing code across Go projects, as well as building and testing all of the code at once.

The top-level directory is called "golang" and is organized to be used with the gb build tool.


johnworth:
I'm an idiot.

PR #146, opened 2016-04-04
Move Go code into a single top-level project.

johnworth:
This pull request includes the changes needed to create a single top-level directory containing all of the code written in Go. This simplifies sharing code across Go projects, as well as building and testing all of the code at once.

The top-level directory is called "golang" and is organized to be used with the gb build tool.


golang/vendor/manifest
```diff
@@ -2,6 +2,18 @@
...
+		},
+		{
+			"importpath": "github.com/mitchellh/go-homedir",
+			"repository": "https://github.com/mitchellh/go-homedir",
+			"revision": "d682a8f0cf139663a984ff12528da460ca963de9",
+			"branch": "master"
+		},
+		{
 			"importpath": "github.com/codegangsta/negroni",
 			"repository": "https://github.com/codegangsta/negroni",
```
ianmcorvidae: (this comment is actually about something way lower in the file)

streadway/amqp I'd fixed down to a recent good commit in templeton, which isn't preserved here. I haven't checked if it's an older version or a newer one, but we might put it back to that in any case.

ianmcorvidae:
I don't see a docker build script or dockerfile for templeton preserved here.

ianmcorvidae:
I'm also not sure how we intend to integrate the various README files, but templeton's seems to have disappeared in this change.

We should also preserve templeton's `test.sh`, which should also be able to be used for jobservices unit tests. (integration stuff is of course elsewhere)

ianmcorvidae:
If you could go over what the changes to Jenkins will be for this that'd probably be good. I think we may for now, pending code reorganization, have to set it up to build things more often than they need to (with an include pattern that includes this whole golang directory, such that templeton is rebuilt even when only jobservices things change and vice versa); in addition to that I guess it'll just be updating the paths and the exact build scripts to use. But I'd be curious if you've thought more about anything I might have missed or more precise include paths/code reorganization/etc.

johnworth:
Templeton's README and test.sh files are now available in the golang directory.

Templeton's Dockerfile has been renamed templeton.docker and is also in the golang directory.

I've separated the binary builds for the code from the Docker image builds. The code is built with build.sh,  while the Docker images are built with the docker-build-*.sh scripts.

johnworth:
Templeton's streadway/amqp dependency has been brought over to the golang manifest with the same ref. The missing elastic.v3 dependecy was brought over as well.

ianmcorvidae:
Looks like everything works with templeton, so :shipit: as far as I can see. I haven't checked jobservices stuff since I assume you have, and I haven't tested the tools stuff except for make-jwt which seemed to work fine, because I'm not completely sure how to.

PR #148, opened 2016-04-05
Cleaned up/generalized iRODS AMQP config changes

tedgin:
This PR expands the configuration changes that allow iRODS and the DE to use separate AMQP brokers.  A few cleanups were made as well.


psarando:

PR #149, opened 2016-04-06
Core 7596

dennisroberts71:
This change modifies the notification agent to publish messages to AMQP and removes the HTTP callback that previously existed (and did nothing).


services/NotificationAgent/src/notification_agent/messages.clj
```diff
@@ -54,35 +55,16 @@
...
    and returns the state object."
-  [msg]
+  [{:keys [user] :as msg}]
   (let [uuid          (persist-msg msg)
         email-request (:email_request msg)]
     (log/debug "UUID of persisted message:" uuid)
     (when-not (nil? email-request)
       (.start (Thread. #(send-email-request uuid email-request))))
-    (send-msg uuid msg)))
+    (amqp/publish-msg user msg)))
```
ianmcorvidae: (unrelated)

it'll be really nice when we can also send the email request as an AMQP message here instead of needing to start extra threads!

dennisroberts71:
I still need to add some integration tests. I'll do that tomorrow.

services/NotificationAgent/src/notification_agent/messages.clj
dennisroberts71: :+1:

ianmcorvidae:
LGTM :boat:

dennisroberts71:
Cool, thanks for the review.

PR #150, opened 2016-04-07
Rewrite user-sessions service in Go

johnworth:
In order to consume fewer resources in our deployment environments, we decided to rewrite user-sessions in Go. It now lives in the golang directory with the rest of the Go code.

The internal docker-compose files have been updated on the user-sessions branch and integration tests have been added to the user-sessions branch of the internal scm repo as well.


services/user-sessions/src/user_sessions/serve.clj
```diff
@@ -1,63 +0,0 @@
-(ns user-sessions.serve
-  (:use [kameleon.user-session-queries]
```
ianmcorvidae: If they're not used elsewhere, we can probably delete some stuff out of this kameleon namespace, which we might as well clean out as we rewrite stuff.

johnworth: I'd rather do that in a separate pull request.

golang/src/queries/users.go
```diff
@@ -0,0 +1,27 @@
+package queries
+
+import "database/sql"
+
+// IsUser returns whether the provided user exists in the database
+func IsUser(db *sql.DB, username string) (bool, error) {
+	var (
+		count int64
+		query = `SELECT COUNT(*) FROM ( SELECT DISTINCT id FROM users WHERE username = $1 ) AS check_user`
```
ianmcorvidae: for the benefit of other reviewers: because this is code/DB stuff I'm unfamiliar with I checked -- this username field is the long username e.g. `mian@iplantcollaborative.org` rather than `mian`; this is the username which terrain (the only client of user-sessions) passes along, so this is correct.

ianmcorvidae: It doesn't strictly need to go into this -- the query you're copying is doing the "take a count and check it's positive" approach too, and this is a simple rewrite -- but I think it'd be clearer to just have this query be `SELECT EXISTS(SELECT id FROM users WHERE username = $1)`, which directly returns a boolean.

golang/src/user-sessions/main.go
```diff
@@ -0,0 +1,393 @@
...
+	"queries"
+	"strings"
+
+	"github.com/gorilla/mux"
+	_ "github.com/lib/pq"
+)
+
+var (
+	version = flag.Bool("version", false, "Print the version information")
+	cfgPath = flag.String("config", "/etc/iplant/de/jobservices.yml", "The path to the config file")
```
ianmcorvidae: This is not a job-related service except in the very vague sense that all our services sort of revolve around jobs at *some* level, so this seems like an odd default. We can always pass this name via our deployment since I believe you're doing it this way because that file will already have the appropriate DB configuration, but for the default I'd prefer for it to be a name that's more obvious, for downstream folks who might want to use this.

golang/src/user-sessions/main.go
```diff
@@ -0,0 +1,393 @@
...
+	}
+}
+
+// hasSessions returns whether or not the given user has a session already.
+func (u *UserSessionsApp) hasSessions(username string) (bool, error) {
+	query := `SELECT COUNT(s.*)
+              FROM user_sessions s,
+                   users u
+             WHERE s.user_id = u.id
+               AND u.username = $1`
```
ianmcorvidae: (another nice-to-have-but-happy-to-ignore)

this could be an EXISTS directly returning a boolean as well: `SELECT EXISTS(SELECT s.id FROM user_sessions s, users u WHERE s.user_id = u.id AND u.username = $1)`

golang/src/user-sessions/main.go
```diff
@@ -0,0 +1,393 @@
...
+		return sessions, err
+	}
+	return sessions, nil
+}
+
+// insertSession adds a new session to the database for the user.
+func (u *UserSessionsApp) insertSession(username, session string) error {
+	query := `INSERT INTO user_sessions (user_id, session)
+                 VALUES ($1, $2)`
+	userID, err := queries.UserID(u.db, username)
```
ianmcorvidae: The database pedant in me is mildly uncomfortable with us doing two queries/database roundtrips here when it could probably be done with one, but combining them would probably require separating the query string from the execution in the `queries` namespace and I'm not sure that's worth it. Just noting it here in case you or someone else has a clever idea that's not too invasive.

golang/src/user-sessions/main.go
```diff
@@ -0,0 +1,393 @@
...
+	if err != nil {
+		errored(writer, fmt.Sprintf("Error reading body: %s", err))
+		return
+	}
+	if err = json.Unmarshal(bodyBuffer, &checked); err != nil {
+		errored(writer, fmt.Sprintf("Error parsing request body: %s", err))
+		return
+	}
+	bodyString := string(bodyBuffer)
+	if !hasSession {
```
ianmcorvidae: With us depending on this value for an upsert-type operation, it would be best if this was in a transaction, I think. Not sure how much of a pain that would be to handle.

golang/src/user-sessions/main_test.go
```diff
@@ -0,0 +1,85 @@
...
+
+import "testing"
+
+func TestConvertBlankSession(t *testing.T) {
+	record := &UserSessionRecord{
+		ID:      "test_id",
+		Session: "",
+		UserID:  "test_user_id",
+	}
+	actual, err := convert(record, false)
```
ianmcorvidae: Unlikely to go wrong but still might be worth adding a quick test or several which call `convert(..., true)`, since it looks like all of these pass `false`.

ianmcorvidae:
Looks in general fine to me. I'd prefer the default config file to be changed before merging since those defaults are primarily documentation and the current one looks weird/wrong. Aside from that, :shipit:

golang/src/user-sessions/main.go
johnworth: I'm hoping to get as much of the configs as I can into a single file. Right now that file is jobservices.yml, but it will probably get renamed to something more generic in the future. For now, I'd prefer to leave it.

services/user-sessions/src/user_sessions/serve.clj
psarando: I disagree. I think putting off cleaning up obsolete code in another commit or PR makes it more likely that cruft will get left behind.
I also think it makes an unnecessary logical disconnect. Looking at a later commit or PR that just removes some obsolete code might not tell you why it became obsolete (was it replaced with something better? Or was it only used in one place and that code was removed for some reason?). If the obsolete code was removed along with the commit or PR that made it obsolete, then it's more obvious why it was removed and requires less explanation.

PR #151, opened 2016-04-07
Rewrite user-preferences in Go

johnworth:
This is similar to the user-sessions rewrite. It even has the same flaws :smiling_imp: 

There are user-preferences branches for the internal scm and ansible repos, as well.

### DIFF
--- a/golang/docker-build-user-preferences.sh
+++ b/golang/docker-build-user-preferences.sh
@@ -21,5 +21,5 @@ docker build -f user-preferences.docker \
              --build-arg buildenv_git_commit=$BUILDENV_GIT_COMMIT \
              --build-arg version=$VERSION \
              --pull --rm -t "$DOCKER_USER/$DOCKER_REPO:dev" .
-#docker push $DOCKER_USER/$DOCKER_REPO:dev
-#docker rmi $DOCKER_USER/$DOCKER_REPO:dev
+docker push $DOCKER_USER/$DOCKER_REPO:dev
+docker rmi $DOCKER_USER/$DOCKER_REPO:dev
--- a/golang/docker-build-user-preferences.sh
+++ b/golang/docker-build-user-preferences.sh
@@ -1,0 +1,25 @@
+#!/bin/sh
+
+set -x
+set -e
+
+if [ -z "$DOCKER_USER" ]; then
+	DOCKER_USER=discoenv
+fi
+
+if [ -z "$DOCKER_REPO" ]; then
+	DOCKER_REPO=user-preferences
+fi
+
+VERSION=$(cat version | sed -e 's/^ *//' -e 's/ *$//')
+GIT_COMMIT=$(git rev-parse HEAD)
+BUILD_USER=$(whoami)
+BUILDENV_GIT_COMMIT=$(docker inspect -f '{{ (index .Config.Labels "org.iplantc.de.buildenv.git-ref")}}' $DOCKER_USER/buildenv:latest)
+
+docker build -f user-preferences.docker \
+             --build-arg git_commit=$GIT_COMMIT \
+             --build-arg buildenv_git_commit=$BUILDENV_GIT_COMMIT \
+             --build-arg version=$VERSION \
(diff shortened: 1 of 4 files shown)


golang/docker-build-user-preferences.sh
```diff
@@ -0,0 +1,25 @@
...
+BUILD_USER=$(whoami)
+BUILDENV_GIT_COMMIT=$(docker inspect -f '{{ (index .Config.Labels "org.iplantc.de.buildenv.git-ref")}}' $DOCKER_USER/buildenv:latest)
+
+docker build -f user-preferences.docker \
+             --build-arg git_commit=$GIT_COMMIT \
+             --build-arg buildenv_git_commit=$BUILDENV_GIT_COMMIT \
+             --build-arg version=$VERSION \
+             --pull --rm -t "$DOCKER_USER/$DOCKER_REPO:dev" .
+#docker push $DOCKER_USER/$DOCKER_REPO:dev
+#docker rmi $DOCKER_USER/$DOCKER_REPO:dev
```
ianmcorvidae: I think we want to push the image, so this should probably be uncommented, or am I missing something?

johnworth: Fixed. Sorry about that.

golang/src/user-preferences/main.go
```diff
@@ -0,0 +1,393 @@
...
+
+// UserPreferencesRecord represents a user preferences stored in the database
+type UserPreferencesRecord struct {
+	ID          string
+	Preferences string
+	UserID      string
+}
+
+// convert makes sure that the JSON has the correct format. "wrap" tells convert
+// whether to wrap the object in a map with "session" as the key.
```
ianmcorvidae: this is incorrect documentation, it's not "session" here. might be nice to move a generalized version of this into a shared namespace, I bet the other two services will also want this and no sense providing more ways to produce copy-paste errors (there's also a reference to 'session' in a few comments below)

golang/src/user-preferences/main.go
```diff
@@ -0,0 +1,393 @@
...
+}
+
+// New returns a new *UserPreferencesApp
+func New(db *sql.DB) *UserPreferencesApp {
+	return &UserPreferencesApp{
+		db: db,
+	}
+}
+
+// hasSessions returns whether or not the given user has preferences already.
```
ianmcorvidae: not sessions here either!

golang/src/user-preferences/main.go
```diff
@@ -0,0 +1,393 @@
...
+             WHERE p.user_id = u.id
+               AND u.username = $1`
+	var count int64
+	if err := u.db.QueryRow(query, username).Scan(&count); err != nil {
+		return false, err
+	}
+	return count > 0, nil
+}
+
+// getSessions returns a []UserPreferencesRecord of all of the preferences associated
```
ianmcorvidae: (etc., I'll stop)

golang/src/user-preferences/main.go
```diff
@@ -0,0 +1,393 @@
...
+	}
+	response, err := convert(&retval, wrap)
+	if err != nil {
+		return nil, fmt.Errorf("Error generating response for username %s: %s", username, err)
+	}
+	var jsoned []byte
+	if len(response) > 0 {
+		jsoned, err = json.Marshal(response)
+		if err != nil {
+			return nil, fmt.Errorf("Error generating session JSON for user %s: %s", username, err)
```
ianmcorvidae: (okay maybe not completely stopped, but this one's in actual code rather than just a comment)

ianmcorvidae:
Other than the variety of `s/session/preference/` things, and checking on the scm changes for the tests, assuming those pass this looks good to me, broadly speaking. I do think some work generalizing some of this into shared namespaces would make sense around now, because the other two services of this sort are going to want a lot of the same stuff (e.g. PUT turning into POST, convert/wrap code, etc.) so it'll probably make those even smaller when you do them.

johnworth:
The references to sessions should be fixed now.

ianmcorvidae:
Seems fine to ship if you're not breaking reusable code out into a shared namespace of some sort at this time.

PR #152, opened 2016-04-07
Add an nginx+consul-template dockerfile/etc. intended for use with service nginx containers.

ianmcorvidae:
This is the first step toward the zero-downtime deployment setup. There's still lots to do in terms of ansible changes, docker-compose file modifications, etc. (not least of all including this image in our docker-compose files) but I think this can be independently reviewed.

Most of the configuration for this comes in through environment variables. As an example:

`docker run -p 5555:5555 -e CONSUL_CONNECT=some-consul-server:8500 -e NGINX_PROXY_PORT=5555 -e NGINX_PROXY_SERVICE_NAME=anon-files -e DE_ENV=de-2 nginx-consul-template`

This invocation would run an nginx on port 5555 (in-container and outside of the container) which proxies for anon-files instances marked as being in de-2. The color (blue/green -- really just an arbitrary tag) which it should use is stored in a consul k/v; in this case, it'd be `/de-2/anon-files/color`.

consul-template operates as a "proper" PID 1 -- cleans up zombies, etc., which is why I have it be that. However, we wanted the nginx logs to stdout too, so I run everything such that those messages won't be hidden.

nginx is smart about how it does things, so if a change in consul produces an invalid nginx config (for example, if the color changes but the new color doesn't have any servers, all of the previous servers will be marked backup, which is an invalid state) the reload won't actually reload the config. This way things stay up, as much as possible. The `|| true` in the consul-template invocation ensures that consul-template itself doesn't exit if the `nginx -s reload` exits non-zero, such that correcting the mistake in consul can then regenerate the config and try again. Basically, other than explicitly killing it or some unforeseen bug in consul-template, it's pretty much not possible for this container to die off entirely.

Since we'll be running this on a server with registrator, we'll probably also end up wanting to add `SERVICE_NAME`, `SERVICE_CHECK_*`, etc. environment variables as well, but that's unrelated to this.

### DIFF
--- a/docker/nginx-consul-template/nginx.conf.tmpl
+++ b/docker/nginx-consul-template/nginx.conf.tmpl
@@ -5,7 +5,7 @@ events {
 http {
     upstream service {
       {{- range service (printf "%s.%s" (env "DE_ENV") (env "NGINX_PROXY_SERVICE_NAME")) }}
-      server {{.Address}}:{{.Port}} max_fails=3 fail_timeout=60 weight=1 {{- if not (.Tags.Contains (key (printf "%s/%s/color" (env "DE_ENV") (env "NGINX_PROXY_SERVICE_NAME"))))}} backup{{end}}; # {{.ID}} {{.Tags}}
+      server {{.Address}}:{{.Port}} max_fails=3 fail_timeout=60 weight=1 {{- if not (.Tags.Contains (key_or_default (printf "%s/%s/color" (env "DE_ENV") (env "NGINX_PROXY_SERVICE_NAME")) "green"))}} backup{{end}}; # {{.ID}} {{.Tags}}
       {{- else}}
       server 127.0.0.1:65535; # Force a 502. No {{env "DE_ENV"}}.{{env "NGINX_PROXY_SERVICE_NAME"}} services were found.
       {{- end}}
--- a/docker/nginx-consul-template/nginx.conf.tmpl
+++ b/docker/nginx-consul-template/nginx.conf.tmpl
@@ -1,0 +1,24 @@
+events {
+    worker_connections 4096;
+}
+
+http {
+    upstream service {
+      {{- range service (printf "%s.%s" (env "DE_ENV") (env "NGINX_PROXY_SERVICE_NAME")) }}
+      server {{.Address}}:{{.Port}} max_fails=3 fail_timeout=60 weight=1 {{- if not (.Tags.Contains (key (printf "%s/%s/color" (env "DE_ENV") (env "NGINX_PROXY_SERVICE_NAME"))))}} backup{{end}}; # {{.ID}} {{.Tags}}
+      {{- else}}
+      server 127.0.0.1:65535; # Force a 502. No {{env "DE_ENV"}}.{{env "NGINX_PROXY_SERVICE_NAME"}} services were found.
+      {{- end}}
+    }
+
+    server {
+      listen {{env "NGINX_PROXY_PORT"}};
+
+      location / {
+        proxy_pass http://service;
+        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+        proxy_set_header Host $host;
+        proxy_set_header X-Real-IP $remote_addr;
+      }
+    }
+}
--- a/docker/nginx-consul-template/nginx.conf.tmpl
+++ b/docker/nginx-consul-template/nginx.conf.tmpl
@@ -3,12 +3,12 @@ events {
 }
 
 http {
+{{- with $service_query := (printf "%s.%s" (env "DE_ENV") (env "NGINX_PROXY_SERVICE_NAME")) }}
+  {{- if service $service_query }}
     upstream service {
-      {{- range service (printf "%s.%s" (env "DE_ENV") (env "NGINX_PROXY_SERVICE_NAME")) }}
+    {{- range service $service_query }}
       server {{.Address}}:{{.Port}} max_fails=3 fail_timeout=60 weight=1 {{- if not (.Tags.Contains (key_or_default (printf "%s/%s/color" (env "DE_ENV") (env "NGINX_PROXY_SERVICE_NAME")) "green"))}} backup{{end}}; # {{.ID}} {{.Tags}}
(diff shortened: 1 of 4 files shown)


ianmcorvidae:
By way of example, the config this generates looks like (scrubbing some IP addresses and stuff):

```
events {
    worker_connections 4096;
}

http {
    upstream service {
      server 9.9.9.9:55555 max_fails=3 fail_timeout=60 weight=1; # server.iplantcollaborative.org:etc_anon_files_1:60000 [de-2 green]
    }
    
    server {
      listen 5555;
    
      location / {
        proxy_pass http://service;
        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
        proxy_set_header Host $host;
        proxy_set_header X-Real-IP $remote_addr;
      }
    }
}
```

ianmcorvidae:
Right, I keep forgetting details. The way this works is we make it so the containers are always tagged (in the consul fashion) with their environment and a color (this is just adding a line to the docker-compose files for registrator to pick up and put into consul). That's how all the templates work, basically, is through those tags.

docker/nginx-consul-template/nginx.conf.tmpl
```diff
@@ -0,0 +1,24 @@
+events {
+    worker_connections 4096;
+}
+
+http {
+    upstream service {
+      {{- range service (printf "%s.%s" (env "DE_ENV") (env "NGINX_PROXY_SERVICE_NAME")) }}
+      server {{.Address}}:{{.Port}} max_fails=3 fail_timeout=60 weight=1 {{- if not (.Tags.Contains (key (printf "%s/%s/color" (env "DE_ENV") (env "NGINX_PROXY_SERVICE_NAME"))))}} backup{{end}}; # {{.ID}} {{.Tags}}
+      {{- else}}
+      server 127.0.0.1:65535; # Force a 502. No {{env "DE_ENV"}}.{{env "NGINX_PROXY_SERVICE_NAME"}} services were found.
```
dennisroberts71: It'd be nice if there were a way to explicitly return a 502 instead of relying on the fact that nobody is listening to port 65535, but I don't see a way to do it in the docs.

dennisroberts71:
Looks reasonable to me. :shipit:

docker/nginx-consul-template/nginx.conf.tmpl
ianmcorvidae: I've changed this to completely replace the contents of the `http` block when no servers can be found. In the case where it can, it puts in the upstream etc. as usual, but where it can't, it just puts a simple server block with `return 502` directly. This returns an HTML default nginx 502 error, which we might want to replace instead with some sort of standard JSON response (which should be possible with an `error_page` declaration, I believe). Thoughts?

docker/nginx-consul-template/nginx.conf.tmpl
```diff
@@ -0,0 +1,32 @@
+events {
+    worker_connections 4096;
+}
+
+http {
+{{- with $service_query := (printf "%s.%s" (env "DE_ENV") (env "NGINX_PROXY_SERVICE_NAME")) }}
```
ianmcorvidae: (I also factored this out into a variable. Because sometimes I can refactor usefully :P

docker/nginx-consul-template/nginx.conf.tmpl
```diff
@@ -0,0 +1,32 @@
...
+  {{- if service $service_query }}
+    upstream service {
+    {{- range service $service_query }}
+      server {{.Address}}:{{.Port}} max_fails=3 fail_timeout=60 weight=1 {{- if not (.Tags.Contains (key_or_default (printf "%s/%s/color" (env "DE_ENV") (env "NGINX_PROXY_SERVICE_NAME")) "green"))}} backup{{end}}; # {{.ID}} {{.Tags}}
```
ianmcorvidae: Two things I'm not sure if I should try to build in or not and I'd love opinions on:

a.) when no color is configured in consul, should I try to have it automatically detect and choose a color that has servers? alternatively, should I make it choose all servers by having it default to `(env "DE_ENV")` (which will always match in this line)

b.) when a color is configured and there are servers, but all for other colors, right now this will generate an invalid upstream in which all the servers are marked 'backup'; then, the config will update but the `nginx -s reload` will fail so it'll keep using the old config. consul-template sadly does not support querying by multiple tags at once in the `service` function it provides, and I thought it'd make sense to have alternate colors listed as backups in any case, but should I put more effort into detecting this case of "you've probably got the wrong color configured"?

dennisroberts71: a) I'm a little ambivalent about this. Automatically detecting and choosing a color that has services (or choosing all servers by default) certainly makes it easier to achieve zero-downtime deployments in the event of a misconfiguration. On the other hand, it also makes it slightly more likely an older version of the service will continue to be used after a deployment. It would also add some complexity to the template. Another advantage of _not_ building this check in is that we're likely to be notified quickly if service calls start producing errors. As long as we take pains not to have services deployed on nights or weekends, that may be preferable to having the wrong version of a service in use.

b) This case would be a little more difficult to troubleshoot once it's happened because nginx will still be running with an outdated configuration and it may or may not produce errors depending on whether or not the previously configured color has servers. If we could configure the template to detect this and alert us if this happens then that might be helpful.

ianmcorvidae: re: a. I think it sounds like there's more advantages of keeping it dumber; I'll let it just default to green arbitrarily like I have it now

re: b: Hm. We don't have a particularly built-up alerts infrastructure of any sort, which is what makes this harder, I think. The log will certainly include the "upstream has no servers" message from the `nginx -s reload` but I don't know that we have a way to propagate that to ourselves or anything. I think perhaps the thing to do is keep it how it is for now and pay attention when we start doing this, and change it once we have a better idea of what we need. It may be that in practice, this sort of misconfiguration will just always result in it becoming a force-502 eventually, since we'll eventually just take down the wrong-color servers anyway, and then we get the same quick notification from errors. So I think I'll leave it.

Thanks for your opinions on it!

PR #153, opened 2016-04-07
CORE-7597: fixed the format of user notifications posted to AMQP by t…

dennisroberts71:
…he notification agent


ianmcorvidae:
I don't quite understand what the change is, other than passing things through a reformatter (I guess the same one used for HTTP gets of notifications?) but it looks sensible to me.

dennisroberts71:
Yeah, this is the same reformatter used for notifications retrieved via HTTP. There are some formatting changes that were required by the UI that weren't being done and the notification ID was not being included.

dennisroberts71:
Thanks for the review.

PR #155, opened 2016-04-08
CORE-7602 Remove notifications and sys message polling. Add Websocket…

sri2k1us:
…s to push notifications and system messages to client using Atmosphere framework. This is first pass. I still have refactoring to do and tests add as well. But I thought I could use some reviews.Thanks!


ansible/inventories/example.cfg
```diff
@@ -1,189 +0,0 @@
-###############################################################################
```
ianmcorvidae: I'm guessing you didn't mean to remove this file.

ui/de-lib/src/main/java/org/iplantc/de/client/DEClientConstants.properties
```diff
@@ -2,4 +2,7 @@ fileUploadServlet = /de/secured/fileUpload
 fileDownloadServlet = /de/secured/fileDownload
 logoutUrl = j_spring_security_logout
 deHelpFile = https://pods.iplantcollaborative.org/wiki/display/DEmanual/
+notificationWS = /de/websocket/notifications
+sysmessageWS = /de/websocket/system-messages
+wsProtocol = ws://
```
ianmcorvidae: In prod, will this be `wss://` since we'd be moving things to SSL there? I'd guess it should be, though maybe that's a task for another patch.

ansible/inventories/example.cfg
sri2k1us: Good Catch!

ui/de-lib/src/main/java/org/iplantc/de/client/DEClientConstants.properties
sri2k1us: In that, I will move this property to de.properties file so that we can environment based protocol setting.

sri2k1us: @dennisroberts71 suggested inspecting current protocol (http vs https) and then deciding between ws and wss

ash3rz:
A lot of this is new to me, but I didn't catch any showstoppers.  Excited for this to get put in!

ui/de-lib/src/main/java/org/iplantc/de/client/DEClientConstants.properties
ianmcorvidae: An automatic inspection like that sounds like it'd make sense. Most clearly user-driven that way (not that we let people use plain http in prod anyway)!

dennisroberts71: My reasoning was a little more selfish: it will allow us to avoid adding a new configuration parameter, which will make testing the DE on our workstations easier. :)

dennisroberts71:
I didn't see any additional issues.

ash3rz:
The refactors look ok to me!

psarando:
Looks like something went wrong with the commits in this PR / srv-push branch. I see a bunch of commits duplicated from dev.
For example, commit df01dff44a4cad7345e045e073601b021150381a in this PR is a duplicate of 919d2d21d2cc1c9aa42b655b7a22c8844ea7ede7, and commit 35e982c225faa41d7260e8ee07665f79d737f458 in this PR is a duplicate of 60f3295b89355428b4f4209615d6dbc286c20752.

Looks like a rebase went awry, but since the changes are duplicates, it doesn't look like it broke anything, so it doesn't require any further actions.

PR #159, opened 2016-04-13
CORE-7168 CORE-7551 Fix db conversion versions.

psarando:


ianmcorvidae:
Just to make sure: what happens when/if this runs against something which already had the conversion applied under the old number?

Change looks fine if that'll work (or at least, if we've got a plan for how to handle it if it doesn't).

psarando:
These changes are only in dev, and our dev and QA environments always run against clean imports from production, so there shouldn't be a danger of this happening in our environments.

psarando:
So we couldn't do a version update like this for versions that are already in the `master` branch.

PR #160, opened 2016-04-13
Perms refactor

dennisroberts71:
This is the initial, bare-bones, implementation of the permissions service, (which currently serves up only a status endpoint). `go-swagger` pulls in a lot of dependencies, but I think that having the automatically generated validation and the Swagger UI will be worthwhile.

At this point in time, most of the code for the permissions service is generated. The exceptions are `permissions/swagger.yml`, `permissions/restapi/configure_permissions.go`, and everything underneath `permissions/restapi/impl`. The goal is to keep all service implementations underneath `permissions/restapi/impl` if possible.

The Swagger UI is not automatically packaged with services generated by `go-swagger`, so we're using a simple Swagger UI base image as the base of the permissions image. The Dockerfile and the custom source for the Swagger UI are in `docker/swagger-base`.


golang/src/permissions/restapi/configure_permissions.go
```diff
@@ -0,0 +1,67 @@
...
+// The middleware configuration is for the handler executors. These do not apply to the swagger.json document.
+// The middleware executes after routing but before authentication, binding and validation
+func setupMiddlewares(handler http.Handler) http.Handler {
+	return handler
+}
+
+// The middleware configuration happens before anything, this middleware also applies to serving the swagger.json
+// document. So this is a good place to plug in a panic handling middleware, logging and metrics
+func setupGlobalMiddleware(handler http.Handler) http.Handler {
+	return uiMiddleware(handler)
```
ianmcorvidae: Looking at these two functions, it looks initially like the `uiMiddleware` more obviously belongs in the other one (since it doesn't need to happen for serving `swagger.json` itself). I'm guessing this isn't the case because this set of middlewares happens before routing, where the other one happens after, and  we don't want to put `/docs` into the routing stuff since it's not really part of the API. Is this correct?

golang/src/permissions/restapi/embedded_spec.go
```diff
@@ -0,0 +1,9 @@
...
... 0x20, 0x31, 0xa, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x7d, 0x2c, 0xa, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x22, 0x73, 0x65, 0x72, 0x76, 0x69, 0x63, 0x65, 0x22, 0x3a, 0x20, 0x7b, 0xa, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x22, 0x74, 0x79, 0x70, 0x65, 0x22, 0x3a, 0x20, 0x22, 0x73, 0x74, 0x72, 0x69, 0x6e, 0x67, 0x22, 0x2c, 0xa, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x22, 0x6d, 0x69, 0x6e, 0x4c, 0x65, 0x6e, 0x67, 0x74, 0x68, 0x22, 0x3a, 0x20, 0x31, 0xa, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x7d, 0x2c, 0xa, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x22, 0x76, 0x65, 0x72, 0x73, 0x69, 0x6f, 0x6e, 0x22, 0x3a, 0x20, 0x7b, 0xa, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x22, 0x74, 0x79, 0x70, 0x65, 0x22, 0x3a, 0x20, 0x22, 0x73, 0x74, 0x72, 0x69, 0x6e, 0x67, 0x22, 0x2c, 0xa, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x22, 0x6d, 0x69, 0x6e, 0x4c, 0x65, 0x6e, 0x67, 0x74, 0x68, 0x22, 0x3a, 0x20, 0x31, 0xa, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x7d, 0xa, 0x20, 0x20, 0x20, 0x20, 0x20, 0x20, 0x7d, 0xa, 0x20, 0x20, 0x20, 0x20, 0x7d, 0xa, 0x20, 0x20, 0x7d, 0xa, 0x7d})
```
ianmcorvidae: Oh, generated code, you so silly sometimes :laughing:

golang/permissions.docker
```diff
@@ -0,0 +1,13 @@
+FROM discoenv/swagger-base
+
+COPY bin/permissions-server-linux-amd64 /bin/permissions
+
+ARG git_commit=unknown
+ARG buildenv_git_commit=unknown
+ARG version=unknown
+LABEL org.iplantc.de.templeton.git-ref="$git_commit" \
+      org.iplantc.de.templeton.version="$version" \
```
ianmcorvidae: These label names should be corrected.

dennisroberts71: Thanks for catching that.

docker/swagger-base/index.html
```diff
@@ -0,0 +1,122 @@
...
+  <script src='lib/handlebars-2.0.0.js' type='text/javascript'></script>
+  <script src='lib/underscore-min.js' type='text/javascript'></script>
+  <script src='lib/backbone-min.js' type='text/javascript'></script>
+  <script src='swagger-ui.js' type='text/javascript'></script>
+  <script src='lib/highlight.7.3.pack.js' type='text/javascript'></script>
+  <script src='lib/jsoneditor.min.js' type='text/javascript'></script>
+  <script src='lib/marked.js' type='text/javascript'></script>
+  <script src='lib/swagger-oauth.js' type='text/javascript'></script>
+
+  <!-- This file was copied and slightly modified from:
```
ianmcorvidae: I don't know if this is something that'd be worth doing, but in case we want to update this based on the swagger-api repository at some point, it might be nice to maintain a patch file of the changes we've made. I'm not really sure what those changes are, so this may or may not be an appropriate sort of thing to do :)

golang/src/permissions/restapi/configure_permissions.go
dennisroberts71: Yeah, that was the general idea. I don't think that `uiMiddleware` would work in `setupMiddlewares` because the routing is done before any of those middleware functions are called.

ianmcorvidae:
The only general comment I have is that I think we could use some notes in a `README` or something about exactly how the code generation happens (even just a link to the `go-swagger` repository). Basically, I'd like to know what to run, what to edit/add, and when, if I want to make a change to this service at some point.

The very basic stuff you've added yourself looks trivial and perfectly reasonable to go ahead and ship, aside from those image labels.

docker/swagger-base/index.html
dennisroberts71: Yeah, it might not hurt to do that. The changes were relatively trivial. I simply removed the form that allows the UI to be pointed to arbitrary Swagger-compliant APIs and added some code to assemble the URI to the `swagger.json` doc.

docker/swagger-base/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+    && git checkout v2.1.4 \
+    && mkdir -p /docs \
+    && cp -pr dist/* /docs/ \
+    && cd / \
+    && rm -rf /tmp/swagger-ui \
+    && apk del git \
+    && rm -rf /var/cache/apk/*
+
+COPY index.html /docs/index.html
+WORKDIR /
```
ianmcorvidae: very minor, this line seems unnecessary though, pretty sure this is default

dennisroberts71:
Thanks for the review. I'll fix the labels, add a README and merge the changes. As trivial as my changes were, it felt good to get it working with the Swagger UI. Hopefully development will move faster now that I have a basic idea of how to set up services.

docker/swagger-base/Dockerfile
dennisroberts71: Okay, I'll remove the line and try it out.

PR #162, opened 2016-04-15
Add a playbook which deploys service nginx containers

ianmcorvidae:
This is basically: copy in everything from `de-pull-images.yaml`, remove non-HTTP services, append `_nginx`; copy in everything from `de-start-containers.yaml`, remove non-HTTP services, append `_nginx`.

We don't need to do an explicit stop/remove with this because these containers don't use config containers and thus `docker-compose up` will do the right thing for them.


psarando:

ianmcorvidae:
Thanks for the review!

PR #165, opened 2016-04-20
Add new de-deploy-service role which is capable of deploying with zero downtime

ianmcorvidae:
Getting this up early so the core functionality can get reviewed/people can start learning about it.

This PR should be able to be merged at any point; everything's backwards-compatible (though we may need updates to automated deployments). Please also review the `zero-downtime` branch of the private ansible, however.

An example playbook would be something like the one I used for testing:

```
---
- name: Deploy anon-files
  hosts: anon-files:&systems
  become: True
  gather_facts: false
  tags:
    - services
    - anon-files
  roles:
    - de-deploy-service
  vars:
    service_name: "anon_files"
    deploy_use_color: "{{use_color|default(false)}}"
```

(for a service which doesn't use colors at all, `deploy_use_color` would be left off or set to false deliberately, presumably; `use_color` is expected to come from group_vars or extra_vars)

Work checklist for zero-downtime work generally:

- [x] Roles and playbooks for deploying consul and registrator, plus inventory etc. updates (in private ansible, merged)
- [x] docker-compose file updates for registrator environment variables (in private ansible, applied for dev but not yet other inventories)
- [x] Role for deploying services which can deploy both with and without color/zero-downtime functionality (this PR)
- [x] modify consul/registrator deploy playbook to use this role (in `zero-downtime` branch of private ansible)
- [x] modify grouper deploy playbook to use this role (in `zero-downtime` branch of private ansible)
- [x] modify service nginx deployment to use this role (this PR)
- [x] remove unused roles (de-services-cfg, de-ui) depending on old util-cfg-docker-* roles (this PR)
- [x] docker-compose file updates for green/blue colors (in private ansible)
- [x] group_vars updates (in private ansible, to track which inventories are set up for colors/zero-downtime)
- [x] Playbooks for deploying services using this role
- [x] remove util-cfg-docker-* roles


ansible/roles/de-deploy-service/tasks/swap-service.yaml
```diff
@@ -0,0 +1,44 @@
...
+- include: restart-service.yaml
+  vars:
+    restart_service_name: "{{service_name}}_{{other_color}}"
+
+- name: wait for {{other_color}} server to be up
+  uri: url=http://localhost:8500/v1/health/service/{{service_name_short}}?passing&tag={{environment_name}}-{{other_color}} method=GET return_content=yes
+  register: health_check
+  until: health_check.content != "[]"
+  retries: 25
+  delay: 2
```
ianmcorvidae: These values are one of the things I'd like opinions on, as far as how many retries and how many seconds between. This current setup gives it 50 seconds to come up, but I wouldn't see any issue with having it retry more frequently or try more times as long as it still gets at least this long to succeed.

ianmcorvidae:
Oh, note: this doesn't do zero-downtime for the UI deployment yet. Since that nginx is a more complicated beast I wanted to get the services done first, then circle back to set that nginx up for automatic switchovers and such.

strootman:
Where is the documentation?

ianmcorvidae:
The process of deployment is unchanged, so I wasn't really sure what documentation needed to be updated. I guess perhaps a conceptual document on how it works?

ansible/roles/de-deploy-service/tasks/swap-service.yaml
dennisroberts71: I think that two seconds between service checks is fine. I wouldn't be opposed to adding more retries, but I don't know if that's necessary or not.

psarando: Why not give it a full minute or 2 to come up? So 30 retries with 2~4 seconds between each?
But 50 seconds is close enough, so I don't mind sticking with it.

psarando:
I didn't catch any issues, so I vote :shipit:

strootman:
Comments sent in email. 
:shipit: with caveats mentioned.

ianmcorvidae:
Thanks for the reviews, everyone. I'll be merging this after `2.7` makes it into `dev`, and possibly after `dev` goes into `beta` since this only affects dev for now anyway.

psarando:
I noticed that `use_color: true` was added to a group_var file in our private de-ansible repo.
So rather relying on ` | default(false)` everywhere this var is used, we could add `use_color: false` to `ansible/inventories/group_vars/all` in this branch/PR.

PR #166, opened 2016-04-22
Create an image-janitor service

johnworth:
The current method of cleaning Docker images on the Condor cluster introduces a race condition where one job can accidentally kill another job by removing an image used by both jobs. The removal causes issues when it occurs after the killed job pulls the image but before it creates a container from it.

Image-janitor is supposed to fix that by periodically grabbing job definition files out of a directory, parsing out a list of all Docker images in use by those jobs and only removing images that are not listed by those jobs.

Road-runner has been modified to copy the job JSON files to image-janitor's monitored directory and to remove those files right before the job exits.

This pull request does include tests, but does not include the Ansible changes needed to deploy image-janitor.


strootman:
:shipit:

golang/src/image-janitor/main.go
```diff
@@ -0,0 +1,255 @@
...
+	cfgPath       = flag.String("config", "/etc/jobservices.yml", "Path to the config.")
+	readFrom      = flag.String("read-from", "/opt/image-janitor", "The directory that job files are read from.")
+	dockerURI     = flag.String("docker", "unix:///var/run/docker.sock", "The URI for connecting to docker.")
+	gitref        string
+	appver        string
+	builtby       string
+	filenameRegex = regexp.MustCompile(`(?i)^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}\.json$`)
+)
+
+func init() {
```
ianmcorvidae: There's no `logcabin.Init` done here like in the other go services, so I'd suspect the logging doesn't work correctly.

golang/src/dockerops/containers.go
```diff
@@ -159,6 +168,26 @@ func (d *Docker) NukeImage(name, tag string) error {
...
+	opts := docker.ListImagesOptions{
+		All: true,
+	}
+	apiImages, err := d.Client.ListImages(opts)
+	if err != nil {
+		return nil, err
+	}
+	var retval []string
+	for _, img := range apiImages {
+		repos := img.RepoTags
```
ianmcorvidae: If I'm not mistaken, this will mean that certain situations will leave around old images or break this logic. Specifically:

* image A:latest is used by a job
* A:latest is pushed with something new (this happens at least with curl-wrapper and backwards-compat, if not always frequently)
* another job uses A:latest, pulling the new image
* the old image that was previously A:latest is now without any RepoTags

This might be handled by the existing docker cleanup of dangling images that we do, but the ansible stuff for that is currently broken and we haven't really been maintaining it (which means it's not on newly-provisioned or reprovisioned nodes), so we might want to get it right here.

PR #170, opened 2016-04-27
CORE-6259: migrate data-info/stat-by-uuid and data-info/owns? usages

ianmcorvidae:
This changes the data-info client's `stat-by-uuid` to use data-info rather than jargon directly. It also removes both uses of its `owns?` function (used, incidentally, the same two places) to use the `permission` value retrieved from data-info instead of calling out with jargon to check the ownership permission.

This doesn't close the ticket, which is quite large, but moves a few more things to properly use data-info instead of jargon calls in terrain.

### DIFF
--- a/services/terrain/src/terrain/clients/data_info.clj
+++ b/services/terrain/src/terrain/clients/data_info.clj
@@ -353,19 +353,6 @@
     (throw+ {:error_code error/ERR_NOT_FOUND :uuid data-id})))
 
 
-(defn ^Boolean owns?
-  "Indicates if a file or folder is owned by a given user.
-
-   Parameters:
-     user       - the username of the user
-     data-path - The absolute path to the file or folder
-
-   Returns:
-     It returns true if the user own the data item, otherwise false."
-  [^String user ^String data-path]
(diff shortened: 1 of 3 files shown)


services/terrain/src/terrain/clients/data_info.clj
```diff
@@ -286,8 +287,11 @@
    Returns:
      It returns a path-stat map containing an additional UUID field."
   [^String user ^UUID uuid]
-  (uuids/path-for-uuid user uuid))
-
+  (-> (raw/collect-stats user :ids [uuid])
+      :body
+      json/decode
+      (get-in ["ids" uuid])
+      walk/keywordize-keys))
```
ianmcorvidae: This code changes the return value very slightly, to no longer duplicate the object ID in both `:id` and `:uuid` keys. The latter is removed (as it was only previously added by `uuids/path-for-uuid`, and is not used by any of the callers of this method).

psarando:
Looks good :shipit:

dennisroberts71:
I'll second that. :shipit:

PR #171, opened 2016-04-28
App Catalog refresh (ontology) services

psarando:
These are the initial database conversions and endpoints in the metadata, apps, and terrain services for the App Catalog refresh.
The terrain API docs can be viewed under the app metadata [Save an Ontology XML Document, Set Active Ontology Version, Save an Ontology Hierarchy, Listing Ontologies and Ontology Hierarchies](https://github.com/cyverse/DE/blob/gh-pages-app-catalog-refresh/api/endpoints/app-metadata.md#save-an-ontology-xml-document), [Listing Apps in an App Group](https://github.com/cyverse/DE/blob/gh-pages-app-catalog-refresh/api/endpoints/app-metadata.md#listing-apps-in-an-app-group), and [Managing App AVU Metadata](https://github.com/cyverse/DE/blob/gh-pages-app-catalog-refresh/api/endpoints/app-metadata.md#managing-app-avu-metadata) sections.

The Swagger API docs for the metadata and apps service endpoints can be previewed on `lana`.

## Endpoint workflow description

1. First an admin must upload an ontology XML document, which will be stored in the metadata database, and later parsed with the OWLAPI libraries in order to build the app categories. An EDAM_1.14.owl XML document has been attached to CORE-7389 for testing these endpoints.

    [POST /admin/ontologies](https://github.com/cyverse/DE/blob/gh-pages-app-catalog-refresh/api/endpoints/app-metadata.md#save-an-ontology-xml-document)

1. Once 1 or more ontology XML documents have been uploaded, the details of each available ontology may be listed. Each uploaded document is stored with a unique "version" ID. This version ID is used in the remaining endpoints in order to save and list ontology classes and hierarchies parsed from the associated XML document.

    [GET /admin/ontologies](https://github.com/cyverse/DE/blob/gh-pages-app-catalog-refresh/api/endpoints/app-metadata.md#listing-ontologies-and-ontology-hierarchies)

1. Next an admin must choose which hierarchies (and the ontology classes in them) will be available for use in the DE.

    [PUT /admin/ontologies/{ontology-version}/{root-iri}](https://github.com/cyverse/DE/blob/gh-pages-app-catalog-refresh/api/endpoints/app-metadata.md#save-an-ontology-hierarchy)

1. Finally, the admin sets an ontology version as the default, "active" version used by the DE when listing hierarchies or ontology classes for regular users.

    [POST /admin/ontologies/{ontology-version}](https://github.com/cyverse/DE/blob/gh-pages-app-catalog-refresh/api/endpoints/app-metadata.md#set-active-ontology-version)

    This active version is used by the apps service in the following endpoints:
    [GET /apps/hierarchies](https://github.com/cyverse/DE/blob/gh-pages-app-catalog-refresh/api/endpoints/app-metadata.md#listing-ontologies-and-ontology-hierarchies)
    [GET /apps/hierarchies/{root-iri}](https://github.com/cyverse/DE/blob/gh-pages-app-catalog-refresh/api/endpoints/app-metadata.md#listing-ontologies-and-ontology-hierarchies)
    [GET /apps/hierarchies/{root-iri}/unclassified](https://github.com/cyverse/DE/blob/gh-pages-app-catalog-refresh/api/endpoints/app-metadata.md#listing-apps-in-an-app-group)

1. An admin may add hierarchies for other ontology versions and preview their filtered hierarchies and app listings, without affecting the active version in use by normal users, with the following endpoints:

    [GET /admin/ontologies/{ontology-version}](https://github.com/cyverse/DE/blob/gh-pages-app-catalog-refresh/api/endpoints/app-metadata.md#listing-ontologies-and-ontology-hierarchies)
    [GET /admin/ontologies/{ontology-version}/{root-iri}](https://github.com/cyverse/DE/blob/gh-pages-app-catalog-refresh/api/endpoints/app-metadata.md#listing-ontologies-and-ontology-hierarchies)
    [GET /admin/ontologies/{ontology-version}/{root-iri}/unclassified](https://github.com/cyverse/DE/blob/gh-pages-app-catalog-refresh/api/endpoints/app-metadata.md#listing-apps-in-an-app-group)

1. Admins or users with app "write" permissions may categorize those apps under ontology classes by attaching the class IRIs as metadata with the following endpoints:

    [POST /admin/apps/{app-id}/metadata](https://github.com/cyverse/DE/blob/gh-pages-app-catalog-refresh/api/endpoints/app-metadata.md#managing-app-avu-metadata)
    [PUT /admin/apps/{app-id}/metadata](https://github.com/cyverse/DE/blob/gh-pages-app-catalog-refresh/api/endpoints/app-metadata.md#managing-app-avu-metadata)
    [POST /apps/{app-id}/metadata](https://github.com/cyverse/DE/blob/gh-pages-app-catalog-refresh/api/endpoints/app-metadata.md#managing-app-avu-metadata)
    [PUT /apps/{app-id}/metadata](https://github.com/cyverse/DE/blob/gh-pages-app-catalog-refresh/api/endpoints/app-metadata.md#managing-app-avu-metadata)

1. Apps with an ontology class attached as metadata (regardless of ontology version) may be listed with the following endpoint:

    [GET /apps/hierarchies/{class-iri}/apps](https://github.com/cyverse/DE/blob/gh-pages-app-catalog-refresh/api/endpoints/app-metadata.md#listing-apps-in-an-app-group)


services/apps/src/apps/routes/admin.clj
```diff
@@ -153,6 +156,58 @@
...
+        :query [params AppListingPagingParams]
+        :return AppListing
+        :summary "List Unclassified Apps"
+        :description
+"Lists all of the apps that are visible to the user that are not under the given `root-iri`, or any of
+ its subcategories, for the given `ontology-version`.
+
+#### Delegates to metadata service
+    POST /ontologies/{ontology-version}/{root-iri}/filter-unclassified"
+        (ok (coerce! AppListing
```
ianmcorvidae: I guess this is consistent with other parts of apps, but I believe this can just be done with the `:coercion` key on the route as long as `apps.util.coercions/custom-coercion-matcher` is made public (perhaps it belongs in common-swagger-api?), which might be nice to move toward rather than using the full reimplementations of `coerce!` etc. from that namespace. Not completely related to the actual work at hand though, just food for thought.

services/apps/src/apps/routes/admin.clj
```diff
@@ -153,6 +156,58 @@
...
+
+  (GET* "/" []
+        :query [params SecuredQueryParams]
+        :return ActiveOntologyDetailsList
+        :summary "List Ontology Details"
+        :description
+"Lists Ontology details saved in the metadata service.
+
+#### Delegates to metadata service
+    GET /ontologies"
```
ianmcorvidae: Another "might be nice" sort of comment -- in data-info we have `get-error-code-block` which we use for automatically and consistently building the "Error Codes:" blocks in the swagger docs. It might be nice to create a similar helper function for delegations like this.

psarando: I thought about that too, but wasn't sure if just using this `HEREDOC` style of string was more readable than `str` concatenation with a helper-function in the middle.

ianmcorvidae: Yeah, I'm not sure I like using `str` either, but I also don't like how repetitive these definitions are. I feel like my ideal one would be a more middleware-style invocation, such that it's something like `(delegated-endpoint-description ["metadata" :get "/ontologies"] "Lists Ontology details saved in the metadata service.")` (or something like that, details fuzzy). Not hugely important if we can't find a nice way to do it though; obviously this works, it's just a bit repetitive.

services/apps/src/apps/clients/metadata.clj
```diff
@@ -0,0 +1,66 @@
+(ns apps.clients.metadata
```
ianmcorvidae: Yet more long-term comments! Because it wouldn't be a review from me without at least three. What would you think of us moving away from having a full client in all of these different services, and instead making shared client libraries? I'd expect the shared ones to basically be the "raw" libraries we have for some things, and each service could still have a wrapper which adds any pre/post-processing features they might need. Ideally we'd probably make these be generated clients off of the swagger docs, where we have those.

To put this in the context I'm working from, upcoming data-info migration work for me will involve adding clients to data-info: an apps client for urlupload, a tree-urls client for the manifest, and possibly adding more to the metadata client to move some extra functionality out of terrain too. And of course, all that migration work has constantly added more stuff to terrain's data-info client, which I'd love to mostly just generate instead.

databases/metadata/src/main/functions/04_ontology_hierarchy.sql
```diff
@@ -0,0 +1,32 @@
+SET search_path = public, pg_catalog;
+
+--
+-- A function that returns all of the ontology subclasses of the class with the given
+-- ontology_version and IRI.
+--
+CREATE OR REPLACE FUNCTION ontology_class_hierarchy(VARCHAR, VARCHAR)
```
ianmcorvidae: In the interest of keeping on top of performance, I'd love to see an `EXPLAIN ANALYZE` of this with some loaded data.

services/apps/src/apps/clients/metadata.clj
```diff
@@ -0,0 +1,66 @@
...
+  {:query-params     params
+   :as               :stream
+   :follow-redirects false})
+
+(defn- post-options
+  [body params]
+  {:query-params     params
+   :body             body
+   :content-type     :json
+   :as               :stream
```
dennisroberts71: Just a small convenience: if you use `:as :json` instead then `clj-http` will parse the JSON and store the resulting data structure in the `:body` key.

services/apps/src/apps/clients/metadata.clj
```diff
@@ -0,0 +1,66 @@
...
+  (:require [cemerick.url :as curl]
+            [cheshire.core :as json]
+            [clj-http.client :as http]
+            [apps.util.config :as config]
+            [apps.util.service :as util]))
+
+(defn- metadata-url-encoded
+  [& components]
+  (str (apply curl/url (config/metadata-base) (map curl/url-encode components))))
+
```
dennisroberts71: `get-options` and `post-options` make the rest of the code in this namespace a lot more concise. 👍

ash3rz:
I'm not the best to review the code changes, but I can confirm that the changes are working as expected for the UI 👍

services/metadata/src/metadata/services/avus.clj
```diff
@@ -141,6 +141,56 @@
...
+       (persistence/remove-orphaned-avus target-type target-id)))
+
+(defn set-avus
+  "Sets AVUs for the given user's data item."
+  [user-id target-type target-id {avus :avus}]
+  (transaction
+   (remove-orphaned-avus target-type target-id avus)
+   (doseq [avu avus]
+     (persistence/add-or-update-avu user-id
+                                    (format-avu target-type target-id avu))))
```
psarando: The new `set-avus` metadata endpoint will add/update AVUs differently than the current `POST /filesystem/data/{data-id}/avus` endpoint, since it's now a requirement that we allow multiple AVUs on a target to share the same attribute with different values (this is the only way to categorize an app in multiple categories with AVUs, but it's also been requested for data AVUs).

The current data "Set AVUs" endpoint attempts to update an AVU based on either its ID or its attribute name. The new `set-avus` endpoint instead assumes that only AVUs with IDs in the request body require an update, and AVUs without IDs should be added as new.

So this new `set-avus` endpoint first deletes all AVUs on the target that do not have a matching ID in the request, before updating or adding new AVUs. This will help prevent "unique constraint" postgres exceptions in the case where the client deletes an AVU, then updates an existing AVU with the same values in the deleted AVU, then submits this final state in 1 `set` request.

For example, suppose a target has the following AVUs:
```clojure
[{:id 1, :attr "a", :value "b", :unit "c"}
 {:id 2, :attr "d", :value "e", :unit "f"}]
```
Then the user decides to delete the AVU with `:id 2`, and before submitting a request to save the AVUs, updates the AVU with `:id 1` to `{:id 1 :attr "d" :value "e" :unit "f"}` in the client. If this function did not delete AVUs whose IDs were not included in the request first, then the AVU with `:id 2` would cause postgres to throw a "unique constraint" exception when it attempted to update the AVU with `:id 1` to the same values.

services/metadata/src/metadata/persistence/avu.clj
```diff
@@ -77,6 +114,35 @@
...
+                     (where (-> (select-keys avu (conj required-keys :id))
+                                (update :target_type db/->enum-val))))))))
+
+(defn add-or-update-avu
+  [user-id avu]
+  (if-let [matching-avu (find-matching-avu avu)]
+    matching-avu
+    (if-let [existing-avu (update-valid-avu user-id avu)]
+      existing-avu
+      (add-avus user-id [avu]))))
```
psarando: This new `add-or-update-avu` function allows the new Set/Update AVU endpoints to add a new AVU with an ID that doesn't already exist in the database.

This function also allows the new Set/Update AVU endpoints to not mark AVUs as updated if they haven't changed (the `modified_on` or `modified_by` fields won't be updated), using the new `find-matching-avu` function.

databases/metadata/src/main/functions/04_ontology_hierarchy.sql
psarando: Sure, here's an `EXPLAIN ANALYZE` for an `ontology_class_hierarchy` query using the Operation class as a root:
```sql
metadata=> EXPLAIN ANALYZE SELECT * FROM ontology_class_hierarchy('http://edamontology.org/1.14/2016/04/20/20-39-12.5', 'http://edamontology.org/operation_0004');
                                                           QUERY PLAN                                                           
--------------------------------------------------------------------------------------------------------------------------------
 Function Scan on ontology_class_hierarchy  (cost=0.25..10.25 rows=1000 width=96) (actual time=86.554..87.403 rows=738 loops=1)
 Planning time: 0.319 ms
 Execution time: 88.533 ms
(3 rows)
```

This [EDAM Ontology Operation class](http://bioportal.bioontology.org/ontologies/EDAM/?p=classes&conceptid=http%3A%2F%2Fedamontology.org%2Foperation_0004) has the larger hierarchy of the 2 that the admins plan to use for the new app category hierarchies.

This query was run against our dev db that had 15 versions of the same ontology hierarchies loaded, for a total of 13599 rows in the `ontology_hierarchies` table. This many ontologies and hierarchies are not likely to be loaded in production any time soon.

services/apps/src/apps/persistence/categories.clj
```diff
@@ -0,0 +1,18 @@
...
+          (values {:version version
+                   :applied_by (get-user-id username)})))
+
+(defn get-active-hierarchy-version
+  []
+  ((comp :version first)
+   (select :app_hierarchy_version
+           (fields :version)
+           (order :applied :DESC)
+           (limit 1))))
```
psarando: The last ontology version `applied` is considered the active version (last applied wins).

databases/metadata/src/main/functions/04_ontology_hierarchy.sql
ianmcorvidae: Ah, I should have been more specific, though once I say it I think you'll see why: the details of plans hide the details inside of functions, just showing "Function Scan", which is why I like to see how the internals of the functions perform separately.

But, given your same values and knowing where there's a dev DB set up, I can do it myself!:

```
                                                                                              QUERY PLAN                                                                                              
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 CTE Scan on subclasses  (cost=2490.71..2500.53 rows=491 width=96) (actual time=0.781..87.847 rows=738 loops=1)
   CTE subclasses
     ->  Recursive Union  (cost=0.82..2490.71 rows=491 width=98) (actual time=0.775..85.447 rows=738 loops=1)
           ->  Limit  (cost=0.82..155.94 rows=1 width=98) (actual time=0.768..0.769 rows=1 loops=1)
                 ->  Nested Loop Left Join  (cost=0.82..155.94 rows=1 width=98) (actual time=0.764..0.764 rows=1 loops=1)
                       Join Filter: ((h.subclass_iri)::text = (c.iri)::text)
                       ->  Index Scan using ontology_classes_pkey on ontology_classes c  (cost=0.41..8.43 rows=1 width=61) (actual time=0.172..0.172 rows=1 loops=1)
                             Index Cond: (((ontology_version)::text = 'http://edamontology.org/1.14/2016/04/20/20-39-12.5'::text) AND ((iri)::text = 'http://edamontology.org/operation_0004'::text))
                       ->  Index Only Scan using ontology_hierarchies_pkey on ontology_hierarchies h  (cost=0.41..147.50 rows=1 width=74) (actual time=0.585..0.585 rows=0 loops=1)
                             Index Cond: ((ontology_version = 'http://edamontology.org/1.14/2016/04/20/20-39-12.5'::text) AND (subclass_iri = 'http://edamontology.org/operation_0004'::text))
                             Heap Fetches: 0
           ->  Nested Loop  (cost=0.82..232.49 rows=49 width=98) (actual time=0.186..10.119 rows=138 loops=8)
                 ->  Nested Loop  (cost=0.41..125.78 rows=45 width=74) (actual time=0.146..4.165 rows=138 loops=8)
                       ->  WorkTable Scan on subclasses sc  (cost=0.00..0.20 rows=10 width=32) (actual time=0.001..0.117 rows=92 loops=8)
                       ->  Index Only Scan using ontology_hierarchies_pkey on ontology_hierarchies h_1  (cost=0.41..12.52 rows=4 width=74) (actual time=0.035..0.037 rows=1 loops=738)
                             Index Cond: ((ontology_version = 'http://edamontology.org/1.14/2016/04/20/20-39-12.5'::text) AND (class_iri = (sc.iri)::text))
                             Heap Fetches: 0
                 ->  Index Scan using ontology_classes_pkey on ontology_classes c_1  (cost=0.41..2.36 rows=1 width=61) (actual time=0.036..0.038 rows=1 loops=1103)
                       Index Cond: (((ontology_version)::text = 'http://edamontology.org/1.14/2016/04/20/20-39-12.5'::text) AND ((iri)::text = (h_1.subclass_iri)::text))
 Planning time: 3.589 ms
 Execution time: 89.081 ms
(21 rows)
```

Also at https://explain.depesz.com/s/abMk;

I'll look at this and then add any extra comments in an extra comment in a bit.

ianmcorvidae: As expected, looks like most of the work is in the recursive loop. But it's also quite fast, since everything's indexed by ontology version, so looks good :+1: 

Thanks for humoring me with this.

services/apps/src/apps/routes/domain/app/category.clj
```diff
@@ -83,3 +85,25 @@
   (-> AppCategoryRequest
       (->optional-param :name)
       (->optional-param :parent_id)))
+
+(defschema AppCategoryOntologyVersionDetails
+  {:version    (describe String "The unique version of the Ontology")
+   :applied_by (describe NonBlankString "The user that set this version as active")
+   :applied    (describe Date "The date this version was set as active")})
```
ianmcorvidae: Something to check -- many of our things return dates as integers (unix timestamps, I believe). I don't know if it'd potentially be worth maintaining that compatibility, or if I'm misremembering, or if we may as well just start doing this like this instead.

PR #172, opened 2016-04-28
Core 7445 display error message when file upload fails due to size limits.

sri2k1us:
…mits.


dennisroberts71:
Looks good to me. :shipit:

ianmcorvidae:
Should this be going into the `2.7` branch instead? `dev` would be for hotfixes at this point until May 10.

PR #173, opened 2016-04-28
CORE-7638: validate path readability in get-root so a nil permission can't seep through

ianmcorvidae:
I also refactored the file generally a bit to be more consistent/clear about what it's doing.

### DIFF
--- a/services/data-info/src/data_info/services/root.clj
+++ b/services/data-info/src/data_info/services/root.clj
@@ -15,6 +15,7 @@
 
 (defn- get-root
   [cm user root-path]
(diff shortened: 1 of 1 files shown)


services/data-info/src/data_info/services/root.clj
```diff
@@ -15,27 +15,26 @@
 
 (defn- get-root
   [cm user root-path]
+  (validators/path-readable cm user root-path) ;; CORE-7638; otherwise a 'nil' permission can pop up and cause issues
```
ianmcorvidae: (this is the line which fixes the ticket. the rest is the refactoring)

psarando:
Looks good to me :shipit:

PR #174, opened 2016-04-28
permissions service integration tests

dennisroberts71:
These are the remaining integration tests for the resource type endpoints in the permissions service.


ianmcorvidae:
Looks pretty reasonable to me. :shipit:

dennisroberts71:
Cool, thanks for the review.

PR #175, opened 2016-04-29
templeton: support multiple bindings on one queue so periodic mode doesn't need two queues

ianmcorvidae:
Just a small thing I hadn't bothered to do correctly the first time since this wasn't implemented in the AMQP client at that point and I wanted it to get done.


johnworth:
:shipit:

ianmcorvidae:
Cool, thanks for the review!

PR #176, opened 2016-04-29
Split tree urls client out from terrain and into its own library

ianmcorvidae:
This is somewhat in anticipation of using it in data-info (for the manifest endpoint), and somewhat as a way to solicit opinions on using this sort of a tactic more generally. I really dislike how many duplicated clients for our own services we maintain, and as we move further into microservice architecture this can only get worse without a bit of deliberate action.

### DIFF
--- a/libs/tree-urls-client/src/tree_urls_client/core.clj
+++ b/libs/tree-urls-client/src/tree_urls_client/core.clj
@@ -1,14 +1,26 @@
-(ns terrain.clients.tree-urls
-  (:use [terrain.util.config]
-        [clojure-commons.error-codes]
+(ns tree-urls-client.core
+  (:use [clojure-commons.error-codes]
         [slingshot.slingshot :only [throw+]])
   (:require [clj-http.client :as http]
             [cemerick.url :refer [url]]
+            [clojure.tools.logging :as log]
             [cheshire.core :as json]))
 
-(defn tree-urls-url
+(def ^:dynamic *tree-urls-base*
+  "Dynamic context to be used in generating URLs."
+  "http://localhost:60000")
+
+(defmacro with-tree-urls-base
+  "A helper macro to change *tree-urls-base* within its body."
+  [tree-urls-base & body]
+  `(let [tree-urls-base# ~tree-urls-base]
+     (binding [*tree-urls-base* tree-urls-base#]
+       ~@body)))
+
+(defn- tree-urls-url
   [sha1]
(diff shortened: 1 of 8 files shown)


services/terrain/src/terrain/core.clj
```diff
@@ -191,6 +198,7 @@
 
 (def secured-routes-handler
   (-> (delayed-handler secured-routes)
+      (wrap-routes wrap-tree-urls-config)
```
ianmcorvidae: I'm not completely sure I like this middleware approach, but some of it's complicated by the fact terrain isn't particularly granular. If we were using compojure-api the whole way down, I'm pretty sure we'd just add this in a `:middlewares` key on just the individual routes which need this configuration to be present.

libs/tree-urls-client/src/tree_urls_client/core.clj
```diff
@@ -1,14 +1,26 @@
-(ns terrain.clients.tree-urls
-  (:use [terrain.util.config]
-        [clojure-commons.error-codes]
+(ns tree-urls-client.core
```
ianmcorvidae: This is just copied from terrain for the most part, obviously. It's possible at some point we should switch to using `swagger-codegen` for projects which have swagger docs (and maybe maintain swagger.yaml files for the ones that don't? not sure), but the interfaces the generated clojure clients expose are somewhat different than what we've generally been doing, and tree-urls doesn't have swagger anyway.

services/terrain/src/terrain/core.clj
```diff
@@ -59,6 +60,12 @@
         (tc/with-logging-context {:user-info (cheshire/encode user-info)}
                                  (handler request))))))
 
+(defn- wrap-tree-urls-config
+  [handler]
+  (fn [request]
+    (with-tree-urls-base (config/tree-urls-base)
+      (handler request))))
```
dennisroberts71: At some point in the future, we'd probably want to put a middleware namespace in the tree-urls-client package (or some related package). The middleware function wouldn't have to change much:

```

ianmcorvidae: Ah, that's a good idea; that'd make it so I don't need to load the core namespace or `with-tree-urls-base` into the core namespaces of the services, which was one of the things that made me mildly uncomfortable about this setup.

dennisroberts71:
Looks good to me. :shipit: whenever you're ready.

PR #178, opened 2016-05-02
Cleanup duplicated 'docker' group_vars.

psarando:
Replaced `docker.registry.host` and `docker.registry.port` with `docker.registry.base`.
Please also review the `ansible-docker-vars-cleanup` branch in the private de-ansible repo, which is the only place where these `docker.registry.host|port` vars were used. The duplicated `docker.compose_path` was also removed from the private de-ansible repo's `group_vars/all`.


ianmcorvidae:
LGTM :shipit:

PR #179, opened 2016-05-02
(WIP) CORE-7627 App Catalog Refresh Belphegor UI

ash3rz:
(This is coupled with PR #171 changes for the backend)

This is the initial code change for the sci-analysts to be able to re-categorize apps from our old App Categories into the new App Hierarchies based off the EDAM ontology.  This code will be updated as more endpoints for the transition to using the EDAM ontology get added and also pending more information on the feature requirements from sci-analysts.

Right now, a user should be able to:
* Upload a new EDAM ontology XML file
* View the list of available ontology versions
* Save the Topic/Operation hierarchies
* View the saved hierarchies
* View the list of apps classified within a hierarchy or unclassified
* Set the "active" ontology version which will be used in the DE

Changes to come:
* Being able to categorize an app into the hierarchy
* Potentially a more dynamic way of choosing and saving hierarchies (right now the Topic and Operation IRIs are hard-coded in)


ui/de-webapp/src/main/resources/application.properties
```diff
@@ -0,0 +1,2 @@
+multipart.max-file-size=10000000
+multipart.max-request-size=10000000
```
ianmcorvidae: Isn't this stuff what `de-application.yaml` does? Or is this some sort of better approach to that (which, hopefully, we can still override in prod)?

sri2k1us: I don't think this file should be checked-in. This is needed just for local dev.

ash3rz: Good catch - I'll remove it

ianmcorvidae: If it's just for local dev but might stick around, please also add it to a `.gitignore` file somewhere or other!

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/ontologies/views/OntologiesViewImpl.java
```diff
@@ -0,0 +1,238 @@
...
+            @Override
+            public void onSelection(SelectionEvent<Ontology> event) {
+                Ontology selectedOntology = event.getSelectedItem();
+                if (selectedOntology == null) {
+                    publishButton.disable();
+                    saveHierarchy.disable();
+                } else {
+                    publishButton.enable();
+                    saveHierarchy.enable();
+                    fireEvent(new SelectOntologyVersionEvent(event.getSelectedItem()));
```
sri2k1us: or just 'selectedOntology'

ash3rz: Yup, oversight on my part after some refactoring.  Thanks for catching!

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/ontologies/views/dialogs/EdamUploadDialog.java
```diff
@@ -0,0 +1,219 @@
+package org.iplantc.de.admin.desktop.client.ontologies.views.dialogs;
```
sri2k1us: Is there way to reuse / refactor 'SimpleFileUploadDialog' class instead of duplicating the whole logic here ?

ash3rz: I could look into it.  For quick testing, it was easier to duplicate since `SimpleFileUploadDialog` had an extra service call to check if the file existed already in irods and I think the success/failed responses were different.

ui/de-lib/src/main/java/org/iplantc/de/desktop/client/presenter/DesktopPresenterWindowEventHandler.java
```diff
@@ -199,7 +200,7 @@ public void onRequestSimpleUpload(RequestSimpleUploadEvent event) {
                                        diskResourceServiceProvider.get(),
                                        eventBus,
                                        diskResourceUtil,
-                                       UriUtils.fromTrustedString(clientConstants.fileUploadServlet()),
+                                       UriUtils.fromTrustedString(clientConstants.ontologyUploadServlet()),
```
sri2k1us: Is this intentional ?

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/ontologies/presenter/OntologiesPresenterImpl.java
```diff
@@ -0,0 +1,287 @@
...
+                                   AdminCategoriesView.Presenter categoriesPresenter,
+                                   AdminAppsGridView.Presenter gridPresenter) {
+        this.serviceFacade = serviceFacade;
+        this.beanFactory = beanFactory;
+        this.treeStore = treeStore;
+        this.appearance = appearance;
+
+        this.categoriesPresenter = categoriesPresenter;
+        this.gridPresenter = gridPresenter;
+        listStore = gridPresenter.getView().getGrid().getStore();
```
sri2k1us: We should try avoid accessing view elements in presenter. Instead call a method  on view that does the operation. so instead of listStore.clear() you will call view.clear() which in turn clears the store.

ui/de-lib/src/main/java/org/iplantc/de/desktop/client/presenter/DesktopPresenterWindowEventHandler.java
ash3rz: Nope, another oversight from testing.  This should be removed now from the latest commit.

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/ontologies/presenter/OntologiesPresenterImpl.java
ash3rz: Makes a lot of sense - I'll add to a refactor commit shortly.

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/ontologies/views/OntologiesViewImpl.java
```diff
@@ -0,0 +1,238 @@
...
+                              @Assisted AppCategoriesView categoriesView,
+                              @Assisted AdminAppsGridView gridView) {
+        this.appearance = appearance;
+        this.treeStore = treeStore;
+        this.categoriesView = categoriesView;
+        this.gridView = gridView;
+
+        initWidget(uiBinder.createAndBindUi(this));
+
+        publishButton.disable();
```
sri2k1us: Can we refactor disable / enable operation into a method ? I see this operation done in few places...

ash3rz: Sure, this was a last minute thing I added when I realized I was clicking the buttons and nothing was happening because I hadn't selected an ontology yet. 😄 I figured I wouldn't be the only one to do that.

ash3rz:
Closing this PR so it can be reopened against the dev branch once the latest changes are complete.

PR #180, opened 2016-05-03
Core 7513

dennisroberts71:
This change grants additional privileges to administrative users. Administrators are now able to view the details of and modify the labels of private apps. Since I was modifying terrain, I also took this opportunity to remove support for direct CAS authentication to terrain.


ansible/playbooks/local-services-cfg.yml
```diff
@@ -31,107 +31,198 @@
...
+        - service_config
+        - de-ui
 
     - role: util-cfg-service
       src: "ui/nginx.conf.j2"
       dest: "{{service_conf_dir}}/nginx/nginx.conf"
       service_name_short: "{{de.service_name_short}}"
+      tags:
+        - service_config
+        - de-ui
```
ianmcorvidae: I've been very lax in reviewing this, sorry about that. If we want to completely match the tags we use for deployment, this (and the other UI ones) should be `ui` rather than `de-ui`.

ianmcorvidae:
aside from my one nitpicky comment, this looks good :ship:

ansible/playbooks/local-services-cfg.yml
dennisroberts71: D'oh! I could have sworn I saw a `de-ui` tag in the playbooks. I'll update it.

PR #183, opened 2016-05-06
Refactor configurate package to remove C global

johnworth:
This removes the dependency on the janky C global defined in the configurate package.

### DIFF
--- a/golang/src/condor-launcher/condor.go
+++ b/golang/src/condor-launcher/condor.go
@@ -46,6 +46,7 @@ import (
 	"text/template"
 	"time"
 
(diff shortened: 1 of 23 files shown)


ianmcorvidae:
There's perpetually more refactoring that can be done so I'll restrain myself from pointing it all out ;) the changes you've made look good. I'd prefer if you could remove lines instead of just commenting them out (which I noticed in some tests), because it's just clutter and is perfectly recoverable from the history if for some reason it's needed. Still, ready to ship I think.

dennisroberts71:
Looks good to me.

PR #184, opened 2016-05-06
CORE-7648: add DISTINCTs to some clj-icat-direct queries

ianmcorvidae:
These queries count incorrectly when a user has access to a file by way of multiple groups, which the original ticket (DE-1115) was triggering in the data_commons folder. Interestingly it seems to be resolved in the production data_commons folder but I could still reproduce this in a QA environment.

These added `DISTINCT` match the ones in the better-functioning queries lower in the file in the `queries` var.

I have no doubt that clj-icat-direct could use some substantial refactoring but I wanted to get this fixed first, and then maybe look at the technical debt.


dennisroberts71:
:shipit:

PR #185, opened 2016-05-06
Resources Endpoints

dennisroberts71:
This change adds CRUD operations for resources in the permissions database.


golang/src/permissions/restapi/impl/db/resource_types.go
```diff
@@ -83,7 +83,7 @@ func GetDuplicateResourceTypeByName(tx *sql.Tx, id *string, name *string) (*mode
 	defer rows.Close()
 
 	// Get the resource type.
-	var resourceTypes []*models.ResourceTypeOut
+	resourceTypes := make([]*models.ResourceTypeOut, 0)
```
ianmcorvidae: I have no doubt this is correct but for the sake of my own education, why this change?

ianmcorvidae:
LGTM. Seems like pretty straightforward CRUD stuff so as long as there's integration tests seems ready to ship.

golang/src/permissions/restapi/impl/db/resource_types.go
dennisroberts71: In the original statement, `resourceTypes` was initialized to `nil`. Normally, this wasn't a problem because appending to a nil slice produces a single-element slice. If there were no list items, however, then the `nil` was being translated to `null` in the JSON that is produced by the endpoint. Using `make` produces an actual empty slice so that the JSON will contain an empty array instead.

ianmcorvidae: Ah, cool, that makes sense. Thanks!

PR #186, opened 2016-05-09
Reimplement several functions in clj-jargon to use more specific queries

ianmcorvidae:
The primary goal of this is speeding up info-typer some. I was investigating why it's so slow and found that a call to `clj-jargon/attribute?` (to check if there's already an `ipc-filetype` AVU) was taking about 90ms most times it ran, which is longer than the actual file identification process. Looking deeper, what we were doing was getting all metadata for the object and then filtering it in clj-jargon down to just the one attribute, by way of a chain of functions.

This changes the chain so it requests only the AVUs it actually wants, from the start. I've also done the same for lookups by attribute + value, which were implemented in a similar fashion.

It seems to have worked, since a brief test afterwards showed the same call taking less than a quarter of the time.

### DIFF
--- a/libs/clj-jargon/src/clj_jargon/metadata.clj
+++ b/libs/clj-jargon/src/clj_jargon/metadata.clj
@@ -42,41 +42,51 @@
       (.findMetadataValuesForCollection collection-ao dir-path)
       (.findMetadataValuesForDataObject data-ao dir-path))))
 
-(defn get-attribute
-  "Returns a list of avu maps for set of attributes associated with dir-path"
-  [cm dir-path attr]
-  (validate-path-lengths dir-path)
-  (filter
-    #(= (:attr %1) attr)
-    (get-metadata cm dir-path)))
+(defn- get-metadata-by-query
+  [{^DataObjectAO data-ao :dataObjectAO
+    ^CollectionAO collection-ao :collectionAO
+    :as cm} path query]
+  (mapv avu2map
+    (if (is-dir? cm path)
+      (.findMetadataValuesByMetadataQueryForCollection collection-ao query path)
+      (.findMetadataValuesForDataObjectUsingAVUQuery data-ao query path))))
 
-(defn get-attributes
-  "Returns a list of avu maps for a set of attributes associated with dir-path."
-  [cm attrs path]
-  (validate-path-lengths path)
-  (filter
-    #(contains? attrs (:attr %1))
-    (get-metadata cm path)))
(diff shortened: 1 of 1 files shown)


libs/clj-jargon/src/clj_jargon/metadata.clj
```diff
@@ -42,41 +42,51 @@
...
+  "Returns a list of avu maps for a specific attribute associated with dir-path"
+  [{^DataObjectAO data-ao :dataObjectAO
+    ^CollectionAO collection-ao :collectionAO
+    :as cm} dir-path attr]
   (validate-path-lengths dir-path)
-  (filter
-    #(= (:attr %1) attr)
-    (get-metadata cm dir-path)))
-
-(defn get-attributes
```
ianmcorvidae: This function was entirely unused so I simply deleted it. If it's needed again, we can probably implement it with an `IN` query.

libs/clj-jargon/src/clj_jargon/metadata.clj
```diff
@@ -42,41 +42,51 @@
...
   [cm dir-path attr]
-  (validate-path-lengths dir-path)
   (pos? (count (get-attribute cm dir-path attr))))
 
 (defn attr-value?
   "Returns a truthy value if path has metadata that has an attribute of attr and
    a value of val."
   ([cm path attr val]
-    (attr-value? (get-metadata cm path) attr val))
+    (pos? (count (get-attribute-value cm path attr val))))
```
ianmcorvidae: This function accepts passing already-fetched metadata, so that's unchanged, but this line I wanted to work like `attribute?` and not require fetching the entire set of metadata.

libs/clj-jargon/src/clj_jargon/metadata.clj
```diff
@@ -42,41 +42,51 @@
...
+                attr) (AVUQueryElement/instanceForValueQuery
+                AVUQueryElement$AVUQueryPart/VALUE
+                AVUQueryOperatorEnum/EQUAL
+                (str val))]]
+    (get-metadata-by-query cm apath query)))
 
 (defn attribute?
   "Returns true if the path has the associated attribute."
   [cm dir-path attr]
-  (validate-path-lengths dir-path)
```
ianmcorvidae: `get-attribute` already checks this, we don't need to do it twice.

psarando:
Looks good! 👍 
:shipit:

PR #189, opened 2016-05-12
Rewrite tree-urls service in Go

johnworth:
tree-urls has been rewritten in Go to reduce its resource usage. The old tree-urls service has been removed, along with the tree-urls entity in kameleon.entities and the kameleon.tree-urls-queries namespace.

Unit tests have been included. Integration tests were added to our internal repository and both the old service and the new service were run against them successfully.

This implementation does not fix some of the existing issues in tree-urls. SHA1 validation is far too lenient and the tree-urls are doubly JSON-encoded. Fixing those would likely require data migrations and changes to upstream services, so they're being left alone for now.


dennisroberts71:
Looks good to me. :shipit:

PR #190, opened 2016-05-12
Rewrite saved-searches in Go

johnworth:
saved-searches has been rewritten in Go to reduce its resource usage. The old saved-searches service has been removed, along with the saved-searches entity in kameleon.entities and the kameleon.user-saved-searches-queries namespace.

Unit tests have been included. Integration tests were added to our internal repository and both the old service and the new service were run against them successfully.


golang/src/saved-searches/main_test.go
```diff
@@ -0,0 +1,497 @@
...
+	"net/http/httptest"
+	"reflect"
+	"strings"
+	"testing"
+)
+
+type MockDB struct {
+	storage map[string]map[string]interface{}
+	users   map[string]bool
+}
```
dennisroberts71: I like the idea of having a mock database access object.

dennisroberts71:
Looks good to me. :shipit:

PR #192, opened 2016-05-13
Perms refactor

dennisroberts71:
This change adds the CRUD endpoints for subjects (generally, users and groups) to the permissions service.


johnworth:
I've got no complaints as long as the tests pass. :shipit:

PR #193, opened 2016-05-16
CORE-7627 App Catalog Refresh Belphegor UI

ash3rz:
This is the continuation of PR #179 

This is the code change for the sci-analysts to be able to re-categorize apps from our old App Categories into the new App Hierarchies based off the EDAM ontology.

A user should be able to:

* Upload a new EDAM ontology XML file
* View the list of available ontology versions
* Save the root hierarchies they want accessible in the DE (e.g. Topic, Operation)
* View the saved hierarchies
* View the list of apps classified within a hierarchy or unclassified
* Categorize apps by
  * A dialog window that allows the user to check/uncheck hierarchies to apply to a single app
  * Drag and drop a hierarchy to one app
  * Drag and drop a single app or multiple apps to a hierarchy
* Set the "active" ontology version which will be used in the DE


ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/ontologies/model/OntologyHierarchyTreeStoreProvider.java
```diff
@@ -0,0 +1,62 @@
...
+    private static final String HIERARCHY_MODEL_KEY = "model_key";
+
+    @Override
+    public TreeStore<OntologyHierarchy> get() {
+        return new TreeStore<>(new ModelKeyProvider<OntologyHierarchy>() {
+            @Override
+            public String getKey(OntologyHierarchy item) {
+                String key = getHierarchyPathTag(item);
+                setChildrenParentTag(key, item);
+                return key;
```
ash3rz: Since IRI's aren't actually unique in a hierarchy (for example, something vague like `Analysis` with an IRI of topic_1234 could exist under `Biology` and `Computer Science`), the key for the TreeStore is the path to that specific hierarchy.
The path/key is stored in a tag. Each `OntologyHierarchy` has two tags - one for its parent (which is the parent's full path), and one for itself (which is the parent's path + its own label).

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/ontologies/presenter/OntologiesPresenterImpl.java
```diff
@@ -0,0 +1,422 @@
...
+        }
+        if (parent == null) {
+            ontologyUtil.addUnclassifiedChild(children);
+            treeStore.add(children);
+
+        } else {
+            treeStore.add(parent, children);
+        }
+
+        helperMap(children);
```
ash3rz: This helperMap, `iriToHierarchyMap`, is needed mostly because if the admin wants to know what hierarchies an App is already tagged with, the metadata saved only includes the IRIs, not the paths.  Since an IRI can exist in a hierarchy multiple times, there's a need for a mapping from IRI to all the different hierarchies/paths.

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/ontologies/views/dialogs/SaveHierarchiesDialog.java
```diff
@@ -0,0 +1,154 @@
...
+import com.sencha.gxt.widget.core.client.form.FieldLabel;
+import com.sencha.gxt.widget.core.client.form.TextField;
+import com.sencha.gxt.widget.core.client.toolbar.ToolBar;
+
+import java.util.List;
+
+/**
+ * @author aramsey
+ */
+public class SaveHierarchiesDialog extends IPlantDialog implements IsHideable {
```
ash3rz: `PublishOntologyDialog` and `CategorizeDialog` were small and simple enough that I didn't do a separate view or do a UI binding.  `SaveHierarchiesDialog` looks big, but it's a small dialog with just a field to type in a root IRI with buttons to either add or delete more fields if you need to add multiple root IRIs.

ui/de-lib/src/main/java/org/iplantc/de/client/models/avu/Avu.java
```diff
@@ -0,0 +1,41 @@
+package org.iplantc.de.client.models.avu;
+
+import org.iplantc.de.client.models.HasId;
+
+import com.google.web.bindery.autobean.shared.AutoBean.PropertyName;
+
+/**
+ * @author aramsey
+ */
+public interface Avu extends HasId {
```
ash3rz: I didn't see a generic AVU model anywhere, so I created this.

ui/de-lib/src/main/java/org/iplantc/de/client/util/OntologyUtil.java
```diff
@@ -0,0 +1,93 @@
...
+
+import com.google.common.collect.Lists;
+import com.google.gwt.core.client.GWT;
+
+import java.util.List;
+
+/**
+ * @author aramsey
+ */
+public class OntologyUtil {
```
ash3rz: I created this helper class so any `Presenter` class related to Ontologies could reuse these methods.  I also wanted to keep any messy Ontology-related String manipulation in one place, especially the Attributes, until we figure out what the long-term solution will be.

ui/de-lib/src/main/java/org/iplantc/de/commons/client/views/dialogs/AbstractFileUploadDialog.java
```diff
@@ -0,0 +1,235 @@
...
+import com.sencha.gxt.widget.core.client.form.FormPanelHelper;
+import com.sencha.gxt.widget.core.client.form.TextField;
+
+import java.util.List;
+
+/**
+ *
+ * @author aramsey
+ */
+public abstract class AbstractFileUploadDialog extends IPlantDialog {
```
ash3rz: For code-reuse, I created this abstract class for the file uploads.  So now any file upload just has to implement what validations happen when the user submits the form, and how the response is interpreted.  I'm open to suggestions here if there's a better, simpler way.

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/ontologies/views/OntologiesViewImpl.java
```diff
@@ -0,0 +1,351 @@
...
+    }
+
+    @UiHandler("addButton")
+    void addButtonClicked(SelectEvent event) {
+        new EdamUploadDialog(UriUtils.fromTrustedString(clientConstants.ontologyUploadServlet())).show();
+    }
+
+    @UiHandler("saveHierarchy")
+    void saveHierarchyClicked(SelectEvent event) {
+        treeStore.clear();
```
psarando: The treeStore should probably not be cleared until it's ready to be reloaded (when a service call successfully returns with a new hierarchy listing).
Clearing the store here means that the admin can click the "Save Hierarchies" button, then change their mind and click "Cancel" on the dialog, and the previously loaded hierarchies are cleared but not replaced with a new listing.

ash3rz: Ah, good catch 👍

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/ontologies/presenter/OntologiesPresenterImpl.java
```diff
@@ -0,0 +1,427 @@
...
+        }
+    }
+
+    @Override
+    public void onSaveOntologyHierarchy(SaveOntologyHierarchyEvent event) {
+        List<String> iris = event.getIris();
+
+        for (final String iri : iris) {
+            serviceFacade.saveOntologyHierarchy(event.getOntology().getVersion(),
+                                                iri, new SaveHierarchyAsyncCallback(iri));
```
psarando: The tree panel should probably have a "Loading" mask while the service calls are processed.
Otherwise, if the service call takes a while, the admin might not know if the request is still processing or not, or failed without an error message.

ash3rz: 👍 good call, I'll add for the next drop

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/ontologies/presenter/OntologiesPresenterImpl.java
```diff
@@ -0,0 +1,427 @@
...
+        }
+
+        view.showTreePanel();
+    }
+
+    @Override
+    public void onCategorizeButtonClicked(CategorizeButtonClickedEvent event) {
+        final App selectedApp = event.getSelectedApp();
+        final List<OntologyHierarchy> hierarchyRoots = event.getHierarchyRoots();
+        serviceFacade.getAppAVUs(selectedApp, new CategorizeCallback(selectedApp, hierarchyRoots));
```
psarando: This appears to also include the `Unclassified` categories.
Apps shouldn't be tagged with this category, so these `Unclassified` categories should be filtered out at some point from the categorization selection tree.

ash3rz: Another good catch - thanks!

psarando:
Whenever apps are re-categorized, either through drag-n-drop, or with the categorize dialog, the currently selected hierarchy's app listing should probably be refreshed, in case it's gained or lost any apps in the request.

Also, the currently selected hierarchy's app listing should be cleared when the Ontology version drop-down selection changes, since that category might not be in the next selected ontology.

PR #195, opened 2016-05-20
Core 7660

sri2k1us:


ui/de-lib/src/main/java/org/iplantc/de/client/services/DiskResourceServiceFacade.java
```diff
@@ -1,5 +1,7 @@
 package org.iplantc.de.client.services;
 
+import java.util.List;
```
ash3rz: I think your IntelliJ is missing one of the code styles that makes sure `org.iplantc.*` imports come first, FYI

ui/de-lib/src/main/java/org/iplantc/de/desktop/client/presenter/DesktopPresenterImpl.java
```diff
@@ -105,9 +106,23 @@
  */
 public class DesktopPresenterImpl implements DesktopView.Presenter {
 
-
-
-    interface AuthErrors {
+    private final class NewSysMessageCountCallback implements AsyncCallback<Counts> {
+		@Override
+		public void onFailure(Throwable caught) {
+			// TODO Auto-generated method stub
```
ash3rz: Not sure if this is still WIP, but we should do something if a failure occurs here I would think.

sri2k1us:
Review comments are addressed. Closing PR.

PR #196, opened 2016-05-23
Remove template and template_id references from metadata AVU endpoints

psarando:
These changes remove the metadata service `/filesystem/data/:id/avus` endpoints (or rename them to `/avus/:type/:id` versions) and update terrain and data-info services to use the new `/avus/:type/:id` endpoints added by PR #171.
So metadata AVUs added to data items are no longer tied to template instances.
A new `metadata-client` library has been added for use by both the apps and data-info services.

Additionally, almost all coordination of data-info and metadata service endpoints has been migrated into the data-info service (with the exception of the csv-parser endpoint).

#### Renamed Endpoints
| Original endpoint | New endpoint |
| --- | --- |
| `POST /avus/{target-type}/{target-id}` | `PUT /avus/{target-type}/{target-id}` |
| `PUT /avus/{target-type}/{target-id}` | `POST /avus/{target-type}/{target-id}` |
| `POST /filesystem/data/{data-id}/avus/copy` | `POST /avus/{target-type}/{target-id}/copy` |

* Set and Add/Update Metadata AVU endpoints have been swapped.

#### Removed Endpoints
* `GET /filesystem/data/{data-id}/avus`
* `POST /filesystem/data/{data-id}/avus`
* `DELETE /filesystem/data/{data-id}/avus/{template-id}`
* `GET /filesystem/data/{data-id}/avus/{template-id}`
* `POST /filesystem/data/{data-id}/avus/{template-id}`
* `DELETE /filesystem/data/{data-id}/avus/{template-id}/{avu-id}`

#### Renamed Endpoints
| Original endpoint | New endpoint |
| --- | --- |
| `POST /admin/apps/{app-id}/metadata` | `PUT /admin/apps/{app-id}/metadata` |
| `PUT /admin/apps/{app-id}/metadata` | `POST /admin/apps/{app-id}/metadata` |
| `POST /apps/{app-id}/metadata` | `PUT /apps/{app-id}/metadata` |
| `PUT /apps/{app-id}/metadata` | `POST /apps/{app-id}/metadata` |

* Set and Add/Update Metadata AVU endpoints have been swapped to match the corresponding metadata endpoints.

#### New Endpoint
* `POST /data/{data-id}/metadata/copy`

#### Renamed Endpoints
| Original endpoint | New endpoint |
| --- | --- |
| `GET /admin/data/{data-id}/avus` | `GET /admin/data/{data-id}/metadata` |
| `POST /admin/data/{data-id}/avus` | `PATCH /admin/data/{data-id}/metadata` |
| `GET /data/{data-id}/avus` | `GET /data/{data-id}/metadata` |
| `POST /data/{data-id}/avus` | `PATCH /data/{data-id}/metadata` |
| `PUT /data/{data-id}/avus` | `PUT /data/{data-id}/metadata` |

#### Removed Endpoints
* `DELETE /admin/data/{data-id}/avus`

#### Renamed Endpoints
| Original endpoint | New endpoint |
| --- | --- |
| `POST /admin/apps/{app-id}/metadata` | `PUT /admin/apps/{app-id}/metadata` |
| `PUT /admin/apps/{app-id}/metadata` | `POST /admin/apps/{app-id}/metadata` |
| `POST /apps/{app-id}/metadata` | `PUT /apps/{app-id}/metadata` |
| `PUT /apps/{app-id}/metadata` | `POST /apps/{app-id}/metadata` |

* Set and Add/Update App Metadata endpoints have been swapped to match the corresponding apps endpoints.


services/data-info/src/data_info/routes/avus.clj
```diff
@@ -13,25 +14,30 @@
     (GET* "/avus" [:as {uri :uri}]
       :query [{:keys [user]} StandardUserQueryParams]
       :return AVUGetResult
+      :middlewares [wrap-metadata-base-url]
       :summary "List AVUs (administrative)"
-      :description (str "List iRODS AVUs associated with a data item. Include administrative/system AVUs."
-(get-error-code-block "ERR_NOT_A_USER, ERR_NOT_READABLE"))
+      :description
+          (str "List all AVUs associated with a data item. Include administrative/system iRODS AVUs."
```
ianmcorvidae: minor, but I'd explicitly mention that this is more than just iRODS AVUs for each of these, especially since it's changing we could confuse ourselves for a while otherwise

services/data-info/src/data_info/routes/avus.clj
```diff
@@ -13,25 +14,30 @@
...
+(get-error-code-block "ERR_NOT_A_USER, ERR_NOT_READABLE, ERR_DOES_NOT_EXIST, ERR_NOT_WRITEABLE, ERR_NOT_AUTHORIZED")
+(get-endpoint-delegate-block "metadata" "PUT /avus/{target-type}/{target-id}"))
+      (svc/trap uri meta/admin-metadata-add data-id body))
 
     (DELETE* "/avus" [:as {uri :uri}]
       :query [{:keys [user attr value]} AVUDeleteParams]
       :return AVUChangeResult
       :summary "Delete AVU (administrative)"
-      :description (str "Delete a single AVU from a data item. Allows deleting any AVU."
+      :description (str "Delete a single iRODS AVU from a data item. Allows deleting any AVU."
```
ianmcorvidae: Why is this the one exception that can only do iRODS AVUs? seems sort of annoyingly inconsistent

services/data-info/src/data_info/routes/avus.clj
```diff
@@ -13,25 +14,30 @@
     (GET* "/avus" [:as {uri :uri}]
       :query [{:keys [user]} StandardUserQueryParams]
       :return AVUGetResult
+      :middlewares [wrap-metadata-base-url]
```
ianmcorvidae: This isn't always true but metadata will probably be relevant for everything here, so I think I'd say put this right into the `context*` instead of each of the routes. But I'd be receptive to other arguments.

services/data-info/src/data_info/services/metadata.clj
```diff
@@ -181,25 +196,82 @@
...
         (common-metadata-delete cm path del-avu))
       (doseq [avu irods-avus]
         (common-metadata-add cm path avu))
 
       {:path path
-       :user user
-       :type type})))
+       :user user})))
+
+(defn- find-attributes
```
ianmcorvidae: I think there's a more efficient way to do this with better primitives in clj-jargon but it might not matter for right now...

services/terrain/src/terrain/clients/data_info/raw.clj
```diff
@@ -200,20 +201,25 @@
...
+  (request :post ["data" path-uuid "avus" "copy"]
+           (mk-req-map user (json/encode copy-request) {:force force})))
 
 (defn admin-add-avus
   "Add AVUs, allowing administrative AVUs to be included, for a data item."
   [user path-uuid avu-map]
-  (request :post ["admin" "data" path-uuid "avus"]
-           (mk-req-map user (json/encode {:irods-avus avu-map}))))
+  (request :put ["admin" "data" path-uuid "avus"]
+           (mk-req-map user (json/encode avu-map))))
```
ianmcorvidae: Is this right? PUT should be a 'set' operation (replacing the resource) vs. POST which can be an addition (modifying the resource). This looks backwards to me.

services/data-info/src/data_info/routes/avus.clj
```diff
@@ -12,26 +13,29 @@
...
+    (PUT* "/avus" [:as {uri :uri}]
       :query [{:keys [user]} StandardUserQueryParams]
-      :body [{:keys [irods-avus]} (describe AVUListing "An list of AVUs to add")]
+      :body [{:keys [irods-avus]} (describe AVUListing "A list of AVUs to add")]
       :return AVUChangeResult
       :summary "Add AVUs (administrative)"
-      :description (str "Associate AVUs with a data item. Allow adding any AVU."
+      :description (str "Associate iRODS AVUs with a data item. Allow adding any AVU."
 (get-error-code-block "ERR_NOT_A_USER, ERR_NOT_READABLE, ERR_DOES_NOT_EXIST, ERR_NOT_WRITEABLE, ERR_NOT_AUTHORIZED"))
       (svc/trap uri meta/admin-metadata-add data-id irods-avus))
```
ianmcorvidae: Yeah, this is definitely backwards.

> The PUT method requests that the enclosed entity be stored under the supplied Request-URI. **If the Request-URI refers to an already existing resource, the enclosed entity SHOULD be considered as a modified version of the one residing on the origin server.**

(highlighting mine) 

vs.

>  The POST method is used to request that the origin server accept the entity enclosed in the request as a new subordinate of the resource identified by the Request-URI in the Request-Line. POST is designed to allow a uniform method to cover the following functions:
>
>      - Annotation of existing resources;
>
>      - Posting a message to a bulletin board, newsgroup, mailing list,
>        or similar group of articles;
>
>      - Providing a block of data, such as the result of submitting a
>        form, to a data-handling process;
>
>      - Extending a database through an append operation.

(noting "Annotation of existing resources" and "Providing a block of data to a data-handling process")

(from https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html)

The PUT and POST should have their original semantics from my migration, not be switched like this.

ianmcorvidae: The section on idempotent methods is also relevant: PUT should be idempotent, but POST need not be.

services/data-info/src/data_info/routes/avus.clj
psarando: I thought the change from `List iRODS AVUs` to `List all AVUs` captured that.
It is still a little vague, but I'm not sure how to word it less awkwardly :)

ianmcorvidae: Hm, it might do fine in aggregate with the other endpoints. There's documentation of the exact return values anyway, so I think I might have been more confused by it in the PR than I would be in reality.

services/data-info/src/data_info/routes/avus.clj
psarando: The data-info and metadata service `DELETE` endpoints accept very different requests.
The data-info service wants to delete by `attr` and `value`.
The metadata service wants to delete by a specific AVU ID.
I think refactoring this endpoint to support both just makes it outside the scope of this PR.

ianmcorvidae: Alright, that makes sense. I'd love a ticket, if just for long-term tracking, which mentions that this endpoint is inconsistent and should be fixed at some vague point, just so we don't lose track of this.

services/data-info/src/data_info/routes/avus.clj
psarando: Good catch, I'll update this.

services/data-info/src/data_info/services/metadata.clj
psarando: Also note that this is a copy of a function that's still left behind in terrain.services.filesystem.metadata for the metadata csv-parser endpoint.
Once that endpoint logic is migrated, the data-info version can be updated for this and the metadata-copy endpoint.

services/terrain/src/terrain/clients/data_info/raw.clj
psarando: The metadata endpoints were already using `POST` requests for `set` operations and `PUT` for `add/update`, so I just updated the data-info endpoints to be consistent with those.

services/data-info/src/data_info/routes/avus.clj
psarando: The argument that `PUT` should be idempotent is the reason that I thought `PUT` should only add/update AVUs (or maybe it should be a `PATCH` endpoint instead?).

Metadata AVUs without IDs in the `set` request will always have a new ID after each subsequent request, since only AVUs with IDs are updated, and every other AVU will be deleted/added by the request, so it seemed like the `set` endpoint is not idempotent.

The highlighted section in the `PUT` quote from the RFC seems to actually argue for the opposite in this case, since the `set` endpoint will be modifying the existing _list_ of AVUs, by setting it to the list given in the request, whereas the `add/update` endpoint will only update or append to the list. I've also seen it mentioned elsewhere that a `PUT` request should create or replace, so it makes sense to me that the `set` endpoint should be a `PUT` request.

I'm fine with switching the data-info endpoints back, and then switching the metadata PUT vs POST endpoints to be consistent with data-info instead, unless anyone has better arguments to keep it this way.

psarando: Do you have any other resource that says this is "definitely backwards"? I've read this RFC before, and based on its wording I tend to agree that `PUT` endpoints should be `set` operations and `POST` endpoint are more general add/update operations, but every time I try to look this up in other articles, I find slightly differing opinions. It seems the real difference between the two is the idempotent nature of `PUT`. Also, some of these [popular SO answers](http://stackoverflow.com/questions/630453/put-vs-post-in-rest) suggest that a `PUT` method should only be used with a URI that address a specific entity, and these metadata endpoints process lists of metadata AVUs rather a specific AVU.

In fact, I think I just convinced myself that these metadata `set` endpoints should be `POST` methods and the `add/update` endpoints should be `PATCH` methods in both the data-info and metadata services.

ianmcorvidae: The highlight I did was intended to show basically what you're saying, that PUT shouldn't be used when it's not a complete representation of the resource at that particular URI (in this case, the list of AVUs), so it needs to be set and not just add ("modified version of the one residing on the origin server" meaning "a complete representation of the resource", since this is REST). PATCH would... probably work for add/update, but it's not really well-defined in our code or generally how PATCH should be implemented so most of the time POST makes more sense (since POST is sort of the catch-all sending some data to a data handling process). Plus, we don't support deletes with this sort of endpoint so it's not a complete patch.

I'd argue that the set is still idempotent with respect to meaningful values in this case. Since the new IDs various AVUs get are random, the preservation of a particular random value doesn't really affect the semantics of idempotence -- as long as after N calls that value is still a new, random value, it's still indistinguishable from running it only once (the fact that it was a different random value in the middle isn't meaningfully different). Arguably we either have useless IDs (we don't actually care much about preserving them as far as I know; other than figuring out how to address endpoints concerning single AVUs I don't think we need them, and we could just change that) or we should change the implementation to preserve IDs even when they aren't provided if there's an appropriate non-surrogate identifier (possibly all three values of the AVU, in this case -- for iRODS AVUs that's good enough but I'm not completely sure of the semantics of our metadata). Either change would make that endpoint properly idempotent.

Basically, I think the idempotence issue is an artifact of the fact we're using sort of poorly defined surrogate identifiers, and the other guidance should be referred to instead.

psarando: It was mentioned in PR #171 that if the client wants to update a metadata AVU, and preserve its `created_*` fields, then it needs to reuse the AVU's ID in the set request. So while we don't have endpoints that address a specific AVU by ID (other than the unused DELETE endpoint), the AVU IDs still have meaning in the set and update endpoints.

> or we should change the implementation to preserve IDs even when they aren't provided

It was addressed in PR #171 why it's not practical to handle this in the `set` endpoint, since it also handles batch AVU deletes in the same request.

> Plus, we don't support deletes with this sort of endpoint so it's not a complete patch.

Is this a requirement? After reviewing our /apps endpoints, it looks like we were already using `POST /apps` to create an app, `PUT /apps/:id` to update/replace a private app (which handles deleting app groups and properties like this metadata `set` endpoint), and `PATCH /apps` to only add/update fields given in the request and leave other fields alone.

I think I'm still leaning towards changing these metadata endpoints to follow the same pattern and use at least `PATCH` for the add/update endpoint. I have no strong opinion against using `PUT` for the `set` endpoint, and that would also be consistent with your arguments and what we're already using in the /apps endpoints.

services/data-info/src/data_info/routes/avus.clj
psarando: On second thought, these DELETE endpoints only address one specific AVU by ID at a time (in iRODS we can only address an AVU by attr/value), so I don't think the data-info endpoint should be changed to handle batch deletes, since the `set` endpoint already handles that.

In fact, I don't think these endpoints are in use, so they should probably just be removed for now.

services/data-info/src/data_info/routes/avus.clj
ianmcorvidae: I don't have strong feelings about PATCH, ultimately. I don't like it because there doesn't seem to be much of a consensus, so I tend to be conservative and use POST instead, but I wouldn't be opposed to using it either (my complaints are more aesthetic than practical). The point about deletes was mostly just because that's how I tend to think about PATCH, but nobody seems to really agree on PATCH, so it's not a requirement, just a feeling! Sorry I wasn't that clear about that.

Mostly, the only thing that is clearly wrong to me is using PUT for something which does not send an entire resource representation (the whole list) at once, which I think you agree on.

As for the rest: since we want the same URI, they can't both be POST. I think PUT is clear enough/idempotent enough to be used for the set endpoint (which is what most things in our code, and externally from what I've seen, use PUT for). POST would also be basically "fine" for the set endpoint, but it's sort of a cop-out on using HTTP verbs to denote semantics. I don't have a strong opinion about whether the add/update should be POST or PATCH, though I'd choose POST by the conservatism arguments I list above and because I prefer PUT for the set endpoint it's available.

Basically, I put these into data-info as PUT for set and POST for add+update because that's my preference, but the only arrangement that I see as definitely wrong is using PUT for add+update. set/add as PUT/POST, POST/PATCH, and PUT/PATCH all seem, if not perfect, at least not wrong. PUT/PATCH is probably the most correct, where PUT/POST is the most if you exclude PATCH due to ill-definedness like I tend to.

ianmcorvidae: re: the SO answer: this *is* a case with a specific ID, because it's working on the aggregate collection (whose ID is of the data-object, not the AVU). What they mean there is: if you don't have a specific ID (or want one generated, most often), then you need to use the nearest specific URI you have (usually the direct parent resource) but then PUT is inappropriate because you're not modifying *that* resource, you're creating a subresource instead, so use POST (especially since the RFC says "The POST method is used to request that the origin server accept the entity enclosed in the request as a new subordinate of the resource identified by the Request-URI in the Request-Line."). Their example makes this clear.

So, the PUT/POST distinction they're making would be between something like `POST /avus` to add an AVU with an auto-generated ID (non-specific, so use a POST to the parent resource) vs. `PUT /avus/:id` to add one with a specific AVU ID (specific, so use a PUT to the specific resource). So in fact what they're saying is exactly why I used POST for the add-avus endpoint: it's creating subresources.

It's weird because we're addressing a specific entity which is a collection of subresources, and the answer didn't specify that what they meant was "if you don't have a specific ID to PUT to, you have to use the specific ID of the parent resource and then POST is appropriate because you're creating a subresource". They didn't address the notion of modifying child resources by replacing the entirety of their parent, which is what our set endpoint does. But it's still replacing a resource with new contents, which is PUT.

psarando: It also doesn't help that this whole time I've been thinking of the `/filesystem/:id/metadata` and `/apps/:id/metadata` endpoints. You're points make a whole lot more sense when referring solely to the data-info `/data/:id/avus` endpoints.

Now that these endpoints are managing more than just iRODS AVUs, I think it might be best to rename them to be consistent with the corresponding apps endpoints: GET/PUT/PATCH `/data/:id/metadata`.

ianmcorvidae: Agreed. I'd used `avus` because it was just iRODS, but `metadata` is a better term for doing both.

services/data-info/src/data_info/routes/avus.clj
psarando: This doesn't appear to work, at least not with the version of compojure-api we're using.
We should revisit this once we update to the latest version of compojure-api.

PR #197, opened 2016-05-24
CORE-7672: add batch processing group to job JSON in apps

ianmcorvidae:
Please review carefully; I'm not sure if this is the right place to be adding this.

Note that this won't work still, for now, because the HTCondor configuration is not correct. But this is a first step.


johnworth:
That looks like the right spot to me. I say :shipit:, it would be easy to modify if it turns out that it needs to be changed later.

PR #198, opened 2016-05-24
CORE-7669: zero-downtime UI deployment

ianmcorvidae:
* move de-ui deployment to color section
* move de-ui-nginx to service-nginx deployment

there is also a private ansible branch for this, which updates the docker-compose files (more than I already had to get the nginx working initially, which is already in the dev environment) and to the documentation (to mention the de-deploy-service-nginx.yaml playbook and when it should be run in deployments (i.e., when downtime is acceptable)

### DIFF
--- a/ansible/playbooks/de-deploy-services.yaml
+++ b/ansible/playbooks/de-deploy-services.yaml
@@ -257,6 +257,23 @@
     - role: util-notify-chat
       msg: ":heavy_check_mark: Done deploying user-sessions"
 
+- name: Redeploy de-ui
+  hosts: ui:&systems
+  become: true
+  gather_facts: false
+  tags:
+    - ui
+    - colored
+  roles:
+    - role: util-notify-chat
+      msg: "Deploying UI"
+    - role: de-deploy-service
+      has_iplant_data: true
+      deploy_use_color: true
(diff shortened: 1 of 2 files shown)


ianmcorvidae:
https://github.com/cyverse/DE/commit/5052b2e7a3db369f4f0711655eccbcd0d259a7a7 is also conceptually part of this; it's what makes there be a template file for consul-template to use for this deployment.

ansible/playbooks/de-deploy-services.yaml
```diff
@@ -299,41 +316,6 @@
...
-  hosts: ui:&systems
-  become: true
-  gather_facts: false
-  tags:
-    - ui
-    - uncolored
-    - de-ui-nginx
-  roles:
-    - role: util-notify-chat
-      msg: "Deploying UI nginx"
```
ianmcorvidae: minor side note: I removed these notifications in moving them to de-deploy-service-nginx, since the other things there don't have notifications. It might make sense to add them back to the whole file at some point in the future

ash3rz:
Looks straightforward to me :shipit: And I think it's ok to drop the nginx msg for now if it's consistent.

ianmcorvidae:
Cool, thanks for the review!

PR #199, opened 2016-05-25
Use more up-to-date group accounting in submit file

ianmcorvidae:
http://research.cs.wisc.edu/htcondor/manual/v8.4/3_4User_Priorities.html#SECTION00447000000000000000
suggests that using accounting_group and accounting_group_user is the
modern way to do this, rather than using +AccountingGroup

I also moved it up in the file since we seem to roughly organize it with the submit-language stuff first and the + lines for custom attributes after.

### DIFF
--- a/golang/src/condor-launcher/condor.go
+++ b/golang/src/condor-launcher/condor.go
@@ -77,12 +77,13 @@ rank = mips
 arguments = --config config --job job
 output = script-output.log
 error = script-error.log
-log = condor.log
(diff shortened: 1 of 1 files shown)


ianmcorvidae:
n.b. this won't do much without https://github.com/cyverse/DE/pull/197

psarando:
Looks good to me.
May as well :shipit: now, then when this change is deployed in dev, the changes in PR #197 can be tested.

PR #200, opened 2016-05-25
Perms refactor

dennisroberts71:
This change adds permissions endpoints to the permissions service. The endpoints are intended to be as flexible as possible, so that clients do not need to register subjects or resources with the permissions service prior to granting permissions. If either the subject or the resource doesn't exist when a resource is being granted then an appropriate entry will be added to the database if possible. Permission deletions are more strict, but it's expected that deletions will only be performed after a related lookup has already occurred.

This change also introduces Grouper integration to the permissions service. At this time, the permissions service only uses Grouper to get group membership information. Querying Grouper web services would randomly take either around 0.1 seconds or between 4 and 5 seconds for unknown reasons. For this reason, we chose to access the Grouper database directly. A grouper client interface was defined so that this can easily be changed later if necessary. This interface is also used to provide a mock Grouper client for unit tests.


golang/src/permissions/restapi/impl/db/permissions.go
```diff
@@ -0,0 +1,352 @@
...
+	}
+
+	// Look up the permission.
+	permission, err := GetPermissionById(tx, permissionId)
+	if err != nil {
+		return nil, err
+	} else if permission == nil {
+		return nil, fmt.Errorf("unable to look up permission after upsert: %s", permissionId)
+	} else {
+		return permission, nil
```
johnworth: Looks like you can just end the function with `return permission, nil`, rather than enclosing it in the else.

johnworth: That's shouldn't prevent a merge, however.

golang/src/permissions/restapi/impl/db/permissions.go
```diff
@@ -0,0 +1,352 @@
...
+	// Check for duplicates. This shouldn't happen because of the uniqueness constraint.
+	if len(permissions) > 1 {
+		return nil, fmt.Errorf("multiple permissions found for subject/resource: %s/%s", subjectId, resourceId)
+	}
+
+	// Return the result.
+	if len(permissions) < 1 {
+		return nil, nil
+	} else {
+		return permissions[0], nil
```
johnworth: This return can be pulled up out of the else as well.

golang/src/permissions/restapi/impl/db/permissions.go
psarando: Yeah, when it comes to returns inside if-checks, my personal preference is to keep each if-return separate. The else's seem redundant to me.

golang/src/permissions/restapi/impl/db/subjects.go
```diff
@@ -138,3 +138,78 @@ func DeleteSubject(tx *sql.Tx, id models.InternalSubjectID) error {
...
+		return nil, fmt.Errorf(
+			"found multiple subjects with ID, %s, and type, %s", string(subjectId), string(subjectType),
+		)
+	}
+
+	// Return the result.
+	if len(subjects) < 1 {
+		return nil, nil
+	} else {
+		return subjects[0], nil
```
johnworth: Another return to pull up.

golang/src/permissions/restapi/impl/db/subjects.go
```diff
@@ -138,3 +138,78 @@ func DeleteSubject(tx *sql.Tx, id models.InternalSubjectID) error {
...
+		return nil, fmt.Errorf(
+			"found multiple subjects with ID, %s", string(subjectId),
+		)
+	}
+
+	// Return the result.
+	if len(subjects) < 1 {
+		return nil, nil
+	} else {
+		return subjects[0], nil
```
johnworth: Here's another return to pull up.

johnworth: Another return to pull up.

johnworth:
I don't know much about grouper and the permissions stuff, but I'm not seeing anything obvious that should prevent this from getting a :shipit:.

dennisroberts71:
Cool, thanks for the review.

PR #201, opened 2016-05-25
Add a time_limit_seconds column to the tools table

johnworth:
This pull requests adds a time_limit_seconds column to the tools table in the DE. It should be a "not null" column and defaults to 0.

A conversion named c270_2016052501.clj was added to update existing databases with the new column.

Kameleon's db version was updated to 2.7.0:20160525.01.

These was tested locally by doing the following:
* Mounting the new database.tar.gz and  facepalm uberjar files into a running discoenv/unittest-dedb:dev container and docker exec'ing the update-dev-database.sh script.

* Creating a new version of the discoenv/unittest-dedb:dev image with the changes loaded by default.

* Creating a new version of the discoenv/unittest-dedb:dev image with the changes loaded by default, and then running the the update-dev-database.sh script.

The database ships with 3 tools integrated, so the above testing also showed that existing tools got updated with the new default value.

### DIFF
--- a/databases/de-database-schema/src/main/conversions/c270_2016052501.clj
+++ b/databases/de-database-schema/src/main/conversions/c270_2016052501.clj
@@ -9,7 +9,7 @@
 (defn- add-time-limit-column
   []
   (println "\t* Adding time_limit_seconds column to the tools table.")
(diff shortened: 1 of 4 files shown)


dennisroberts71:
Looks good to me. :shipit:

PR #203, opened 2016-05-26
CORE-6613 fix issue when cancel is selected from filefolderselector d…

sri2k1us:
…ialog, it doesn't appear in the copy dialog.

### DIFF
--- a/ui/de-lib/src/main/java/org/iplantc/de/diskResource/client/views/widgets/MultiFileSelectorField.java
+++ b/ui/de-lib/src/main/java/org/iplantc/de/diskResource/client/views/widgets/MultiFileSelectorField.java
@@ -58,10 +58,13 @@ import com.sencha.gxt.dnd.core.client.DndDropEvent.DndDropHandler;
 import com.sencha.gxt.dnd.core.client.DropTarget;
 import com.sencha.gxt.dnd.core.client.StatusProxy;
 import com.sencha.gxt.widget.core.client.Composite;
+import com.sencha.gxt.widget.core.client.Dialog.PredefinedButton;
 import com.sencha.gxt.widget.core.client.button.TextButton;
 import com.sencha.gxt.widget.core.client.event.HideEvent;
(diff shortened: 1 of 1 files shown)


ianmcorvidae:
I don't know the significance of the change from `onHide` to `onDialogHide` but the rest of the changes (i.e. guarding the operation with a check that it's the appropriate button) look good to me.

ui/de-lib/src/main/java/org/iplantc/de/diskResource/client/views/widgets/MultiFileSelectorField.java
```diff
@@ -130,23 +133,25 @@ public FileSelectDialogHideHandler(final FileSelectDialog dlg,
...
+        	if(event.getHideButton().equals(PredefinedButton.OK)) {
+	            Set<File> files = diskResourceUtil.filterFiles(dlg.getDiskResources());
+	            if (files.isEmpty()) {
+	                return;
+	            }
+	            store.addAll(files);
+	            if (userSettings.isRememberLastPath() && store.size() > 0) {
+	                userSettings.setLastPath(diskResourceUtil.parseParent(store.get(0).getPath()));
+	                eventBus.fireEvent(new LastSelectedPathChangedEvent(true));
+	            }
```
ianmcorvidae: Though, hm. Should the "remember last path" functionality still work even when the dialog is cancelled? With this change, it looks like the path will only be remembered if "OK" is clicked, which may or may not be the intent.

sri2k1us: Good catch...I will add logic to save last path even when cancel is clicked.

sri2k1us: I just checked the behavior in other places, it does look like we only save last path only when user hits 'OK' button. So I will stick to that.

PR #204, opened 2016-05-26
time_limit_seconds and restricted fields are included in the job submission JSON

johnworth:
The apps service has been modified to include the time_limit_seconds and restricted fields from the tools table in the JSON sent to the jex-adapter service.

I ran the apps tests against this and there weren't any errors.


dennisroberts71:
This looks good to me. At one time I thought there was a schema check, so I was a little concerned about that failing when a job submission was done, but I didn't see one when I did a code search. This seems like it should be good to go. :shipit:

PR #207, opened 2016-06-02
Core 7688

sri2k1us:
This is in addition to the changes made by @ianmcorvidae in PR #206 . I have added logic to capture user unauthorized exception. This in-turn will cause the UI to clean-up itself (close open websockets) and then re-direct user to logged-out page.


dennisroberts71:
Looks good to me. :shipit:

PR #208, opened 2016-06-02
Core 7664

dennisroberts71:
This change was requested by the ODUM Institute, which doesn't have control over the LDAP instance that it uses and can't add the `de_grouper` user. The primary thing that I'd like to verify with this PR is that I've caught all of the cases where the `de_grouper` user was referenced directly in code.


johnworth:
:shipit:

de_grouper is still referenced in docker/test-ldap/test-de.ldif, docker/unittest-dedb/setup-grouper-database.sh, and as the default user in sharkbait, but none of that will affect any functionality. Just noting it for completeness.

PR #211, opened 2016-06-03
Perms refactor

dennisroberts71:
This change adds a client library for the permissions service. The testing for this is fairly thorough in that it attempts to verify that client correctly constructs all URIs and returns results in the appropriate format. An exhaustive test of all permissions service features is not done here. That level of detail is left to the permissions service integration tests.

Here's an integration test running against an instance of the permissions service running on my workstation:

```
$ RUN_INTEGRATION_TESTS=1 PERMISSIONS_BASE_URI=http://localhost:60000/ lein test

lein test permissions-client.core-test

lein test permissions-client.integration-test

Ran 40 tests containing 77 assertions.
0 failures, 0 errors.
```

*Warning:* do not attempt to run these integration tests against an existing permissions service instance. Doing so will erase all permissions information.

The unit tests can be run independently by not setting the `RUN_INTEGRATION_TESTS` environment variable:

```
$ lein test

lein test permissions-client.core-test

lein test permissions-client.integration-test

Ran 20 tests containing 23 assertions.
0 failures, 0 errors.
```

### DIFF
--- a/libs/permissions-client/src/permissions_client/core.clj
+++ b/libs/permissions-client/src/permissions_client/core.clj
@@ -1,0 +1,200 @@
+(ns permissions-client.core
+  (:require [cemerick.url :as curl]
+            [clj-http.client :as http]))
+
+(defprotocol Client
+  "A client library for the Permissions API."
+  (get-status [_]
+    "Retrieves information about the status of the permissions service.")
+
+  (list-subjects [_]
+    "Lists all subjects defined in the permissions service.")
+
+  (add-subject [_ external-id subject-type]
+    "Registers a subject in the permissions service. The external-id field is the subject ID known to the
+     client. For clients that use Grouper, this subject ID should the same as the one used by Grouper. The
+     subject ID must be unique within the permissions database. The subject-type field can be either 'user'
+     or 'group'.")
+
+  (delete-subject [_ id]
+    "Removes the subject with the given internal ID from the permissions database.")
+
+  (update-subject [_ id external-id subject-type]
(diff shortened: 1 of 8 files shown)


libs/permissions-client/test/permissions_client/integration_test.clj
```diff
@@ -0,0 +1,267 @@
...
+  (add-test-subjects)
+  (add-test-resources)
+  (add-test-permissions)
+  (f)
+  (remove-test-subjects)
+  (remove-test-resources)
+  (remove-test-resource-types)
+  (remove-test-permissions))
+
+(use-fixtures :each run-integration-tests with-base-uri with-test-data)
```
psarando: Nice! 👍
I like that we can run integration tests like this with clojure, rather than depending on external scripts.

libs/permissions-client/src/permissions_client/core.clj
```diff
@@ -0,0 +1,202 @@
...
+     set to 'true' and the subject happens to be a user then the most privileged permissions available to the
+     user or any group that the user belongs to (as determined by Grouper) will be listed. If the 'lookup?'
+     flag is set to 'false' or the subject is a group then only permissions that were granted directly to the
+     subject will be listed."))
+
+(defn- build-url [base-url & path-elements]
+  (str (apply curl/url base-url path-elements)))
+
+(deftype PermissionsClient [base-url]
+  Client
```
psarando: Very clean way of allowing the `base-uri` to be configurable outside of this library 👍

psarando:
Looks good! :shipit:

PR #212, opened 2016-06-07
time_limit_seconds and restricted fields are available through tools …

johnworth:
…API.

The apps services's tools API has been updated to allow the GET tools/:tool-id endpoint to return the time_limit_seconds and restricted fields. The PATCH /admin/tools/:tool-id endpoint has been updated to enable updating the same fields.

This was tested with the new integration tests added in the test/apps/tools_test.clj file and through the swagger interface that I ran after building the apps service locally and running it with Docker against our integration test environment.


psarando:

PR #213, opened 2016-06-08
CORE-7688 add session time-out redirection logic to belphegor

sri2k1us:
This is similar to the change done for DE UI. In addition spring config has to be modified little bit so that when session times out, requests hit the DeCasAuthenticationEntryPoint and gets 401 back.


dennisroberts71:
Looks good to me. :shipit:

PR #214, opened 2016-06-09
CORE-7681 Update metadata csv-parser to no longer add iRODS AVUs.

psarando:
This change also migrates the csv-parser logic into a new `POST /data/:data-id/metadata/csv-parser` endpoint in the data-info service, which no longer requires the `template-id` or `force` parameters.

The `force` parameter is no longer required since AVUs with duplicate attributes (but different values) are now allowed by the metadata service.


ianmcorvidae:
Looks good to me, pretty straightforward stuff. :shipit:

PR #215, opened 2016-06-09
CORE-7653 Script Tag Failure error

ash3rz:
This PR is to close the issue where users are getting Script Tag Failure errors in random places around the DE.  This error occurs when the user is navigating somewhere within the DE after their session has expired and then a GWT code split is requested and fails.  The solution here is to make sure that the code split requests first get returned with a status of 401, then the UI has to handle the redirect to the landing page properly.

This PR also pulls in some required changes to handling StatusCode 0 exceptions from dev that are also needed for this fix.

### DIFF
--- a/ui/de-lib/src/main/java/org/iplantc/de/shared/AsyncCallbackWrapper.java
+++ b/ui/de-lib/src/main/java/org/iplantc/de/shared/AsyncCallbackWrapper.java
@@ -5,7 +5,6 @@ import org.iplantc.de.shared.events.UserLoggedOutEvent;
 import org.iplantc.de.shared.exceptions.AuthenticationException;
 import org.iplantc.de.shared.exceptions.HttpRedirectException;
 
-import com.google.gwt.core.client.GWT;
 import com.google.gwt.user.client.Window;
 import com.google.gwt.user.client.rpc.AsyncCallback;
 import com.google.gwt.user.client.rpc.StatusCodeException;
@@ -23,7 +22,6 @@ import java.util.logging.Logger;
  * @param <T> the type of the result we're expecting to get from the server.
  */
 public class AsyncCallbackWrapper<T> implements AsyncCallback<T> {
-    private static final String LANDING_PAGE = "logged-out";
 
     Logger LOG = Logger.getLogger(AsyncCallbackWrapper.class.getName());
     /**
@@ -39,13 +37,6 @@ public class AsyncCallbackWrapper<T> implements AsyncCallback<T> {
 
     public AsyncCallbackWrapper(AsyncCallback<T> callback) {
         this.callback = callback;
-    }
-
-    /**
-     * Redirects the user to the DE landing page.
-     */
-    private void redirectToLandingPage() {
-        Window.Location.replace(GWT.getHostPageBaseURL() + LANDING_PAGE);
     }
 
     /**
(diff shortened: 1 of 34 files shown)


ui/de-lib/src/main/java/org/iplantc/de/shared/AsyncCallbackWrapper.java
```diff
@@ -50,24 +50,26 @@ private void redirectToLandingPage() {
...
      *
      * @param error the exception or error that indicates why the call failed.
      */
     @Override
     public void onFailure(Throwable error) {
-        if (error instanceof AuthenticationException) {
+       if (error instanceof AuthenticationException) {
+            EventBus.getInstance().fireEvent(new UserLoggedOutEvent());
             LOG.log(Level.SEVERE, "Auth error!!!!!", error);
-            redirectToLandingPage();
```
ianmcorvidae: if I'm reading this right, then `redirectToLandingPage()` is private to this file and not used after this PR, so its implementation should probably be removed (along with `LANDING_PAGE` which I believe only it currently cares about)

sri2k1us: 👍

ash3rz: Good catch, I'll remove it 👍

ui/de-lib/src/main/java/org/iplantc/de/shared/AsyncProviderWrapper.java
```diff
@@ -0,0 +1,55 @@
...
+
+    }
+
+    public void get(final AsyncCallback callback) {
+        provider.get(new AsyncCallback<T>() {
+            @Override
+            public void onFailure(Throwable caught) {
+                LOG.log(Level.SEVERE, caught.getMessage());
+                EventBus.getInstance().fireEvent(new UserLoggedOutEvent());
+            }
```
dennisroberts71: It would be nice if there were some way to examine the error to determine if the error was caused by an invalid session or by something else. In this case, anything that causes an error will result in the user being redirected to the landing page, whether or not the user's session is still active. If this is anything like the RPC callbacks, however, I suspect that it's difficult to find definitive information in the exception to determine the cause of the error, though.

dennisroberts71:
Assuming there's nothing we can do to distinguish errors caused by invalid sessions from other errors, this looks good to me. :shipit:

ianmcorvidae:
I can't completely claim to understand this, usual caveats about me and GWT, etc. but it looks pretty good to me in general. I like that the logout logic looks to have been centralized a bit more by way of the event bus (and this seems like an appropriate thing to use a global event bus for!). I worry that the very generic code that logs anyone out which fails in AsyncProviderWrapper will cause extra logouts, but it sounds like that's a necessary evil (basically, what @dennisroberts71 said while I was typing this comment!)

ui/de-lib/src/main/java/org/iplantc/de/shared/AsyncProviderWrapper.java
ash3rz: Agreed.  The error type that we're looking for that comes back is `AsyncFragmentLoader$HttpDownloadFailure` which is protected and not public, so I'm not sure of a way to capture that specifically.

PR #216, opened 2016-06-10
Migrate terrain ticket-related endpoints (CORE-6249, CORE-6225, CORE-6218)

ianmcorvidae:
This adds three new endpoints to data-info:

* POST /tickets (creates tickets)
* POST /ticket-lister (lists tickets for a set of paths in bulk)
* POST /ticket-deleter (deletes tickets for a set of ticket IDs in bulk)

These are pretty straightforward migrations of the terrain endpoints, which now use these endpoints.

In addition, requisite dependencies and configuration properties have been moved out of terrain and into data-info.

Any comments welcome, even improvements to code which I didn't originally write, stylistic tweaks, etc.


services/data-info/src/data_info/routes/tickets.clj
```diff
@@ -0,0 +1,45 @@
...
+        [data-info.routes.domain.common]
+        [data-info.routes.domain.tickets])
+  (:require [data-info.services.tickets :as tickets]
+            [data-info.util.service :as svc]
+            [data-info.util.schema :as s]))
+
+
+(defroutes* ticket-routes
+
+  (context* "/tickets" []
```
ianmcorvidae: (I added this as a context because it might be useful/desirable to add operations which deal with individual tickets -- I could see a use in a `GET /tickets/:ticket-id` and `DELETE `/tickets/:ticket-id` at least. Those aren't in here, though, because I really just wanted to move the terrain endpoints for now and that's all bulk operations.

dennisroberts71:
I didn't notice any problems. This one gets a :shipit: from me.

ianmcorvidae:
Thanks for the review!

ansible/roles/util-cfg-service/templates/data-info.properties.j2
```diff
@@ -7,6 +7,8 @@ data-info.kifshare-download-template = {{ kifshare.de_url }}
 data-info.max-paths-in-request       = {{ fs_max_paths_in_request }}
 data-info.metadata.base-url          = http://{{ metadata.host }}:{{ metadata.port }}
 data-info.tree-urls.base-url         = http://{{ tree_urls.host }}:{{ tree_urls.port }}
+data-info.kifshare-download-template = {{ kifshare.de_url }}
```
psarando: Looks like this config was already in this properties file (it's duplicated above `data-info.max-paths-in-request`).

ianmcorvidae: Oh, weird. I didn't even think to check since it's unused. I'll remove the extra.

ianmcorvidae: Er, since it was unused before this PR, rather.

PR #217, opened 2016-06-10
CORE-7708 Remove dup attr checks from POST /data/{data-id}/metadata/copy

psarando:
Remove duplicate attribute checks and the corresponding `force` param
from the POST /data/{data-id}/metadata/copy endpoint.
Move path-count validation from the terrain endpoint and into data-info.


dennisroberts71:
Looks good to me! :shipit:

PR #218, opened 2016-06-11
Perms refactor

dennisroberts71:
This change adds several new features to the permissions service and modifies the apps service to use the new permissions service. I haven't completely finished testing the changes to the apps service yet. I'll be adding some new integration tests next week. I wanted to get some eyes on this as soon as I could, though.

The most complicated (and least well tested) part of this PR is the `apps` service, so I'd like a slightly closer evaluation of these changes if possible. I'm going to modifying the existing integration tests to work with the new permissions service next week.


services/apps/test.sh
```diff
@@ -1,51 +0,0 @@
...
-# Wait for the DE database container to start up.
-sleep 10
-
-# Run the tests.
-docker run --rm -v $(pwd):/build -w /build --link $DBCONTAINER:postgres discoenv/buildenv lein $CMD \
-    || error_exit 'there were unit test failures'
-
-# Display a success message.
-echo 1>&2
-echo "TEST SUCCEEDED" 1>&2
```
dennisroberts71: I'm planning to update the Jenkins job before merging this PR.

services/apps/test/apps/containers_test.clj
```diff
@@ -1,26 +1,89 @@
...
+              tool-map           tool
+              settings-map       settings
+              devices-map        device
+              volume-map         volume
+              volumes-from-map   volumes-from]
+      (f)
+      (remove-test-data))))
+
+(use-fixtures :each with-test-db run-integration-tests with-test-data)
+
```
dennisroberts71: For these changes, I had to prevent any database operations from occurring before the `with-test-db` test fixture. The easiest way to do that was to create another test fixture that adds the test data.

services/apps/test/apps/containers_test.clj
```diff
@@ -1,26 +1,89 @@
...
+(defn- with-test-data [f]
+  (let [[image-info data-container tool settings device volume volumes-from] (add-test-data)]
+    (binding [image-info-map     image-info
+              data-container-map data-container
+              tool-map           tool
+              settings-map       settings
+              devices-map        device
+              volume-map         volume
+              volumes-from-map   volumes-from]
```
dennisroberts71: I tried to call the individual functions called by `add-test-data` directly in the `binding` form, but the binding didn't happen soon enough to be referenced directly. I created the `add-test-data` function because `with-test-data` was becoming a little cluttered.

services/apps/test/apps/persistence/categories_test.clj
```diff
@@ -1,7 +1,14 @@
...
         [apps.persistence.categories]
-        [apps.test-db-setup]))
+        [apps.test-fixtures :only [run-integration-tests with-test-db]]
+        [korma.core]))
+
+(defn clean-up-hierarchy-versions [f]
+  (f)
+  (delete :app_hierarchy_version))
+
+(use-fixtures :each with-test-db run-integration-tests clean-up-hierarchy-versions)
```
dennisroberts71: This change allows the tests to be repeatable.

services/apps/test/apps/containers_test.clj
```diff
@@ -1,26 +1,89 @@
...
+              (where {:container_images_id (:id image-info-map)})))
+
+(defn- remove-test-data []
+  (delete container-volumes-from (where {:id (:id volumes-from-map)}))
+  (delete container-volumes (where {:id (:id volume-map)}))
+  (delete container-devices (where {:id (:id devices-map)}))
+  (delete container-settings (where {:id (:id settings-map)}))
+  (delete data-containers (where {:id (:id data-container-map)}))
+  (remove-container-image-references)
+  (delete container-images (where {:id (:id image-info-map)})))
```
dennisroberts71: This change allows the tests to be easily repeatable.

services/apps/test/apps/test_fixtures.clj
```diff
@@ -0,0 +1,45 @@
...
+(def default-iplant-groups-uri "http://iplant-groups:60000")
+(def default-metadata-uri "http://metadata:60000")
+(def default-permissions-uri "http://permissions:60000")
+
+(def ^:dynamic jex-uri nil)
+(def ^:dynamic data-info-uri nil)
+(def ^:dynamic notification-agent-uri nil)
+(def ^:dynamic iplant-groups-uri nil)
+(def ^:dynamic metadata-uri nil)
+(def ^:dynamic permissions-uri nil)
```
dennisroberts71: I may end up removing these. It seemed as though it would be useful when I added them, but loading the settings from a configuration file is both necessary and easier in most cases.

golang/src/permissions/restapi/impl/db/resources.go
```diff
@@ -127,35 +147,41 @@ func UpdateResource(tx *sql.Tx, id *string, name *string) (*models.ResourceOut,
...
 }
 
-func ListResources(tx *sql.Tx) ([]*models.ResourceOut, error) {
+func ListResources(tx *sql.Tx, resourceTypeName, resourceName *string) ([]*models.ResourceOut, error) {
 
 	// Query the database.
-	query := `SELECT r.id, r.name, t.name AS resource_type
+	var rows *sql.Rows
+	var err error
+	if resourceTypeName != nil && resourceName != nil {
```
ianmcorvidae: This is a longer-term comment (i.e. ignore me if you'd like!), but I'm mildly uncomfortable with the repetitiveness of these chains of else-ifs. I wonder what you'd think of using something like https://github.com/Masterminds/squirrel (much as I wonder at my own sanity suggesting a programmatic query generation library!) to enable us to not repeat the same `SELECT r.id, r.name, t.name AS resource_type` that's in each branch of the if here.

dennisroberts71: That does sound useful, and query generation libraries have worked out fairly well for us in Clojure. I'm going to leave this as-is for the time being, but this does sound like a useful change for the future.

services/apps/src/apps/clients/iplant_groups.clj
```diff
@@ -86,227 +34,3 @@
...
-
-(defn unshare-analysis
-  "Unshares an analysis with a user."
-  [analysis-id subject-id]
-  (try+
-   (unshare-analysis* analysis-id subject-id)
-   (catch clj-http-error? {:keys [status body]}
-     (let [reason (get-error-reason body status)]
-       (log/error (str "uanble to unshare " analysis-id " with " subject-id ": " reason)))
-     "the analysis unsharing request failed")))
```
ianmcorvidae: Nothing like deleting hundreds of lines of code! :+1:

services/apps/test/apps/persistence/categories_test.clj
psarando: You could also use `:once` for the `run-integration-tests` and `with-test-db` fixtures:
```clojure
(use-fixtures :once run-integration-tests with-test-db)
```
That way you could ensure the fixtures run before every test, and only need to run once for the entire test run.
I did this for the metadata integration tests:
https://github.com/cyverse/DE/blob/dev/services/metadata/test/integration_tests.clj
https://github.com/cyverse/DE/blob/dev/services/metadata/test/integration_tests/metadata/services/avus_test.clj#L21

services/apps/src/apps/containers.clj
```diff
@@ -91,7 +91,9 @@
   (let [tag  (get-tag image-map)
         name (:name image-map)
         url  (:url image-map)]
-    (when-not (image? image-map)
+    (if (image? image-map)
+      (first (select container-images
+                     (where {:name name :tag tag})))
```
ianmcorvidae: I'm guessing this change is for getting some tests to work? It doesn't seem permissions-related but neither does it seem necessarily wrong. What's going on here?

services/apps/test/apps/persistence/categories_test.clj
dennisroberts71: That's true, as long as the cleanup is done after the tests:

```clojure
(use-fixtures :once run-integration-tests with-test-db clean-up-hierarchy-versions)
```

That's roughly equivalent to the code as it is right now, though, because there's only one test. The primary reason I made this change was so that I can run `lein test` repeatedly without having to reinitialize the database; it helps me to make my testing cycle faster.

services/apps/src/apps/containers.clj
dennisroberts71: Yeah, some of the tests were failing when the image map already existed because the function was returning `nil`.

services/apps/test/apps/containers_test.clj
psarando: I probably should have made my previous comment here instead:

You could use `:once` for the `run-integration-tests` and `with-test-db` fixtures here:
```clojure
(use-fixtures :once run-integration-tests with-test-db)
```
That way you could ensure the `with-test-db` test fixture runs first, once before every test, and then `with-test-data` can still be in an `:each` fixture that runs for each test.

I did this for the metadata integration tests:
https://github.com/cyverse/DE/blob/dev/services/metadata/test/integration_tests.clj
https://github.com/cyverse/DE/blob/dev/services/metadata/test/integration_tests/metadata/services/avus_test.clj#L21

dennisroberts71: That would work. I've been using `:each` because it helps to ensure that the database is cleaned up each time. In this case, that's probably not necessary, though.

services/apps/test/apps/service/apps/permissions_test.clj
```diff
@@ -0,0 +1,79 @@
...
+(defn- create-user [i]
+  (let [username (str "testde" i)]
+    {:user       username
+     :first-name username
+     :last-name  username
+     :email      (str username "@mail.org")}))
+
+(defn- create-user-map []
+  (->> (take 10 (iterate inc 1))
+       (mapv (comp (comp (juxt (comp keyword :user) identity) create-user)))
```
ianmcorvidae: This is minor and possibly more about me understanding it than anything else: is there a reason for the extra `comp` with only one argument? Expanded for posterity:

```clojure
(mapv
    (comp ; <- this one has only one argument, the second `comp`
        (comp 
            (juxt
                (comp keyword :user)
                identity)
            create-user)))
```

dennisroberts71: Nope, that was an editing error. I'll fix it.

ianmcorvidae:
Now that I've made it through, this looks good to me; I'd say :shipit:. Nothing functionality-breaking in the comments.

dennisroberts71:
Thanks for the reviews!

PR #219, opened 2016-06-13
Provide accurate byte ranges in anon-files

ianmcorvidae:
Previously, we were returning what RFC 2616 specifically mentions is
incorrect (a Content-Range header with no last-byte-pos). Though RFC
7233 which supersedes it does not include that line, we have a perfectly
accurate count of the last byte we are returning, and should return it
since the Content-Range header is intended to specify what was returned,
not what was requested. Therefore, this sets the filesize as the upper
end of the range for unbounded requests.

I don't actually know how important this is, but I do believe the changed value is more correct!


johnworth:
:shipit:

PR #221, opened 2016-06-14
Add a Python 2.7 tool and app to the database

johnworth:
We need to have a Python 2.7 tool and app that uses the time, memory, and cpu limits, so I've added one to be included with the DE when the database is init'ed. I've also included a conversion to add the same tool and app to existing databases.

The SQL for adding a new tool and app to the database is convoluted enough that I didn't want to add it to the existing data file, 21_internal_apps.sql.

This has been tested by init'ing a new database with the discoenv/unittest-dedb:dev image as well as by upgrading an existing database with the same image.

### DIFF
--- a/databases/de-database-schema/src/main/conversions/c270_2016061401.clj
+++ b/databases/de-database-schema/src/main/conversions/c270_2016061401.clj
@@ -1,0 +1,18 @@
+(ns facepalm.c270-2016061401
+  (:use [korma.core]
+        [kameleon.sql-reader :only [load-sql-file]]))
+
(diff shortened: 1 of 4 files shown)


dennisroberts71:
Looks good to me. :shipit:

psarando:
Why not keep `c270_*` conversions under a `databases/de-database-schema/src/main/conversions/v2.7.0/` subfolder, like the other conversions?

This `databases/de-database-schema/src/main/conversions` folder was really cluttered before, without the subfolders per version.

PR #222, opened 2016-06-14
Belphegor App Catalog enhancements

ash3rz:
This PR is a continuation of the Belphegor App Catalog/Ontology changes to address feedback from sci-analysts and the dev team.  This is mostly enhancements to CORE-7627 but also addresses a bug in CORE-7701.

Changes include:
* Rename the Ontology tab to App Catalog
* Auto-select the active ontology when Belphegor loads
* Auto-refresh the drop down list of ontologies after uploading a new ontology file
* Clear the ontology apps grid when uploading a new ontology file and/or selecting a different ontology version from the drop down
* Switch from clickable app name to an info button in the apps grid
* Fix a bug where only the last saved hierarchy IRI is displayed in the hierarchy tree
* Refresh the list of apps in a hierarchy after drag-and-drop
* When opening the Categorize dialog, auto-expand any hierarchies the selected app is tagged with


ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/ontologies/views/AppCategorizeViewImpl.java
```diff
@@ -42,12 +43,14 @@
...
 
     final private Widget widget;
+    private OntologyUtil ontologyUtil;
 
     @Inject
     public AppCategorizeViewImpl(final TreeStore<OntologyHierarchy> treeStore) {
         this.treeStore = treeStore;
         initCategoryTree();
         widget = uiBinder.createAndBindUi(this);
+        ontologyUtil = GWT.create(OntologyUtil.class);
```
sri2k1us: Why not just OntologyUtil.getInstance()? Just curious because I have not see this usage before.

ash3rz: Heh, not sure.  Must've been on auto-pilot, I'll change it.

sri2k1us:
Everything else looks good to me!

ash3rz:
Cool, thanks for the review!

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/ontologies/presenter/OntologiesPresenterImpl.java
```diff
@@ -331,10 +360,16 @@ void helperMap(List<OntologyHierarchy> children) {
...
+        boolean isLast = false;
+        for (String iri : iris) {
+            if (iris.indexOf(iri) == iris.size() - 1) {
+                isLast = true;
+            }
+            view.maskHierarchyTree();
             serviceFacade.saveOntologyHierarchy(event.getOntology().getVersion(),
-                                                iri, new SaveHierarchyAsyncCallback(iri));
+                                                iri, new SaveHierarchyAsyncCallback(iri, clearHierarchies, isLast));
+            clearHierarchies = false;
```
psarando: There is no guarantee that the first async call will return success first, or that the last async call will return success last.
So I don't think we can rely on these boolean flags for clearing and refreshing the hierarchy listing.
If we still want to support adding multiple hierarchies at a time, the easiest solution might be to just refresh the hierarchies from the service after each `onSuccess` returns.

ash3rz: That's a really good point that I hadn't thought of 👍 
I'll switch to your method of refreshing after each success.

PR #224, opened 2016-06-15
CORE-7712 Add 'avu' enum to metadata database target_enum type.

psarando:
Allows AVUs in the metadata db to target other AVUs.


ianmcorvidae:
Looks pretty straightforward. I'm assuming the weirdness with `@migrator/admin-db-spec` is about ensuring this runs outside a transaction. :shipit: from me!

psarando:
Yep, exactly.
Thanks for the review 👍

PR #225, opened 2016-06-15
Database Connection Retries

dennisroberts71:
`saved-searches` and `tree-urls` are often crashing upon startup when launched in our integration testing environment. This happens because these services start fast and often attempt to connect to Postgres before the database is ready. This change modifies both services so that they attempt to connect for a minute before giving up.


johnworth:
:shipit:

dennisroberts71:
Thanks for the review!

PR #191, opened 2016-05-12
Core 7622 Refactor metadata view so that users can view metadata using different templates.

sri2k1us:

PR #226, opened 2016-06-15
CORE-7622 refactor code. Extend Auv to use it with DiskresourceMetadata.

sri2k1us:
I was able to address most of the comments made by @psarando under #191 .


ui/de-lib/src/main/java/org/iplantc/de/diskResource/client/presenters/metadata/MetadataPresenterImpl.java
```diff
@@ -155,10 +154,11 @@ public void onFailure(Throwable caught) {
...
 
-        DiskResourceUserMetadata umd =
-                AutoBeanCodex.decode(autoBeanFactory, DiskResourceUserMetadata.class, "{}").as();
-        umd.setAvus(view.getUserMetadata());
-        drService.setDiskResourceMetaData(resource, umd, view.getAvus(), batchAvuCallback);
+        DiskResourceMetadataList umd =
+                AutoBeanCodex.decode(autoBeanFactory, DiskResourceMetadataList.class, "{}").as();
+        umd.setUserMetadata(view.getUserMetadata());
+        umd.setOtherMetadata(view.getAvus());
+        drService.setDiskResourceMetaData(resource, umd, batchAvuCallback);
```
psarando: Instead of generating a `DiskResourceMetadataList` with the `AutoBeanCodex` from a string, could this be created with `autoBeanFactory.metadataList()` instead?

ui/de-lib/src/main/java/org/iplantc/de/client/models/diskResources/DiskResourceMetadataList.java
```diff
@@ -13,14 +15,14 @@
...
-    void setUserMetadata(List<DiskResourceMetadata> userMetadata);
+    void setUserMetadata(List<Avu> userMetadata);
 
     @PropertyName("irods-avus")
-    List<DiskResourceMetadata> getOtherMetadata();
+    List<Avu> getOtherMetadata();
 
     @PropertyName("avus")
-    List<DiskResourceMetadata> getUserMetadata();
+    List<Avu> getUserMetadata();
```
psarando: This can extend `org.iplantc.de.client.models.avu.AvuList` rather than redefining `getUserMetadata` and `setUserMetadata`.
This is also how the backend service works: it merges the `irods-avus` key with the same metadata response that `AvuList` decodes.

psarando: So in the future, if the metadata service changes it's response format, then only `AvuList` would have to be updated to match, and not `DiskResourceMetadataList` as well.

ui/de-lib/src/main/java/org/iplantc/de/diskResource/client/presenters/metadata/MetadataPresenterImpl.java
sri2k1us: 👍

ui/de-lib/src/main/java/org/iplantc/de/client/models/diskResources/DiskResourceMetadataList.java
sri2k1us: 👍

PR #227, opened 2016-06-15
Core 7702 Frequent 'Unexpectedly disconnected' errors in logs relating to Websockets

sri2k1us:
Turns out that it was logged by our code at ERROR level although it not required. Changed that to INFO level. Refactored other logging messages.

### DIFF
--- a/ui/de-webapp/src/main/java/org/iplantc/de/server/websocket/MessageHandler.java
+++ b/ui/de-webapp/src/main/java/org/iplantc/de/server/websocket/MessageHandler.java
@@ -41,9 +41,9 @@ public abstract class MessageHandler extends WebSocketHandlerAdapter {
     @Override
     public void onOpen(WebSocket webSocket) throws IOException {
         notificationReceiver = new ReceiveNotificationsDirect();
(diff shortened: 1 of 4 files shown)


ui/de-webapp/src/main/java/org/iplantc/de/server/websocket/MessageHandler.java
```diff
@@ -65,7 +65,7 @@ public void onDisconnect(AtmosphereResourceEvent event) {
                     }
                 } catch (IOException e) {
                     e.printStackTrace();
-                    logger.error("^^^^^^^^^^^ exception aborting channel! ^^^^^^^^");
+                    logger.error("Exception aborting channel:" + e.getMessage());
```
psarando: Is it possible to pass the Throwable as another param to these `error` log messages, rather than concatenating the exception message and separately calling the `printStackTrace` method?

For example:
```java
logger.error("Exception aborting channel", e);
```

sri2k1us: 👍

PR #231, opened 2016-06-17
CORE-7715 Allow metadata AVU's to attach to other metadata AVU’s.

psarando:
Updated the metadata service GET/PUT/POST avus endpoints to support each AVU having its own `avus` list.

### DIFF
--- a/services/metadata/src/metadata/persistence/avu.clj
+++ b/services/metadata/src/metadata/persistence/avu.clj
@@ -5,7 +5,11 @@
   (:require [kameleon.db :as db]
             [korma.core :as sql]))
 
-(def ^:private data-types [(db/->enum-val "file") (db/->enum-val "folder")])
+(defn- target-where-clause
+  "Adds a where-clause to the given query for the given target."
+  [query target-type target-id]
+  (where query {:target_id   target-id
+                :target_type (db/->enum-val target-type)}))
 
 (defn filter-targets-by-attr-values
   "Finds the given targets that have the given attribute and any of the given values."
@@ -32,9 +36,7 @@
 (defn get-avus-for-target
   "Gets AVUs for the given target."
   [target-type target-id]
-  (select :avus
-          (where {:target_id   target-id
-                  :target_type (db/->enum-val target-type)})))
+  (select :avus (target-where-clause target-type target-id)))
 
 (defn get-avus-by-attr
   "Finds all existing AVUs by the given targets and the given set of attributes."
@@ -59,9 +61,12 @@
 (defn add-avus
   "Adds the given AVUs to the Metadata database."
   [user-id avus]
-  (let [fmt-avu (comp #(update % :target_type db/->enum-val)
-                      #(assoc % :created_by  user-id
-                                :modified_by user-id))]
+  (let [fmt-avu (fn [avu]
+                  (-> avu
+                      (select-keys [:id :attribute :value :unit :target_type :target_id])
+                      (update :target_type db/->enum-val)
+                      (assoc :created_by  user-id
+                             :modified_by user-id)))]
     (insert :avus (values (map fmt-avu avus)))))
 
 (defn update-avu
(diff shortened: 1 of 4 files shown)


services/metadata/src/metadata/persistence/avu.clj
```diff
@@ -98,25 +103,38 @@
...
-      (delete (where query {:id [not-in avu-ids]})))))
+  "Removes AVUs for the given target-id that are not in the given set of avu-ids-to-keep."
+  [target-type target-id avu-ids-to-keep]
+  (let [avus-to-remove (-> (select* :avus)
+                           (target-where-clause target-type target-id)
+                           (add-orphaned-ids-where-clause avu-ids-to-keep)
+                           select)]
+    ;; Remove orphaned sub-AVUs of any AVUs to be removed by this request.
+    (doseq [{:keys [id]} avus-to-remove]
+      (remove-orphaned-avus "avu" id nil))
```
psarando: This recursive call to `metadata.persistence.avus/remove-orphaned-avus` ensures sub-AVUs of any AVUs about to be removed are cleaned up as well.

services/metadata/src/metadata/services/avus.clj
```diff
@@ -25,20 +25,29 @@
...
 
 (defn- remove-orphaned-avus
   "Removes any AVU for the given target-type and target-id that does not have a matching ID in the given
    set of avus."
-  [target-type target-id avus]
-  (->> avus
+  [target-type target-id avus-to-keep]
+  ;; Cleanup orphaned sub-AVUs of any AVUs to be kept by this request
+  (doseq [{:keys [id avus]} avus-to-keep]
+    (remove-orphaned-avus "avu" id avus))
```
psarando: This recursive call to `metadata.services.avus/remove-orphaned-avus` cleans up sub-AVUs of any AVUs to be kept.

services/metadata/src/metadata/persistence/avu.clj
ianmcorvidae: I don't know what korma provides for us, but in general I'd implement this with a recursive CTE passed to the delete, letting postgres do the work. Roughly:

```sql
WITH RECURSIVE orphaned AS
  (SELECT id FROM avus WHERE target_id = $1 AND target_type = $2 AND <avus to keep etc>
     UNION
   SELECT id FROM avus WHERE target_type = "avu" AND target_id IN (SELECT id FROM orphaned))
DELETE FROM avus WHERE id IN (SELECT id FROM orphaned)
```

but, I don't know if this is worth the time implementing/testing

dennisroberts71:
I didn't notice any improvements aside from the recursive CTE suggestion that Ian mentioned. This gets a :shipit: from me.

services/metadata/src/metadata/routes/domain/avus.clj
```diff
@@ -24,7 +24,8 @@
    :created_by (describe String "The ID of the user who created the AVU")
    :modified_by (describe String "The ID of the user who last modified the AVU")
    :created_on (describe Long "The date the AVU was created in ms since the POSIX epoch")
-   :modified_on (describe Long "The date the AVU was last modified in ms since the POSIX epoch")})
+   :modified_on (describe Long "The date the AVU was last modified in ms since the POSIX epoch")
+   (s/optional-key :avus) (describe [(s/recursive #'Avu)] "AVUs attached to this AVU")})
```
ianmcorvidae: Huh! Does using `s/recursive` actually work in swagger?

services/metadata/src/metadata/services/avus.clj
ianmcorvidae: Could you clarify what you mean here? If they're being kept, shouldn't it always keep everything underneath them anyway, or am I misunderstanding something about how `avus-to-keep` is structured?

ianmcorvidae: (alternatively if you have a test of this just pointing me at it will do, I think I'm probably just not understanding)

dennisroberts71: I understood `avus-to-keep` to be the set of AVUs that are still applied to the AVU being modified and that any other AVUs that were formerly applied to the target should be removed.

services/metadata/src/metadata/routes/domain/avus.clj
psarando: Yes! It's also how we're documenting our [data-info NavigationResponse](https://github.com/cyverse/DE/blob/2.6.0-RELEASE/services/data-info/src/data_info/routes/domain/navigation.clj#L12-L19) and [app tree parameters](https://github.com/cyverse/DE/blob/2.6.0-RELEASE/services/apps/src/apps/routes/domain/app.clj#L45-L51).

ianmcorvidae: Nice. I'm always mildly amazed when stuff like this works right :laughing:

services/metadata/src/metadata/services/avus.clj
psarando: [This test might describe this situation](https://github.com/cyverse/DE/pull/231/files#diff-8d52c057dadd66993a99da85bf06ff26R148).

Dennis is correct.

These `remove-orphaned-avus` functions are only called in the `Set AVUs` endpoint. So it's possible that an AVU may have 2 existing sub-AVUs, but only 1 of them is included in the request. This means the client intends to delete one sub-AVU, and keep/modify the other.

ianmcorvidae:
:shipit: seems good to me

services/metadata/src/metadata/persistence/avu.clj
psarando: I don't think korma gives us any functions or macros for recursive queries, but there's still the option to use `exec-raw` with a parameterized SQL string.

psarando: Although I don't know if `exec-raw` supports list parameters (for the AVU id list), so I'll merge this PR and maybe experiment with `exec-raw` in another PR.

PR #232, opened 2016-06-17
CORE-7727, CORE-7728: metadata template blank/template CSV and guide CSV endpoints

ianmcorvidae:
This adds four endpoints between metadata and terrain:

# Metadata

* `/templates/:template-id/blank-csv`: blank CSV file suitable for filling out and using for bulk upload
* `/templates:template-id/guide-csv`: guide to fields in the template, in CSV form

# Terrain

* `/secured/filesystem/metadata/template/:template-id/blank-csv` passing to the metadata blank-csv endpoint
* `/secured/filesystem/metadata/template/:template-id/guide-csv` passing to the metadata guide-csv endpoint

The metadata endpoints accept an `attachment` query parameter, which if set to false (default true) will exclude the "attachment" bit from the content-disposition, which makes it easier to see the output in swagger docs.

There's a corresponding gh-pages-metadata-template-csv branch documenting the terrain changes.

### DIFF
--- a/services/metadata/src/metadata/routes/domain/template.clj
+++ b/services/metadata/src/metadata/routes/domain/template.clj
@@ -1,5 +1,5 @@
 (ns metadata.routes.domain.template
-  (:use [common-swagger-api.schema :only [describe]])
+  (:use [common-swagger-api.schema :only [describe StandardUserQueryParams]])
   (:require [schema.core :as s]
             [metadata.persistence.templates :as tp])
   (:import [java.util Date UUID]))
@@ -105,3 +105,8 @@
 
    :name
    (describe String "The metadata template name")})
+
+(s/defschema CSVDownloadQueryParams
+  (assoc StandardUserQueryParams
+    (s/optional-key :attachment)
+    (describe Boolean "Download file contents as attachment. Default true.")))
--- a/services/metadata/src/metadata/routes/templates.clj
+++ b/services/metadata/src/metadata/routes/templates.clj
@@ -5,6 +5,14 @@
         [ring.util.http-response :only [ok]])
   (:require [metadata.services.templates :as templates]))
 
+(defn- csv-download-resp
+  [attachment filename body]
+  (let [attachment? (or (nil? attachment) attachment)
+        disposition (str (if attachment? "attachment; " "") "filename=\"" filename "\"")]
+    (assoc (ok body)
+           :headers {"Content-Type" "text/csv; charset=utf-8"
+                     "Content-Disposition" disposition})))
+
 (defroutes* templates
   (context* "/templates" []
     :tags ["template-info"]
@@ -38,13 +46,13 @@
         :query [{:keys [attachment]} CSVDownloadQueryParams]
         :summary "Get a blank CSV template file for a metadata template."
         :description "This endpoint returns a CSV file suitable for a specific template, ready to be filled in with specific values. It's intended to be downloaded to be filled out by the user, then reuploaded for use with the bulk metadata endpoints."
-        (templates/csv-download-resp attachment "metadata.csv" (templates/view-template-csv template-id)))
+        (csv-download-resp attachment "metadata.csv" (templates/view-template-csv template-id)))
 
       (GET* "/guide-csv" []
         :query [{:keys [attachment]} CSVDownloadQueryParams]
         :summary "Get a CSV guide file for a metadata template."
         :description "This endpoint returns a CSV file providing a guide for a specific template. It's intended to be downloaded to be used as a reference while filling out a file from the blank-csv endpoint for the same template."
-        (templates/csv-download-resp attachment "guide.csv" (templates/view-template-guide template-id))))))
+        (csv-download-resp attachment "guide.csv" (templates/view-template-guide template-id))))))
 
 (defroutes* admin-templates
   (context* "/admin/templates" []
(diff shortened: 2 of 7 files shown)


services/metadata/src/metadata/routes/domain/template.clj
```diff
@@ -105,3 +105,8 @@
 
    :name
    (describe String "The metadata template name")})
+
+(s/defschema CSVDownloadQueryParams
+  (assoc StandardUserQueryParams
```
ianmcorvidae: This doesn't really require the user, so I wouldn't be opposed to removing it. But, it's nice for logs to know who requested something, so I kept it in for that reason.

services/metadata/src/metadata/routes/templates.clj
```diff
@@ -24,13 +24,27 @@
...
+        :return MetadataTemplate
+        :summary "View a Metadata Template"
+        :description "This endpoint returns the details of a single metadata template."
+        (ok (templates/view-template template-id)))
+
+      (GET* "/blank-csv" []
+        :query [{:keys [attachment]} CSVDownloadQueryParams]
+        :summary "Get a blank CSV template file for a metadata template."
+        :description "This endpoint returns a CSV file suitable for a specific template, ready to be filled in with specific values. It's intended to be downloaded to be filled out by the user, then reuploaded for use with the bulk metadata endpoints."
+        (templates/csv-download-resp attachment "metadata.csv" (templates/view-template-csv template-id)))
```
ianmcorvidae: this filename, and the other, are generic like this because we don't really sanitize the template names at all so I didn't want to use them, and the UUIDs are pretty much internal/meaningless to a user. if it should be a requirement that the suggested download filename include the name of the template, we could look into filename sanitization. The user can, of course, override this name anyway.

services/metadata/src/metadata/util/csv.clj
```diff
@@ -0,0 +1,8 @@
+(ns metadata.util.csv
+  (:require [clojure.data.csv :as csv]))
```
ianmcorvidae: I used clojure.data.csv rather than opencsv because it made more human-friendly CSV files -- quoting only when necessary, not for every field. Its interface is also nicer since it's clojure and not java.

services/metadata/src/metadata/routes/domain/template.clj
psarando: Yep, we've been keeping the user parameter on pretty much all of our endpoints, even if only for this reason.

services/metadata/src/metadata/routes/templates.clj
```diff
@@ -24,13 +24,27 @@
...
+        :query [params StandardUserQueryParams]
+        :return MetadataTemplate
+        :summary "View a Metadata Template"
+        :description "This endpoint returns the details of a single metadata template."
+        (ok (templates/view-template template-id)))
+
+      (GET* "/blank-csv" []
+        :query [{:keys [attachment]} CSVDownloadQueryParams]
+        :summary "Get a blank CSV template file for a metadata template."
+        :description "This endpoint returns a CSV file suitable for a specific template, ready to be filled in with specific values. It's intended to be downloaded to be filled out by the user, then reuploaded for use with the bulk metadata endpoints."
```
psarando: Minor recommendations:
These docs could use line-breaks to help readability in the code.
I think the 2nd `to be` could be replaced with a comma: `intended to be downloaded to be filled out by the user, ...` -> `intended to be downloaded, filled out by the user, ...`

services/metadata/src/metadata/routes/templates.clj
```diff
@@ -24,13 +24,27 @@
...
+        (templates/csv-download-resp attachment "metadata.csv" (templates/view-template-csv template-id)))
+
+      (GET* "/guide-csv" []
+        :query [{:keys [attachment]} CSVDownloadQueryParams]
+        :summary "Get a CSV guide file for a metadata template."
+        :description "This endpoint returns a CSV file providing a guide for a specific template. It's intended to be downloaded to be used as a reference while filling out a file from the blank-csv endpoint for the same template."
```
psarando: `to be` or ~~to be~~?

services/metadata/src/metadata/services/templates.clj
```diff
@@ -14,6 +17,30 @@
...
+
+(defn view-template-guide
+  [template-id]
+  (csv/csv-string
+    (->> (view-template template-id)
+      :attributes
+      (map (juxt :name :description :required :type #(if (:values %) (string/join ", " (map :value (:values %))) "")))
+      (cons ["attribute name", "attribute description", "required (If you cannot provide,  enter 'not collected', 'not applicable' or 'missing'.)", "value type definition", "enum value options (you must enter one of these values)"]))))
+
+(defn csv-download-resp
```
psarando: Minor suggestion: I wonder if this function should be defined up at the `metadata.routes.templates` level. Then the `ok` function and the details of the HTTP response could be kept there?

psarando:
My minor suggestions shouldn't prevent this PR from being merged.
Looks good to me otherwise, so :shipit: when you feel it's ready.

services/metadata/src/metadata/services/templates.clj
ianmcorvidae: That might make some sense. Putting it here was mostly an artifact of initially thinking the filenames would include info from the template (the name) which I decided against, so I think moving it back to where `ok` is already imported would make more sense.

PR #233, opened 2016-06-17
CORE-7717 Update ontology AVUs to add label AVU to hierarchy AVU

ash3rz:
These are the UI changes to complement CORE-7715 (PR #231)

### DIFF
--- a/ui/de-lib/src/main/java/org/iplantc/de/client/util/OntologyUtil.java
+++ b/ui/de-lib/src/main/java/org/iplantc/de/client/util/OntologyUtil.java
@@ -22,6 +22,8 @@ public class OntologyUtil {
 
     final String OPERATION_ATTR = "rdf:type";
     final String TOPIC_ATTR = "http://edamontology.org/has_topic";
+    final String LABEL_ATTR = "rdfs:label";
+    final String LABEL_UNIT = "value";
 
     final String UNCLASSIFIED_LABEL = "Unclassified";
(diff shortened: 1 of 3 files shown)


ui/de-lib/src/main/java/org/iplantc/de/client/util/OntologyUtil.java
```diff
@@ -69,7 +71,16 @@ public Avu convertHierarchyToAvu(OntologyHierarchy hierarchy) {
         Avu avu = avuFactory.getAvu().as();
         avu.setAttr(getAttr(hierarchy));
         avu.setValue(hierarchy.getIri());
-        avu.setUnit(hierarchy.getLabel());
+        avu.setUnit("");
```
ash3rz: It looked to me like Unit wasn't optional, so I'm passing an empty string here. Let me know if this needs to change.

psarando: That should be correct: the `Unit` should be required, even if it's an empty string.

psarando:
Looks good to me (pretty easy change too)! 👍 
#226 was just merged, which renamed `Avu.get/setAttr` to `get/setAttribute`, so you may want to rebase before merging this change, to ensure your changes are compatible.

ash3rz:
Good call - I'll rebase, fix any issues, and then close.  Thanks for the review!

PR #234, opened 2016-06-17
CORE-7642 and CORE-5906

ash3rz:
CORE-7642 - This fixes a bug where validations added to an App (in particular for Integer ranges, Decimal ranges, and Text regex/max char which all use liststores) were not saved in the App Editor window.  There were two issues. First issue was that sometimes the validators in App Editor were never actually saved.  The second issue was that when a user went to launch the app, there was logic preventing the saved validators from ever being applied.

CORE-5906 - The static IDs in this ticket have all been validated previously except for the checkboxes in the grid in the Collaborators dialog. My assumption was that setting the IDs after the dialog had been shown (show method) would be sufficient, but since the checkboxes are inside the grid you have to wait to set the IDs after the grid itself has been rendered (ViewReadyHandler).

### DIFF
--- a/ui/de-lib/src/main/java/org/iplantc/de/apps/integration/client/view/propertyEditors/widgets/ArgumentValidatorEditor.java
+++ b/ui/de-lib/src/main/java/org/iplantc/de/apps/integration/client/view/propertyEditors/widgets/ArgumentValidatorEditor.java
@@ -81,7 +81,7 @@ public class ArgumentValidatorEditor extends Composite implements ValueAwareEdit
             final AutoBean<ArgumentValidator> autoBean = AutoBeanUtils.getAutoBean(dlg.getArgumentValidator());
             autoBean.setTag(ArgumentValidator.TMP_ID_TAG, "tmpId-" + uniqueIdNum++);
             validators.getStore().add(autoBean.as());
-            ValueChangeEvent.fire(ArgumentValidatorEditor.this, validators.getStore().getAll());
+            ValueChangeEvent.fire(ArgumentValidatorEditor.this, getValidators());
         }
     }
 
@@ -317,7 +317,7 @@ public class ArgumentValidatorEditor extends Composite implements ValueAwareEdit
         for (ArgumentValidator av : selection) {
             validatorStore.remove(av);
         }
-        ValueChangeEvent.fire(this, selection);
+        ValueChangeEvent.fire(this, getValidators());
     }
 
     @UiHandler("edit")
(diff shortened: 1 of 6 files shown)


ui/de-lib/src/main/java/org/iplantc/de/apps/widgets/client/view/editors/arguments/converters/ArgumentEditorConverter.java
```diff
@@ -142,14 +142,11 @@ public void applyValidators(List<ArgumentValidator> validators, boolean isValida
...
-                    if ((field instanceof Field<?>) && (validator != null)) {
-                        @SuppressWarnings("unchecked")
-                        Validator<T> val = (Validator<T>)validator;
-                        ((Field<T>)field).addValidator(val);
-                    }
+                // Re-apply the actual validator
+                Object validator = autoBean.getTag(ArgumentValidator.VALIDATOR);
+                if ((field instanceof Field<?>) && (validator != null)) {
+                    @SuppressWarnings("unchecked") Validator<T> val = (Validator<T>)validator;
+                    ((Field<T>)field).addValidator(val);
```
psarando: Was this the "logic preventing the saved validators from ever being applied"?

ui/de-lib/src/main/java/org/iplantc/de/apps/integration/client/view/propertyEditors/widgets/ArgumentValidatorEditor.java
```diff
@@ -338,4 +338,10 @@ void onEditButtonSelected(@SuppressWarnings("unused") SelectEvent event) {
         dlg.show();
     }
 
+    List<ArgumentValidator> getValidators() {
+        List<ArgumentValidator> values = Lists.newArrayList();
+        values.addAll(validatorStore.getAll());
+        return values;
+    }
```
psarando: Was it necessary to copy the unmodifiable list of validators to the event because of this call above?
```java
model.setValidators(event.getValue());
```

ui/de-lib/src/main/java/org/iplantc/de/apps/widgets/client/view/editors/arguments/converters/ArgumentEditorConverter.java
ash3rz: Yes, just getting rid of the check for `isValidationDisabled` because it's always set to disabled prior to this method getting called and then re-enabled afterwards.  From what I could tell, we disabled validations prior to adding any validators otherwise as soon as the app was launched, every single field with validators would immediately have the red box around it instead of validating after input was entered.

ui/de-lib/src/main/java/org/iplantc/de/apps/integration/client/view/propertyEditors/widgets/ArgumentValidatorEditor.java
ash3rz: Yeah, it's specifically because I'm modifying the model.  If I leave it as an unmodifiable list, the next time the editor gets flushed, an error gets thrown because the ListStoreEditor will try to run
`model.clear();`
`model.addAll(store.getAll());`
and then error out.

psarando:
Looks good to me! :shipit:

ash3rz:

PR #236, opened 2016-06-20
CORE-7730 fix notification context click.

sri2k1us:
When user is logged in, after launching a job (or receive some form of notification), clicking on the notification did not open corresponding windows. This was result of websocket implementation code not setting up context after it receives a notification. Added logic to fix that issue.


ash3rz:
Looks straightforward to me :shipit:

PR #237, opened 2016-06-20
Core 7731Fix UI security bug where sensitive passwords and port numbers are exposed.

sri2k1us:


ui/de-lib/src/main/java/org/iplantc/de/shared/DEProperties.java
```diff
@@ -211,6 +213,27 @@ public static DEProperties getInstance() {
...
+        keys.add("org.iplantc.discoveryenvironment.muleServiceBaseUrl");
+        keys.add("org.iplantc.services.de-data-mgmt.base");
+        keys.add("org.iplantc.services.permIdRequests");
+        keys.add("org.iplantc.services.permIdRequests");
+        keys.add( "org.iplantc.services.admin.de-data-mgmt.base");
+        keys.add("org.iplantc.services.file-io.base.secured");
+        keys.add("org.iplantc.discoveryenvironment.notifications.poll-interval");
+        keys.add("org.iplantc.discoveryenvironment.contextMenu.enabled");
+        keys.add("org.iplantc.communitydata.path");
+        return keys;
```
ash3rz: Are these all the keys that the UI needs?  I thought the list was longer, like one for the Beta App Category id and one for trash I thought should be in here.

ash3rz:
If all the keys the UI needs are in the list, then this looks ready to me 🚢

ui/de-lib/src/main/java/org/iplantc/de/shared/DEProperties.java
sri2k1us: I picked the property list from whats listed in the DEProperties class. Hmm let me check belphegor side once.

sri2k1us: Seems like thats the list.

ash3rz: When I go to Belphegor with these changes, I get an exception that seems to stem from not finding `org.iplantc.discoveryenvironment.keepalive.interval`, so that key at least is missing.

Also, can we switch from adding the strings directly to using the variables they're assigned to, like `COMMUNITY_DATA_PATH` instead of `org.iplantc.communitydata.path`?

sri2k1us: I have submitted another PR that relates to deleting all the code that involves KeepAliveTimer. Add variables instead of strings sounds like a good plan

PR #238, opened 2016-06-20
CORE-7732 Re-enable red text for disabled apps in Belphegor

ash3rz:
PR #222 addressed a request to change Belphegor from clickable app names which would open up the App Details window to assigning that action to an Info button instead.  With that change, the red text that was applied to any disabled apps was lost.

This PR adds back the red text in the apps grid when an app is disabled in Belphegor.
(Most of the changes are file/method renames for clarity)


sri2k1us:
Looks good to me!

ash3rz:

PR #242, opened 2016-06-21
CORE-7691: added a script to register existing apps in the permission…

dennisroberts71:
…s database.

This utility can be used in the development and QA environments where we want to register existing apps, but not retain analyses or apps from before.


dennisroberts71:
@johnworth already reviewed the new utility, but I also made some Ansible changes, so if someone could look over those, it would be helpful.

ansible/roles/de-init-app-permissions/README.md
```diff
@@ -0,0 +1,34 @@
+Role Name
+=========
+
+A brief description of the role goes here.
```
psarando: Looks like this template file still needs editing...

psarando:
Other than my minor documentation comment, I didn't catch any issues with the new ansible playbooks/roles :shipit:

ansible/roles/de-init-app-permissions/README.md
dennisroberts71: Thanks. I'll fix that.

PR #243, opened 2016-06-23
CORE-7741 Allow null ontology IRI’s in metadata responses.

psarando:
Updated common-swagger-api.schema.ontologies/OntologyDetails to allow null `iri` fields.

This prevents `ERR_SCHEMA_VALIDATION` errors when uploading an ontology XML document that does not define the ontology's IRI.

Storing the ontology with a null IRI should not break any other functionality, since a unique ontology version is still created by the metadata service that's used in all other endpoints.


johnworth:
:shipit:

PR #244, opened 2016-06-23
CORE-7738 fix bug where template md in not imported to grid view when…

sri2k1us:
… the grid is empty.

### DIFF
--- a/ui/de-lib/src/main/java/org/iplantc/de/diskResource/client/presenters/metadata/MetadataPresenterImpl.java
+++ b/ui/de-lib/src/main/java/org/iplantc/de/diskResource/client/presenters/metadata/MetadataPresenterImpl.java
@@ -1,8 +1,5 @@
 package org.iplantc.de.diskResource.client.presenters.metadata;
 
-import java.util.ArrayList;
-import java.util.List;
-
 import org.iplantc.de.client.models.avu.Avu;
 import org.iplantc.de.client.models.diskResources.DiskResource;
 import org.iplantc.de.client.models.diskResources.DiskResourceAutoBeanFactory;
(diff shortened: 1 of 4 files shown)


ui/de-lib/src/main/java/org/iplantc/de/diskResource/client/presenters/metadata/MetadataPresenterImpl.java
```diff
@@ -217,6 +222,16 @@ public static Avu newMetadata(String attr, String value, String unit) {
         return avu;
     }
 
+    @Override
+    public Avu  setAvuModelKey(Avu avu) {
+        if (avu != null) {
+            final AutoBean<Avu> avuBean = AutoBeanUtils.getAutoBean(avu);
+            avuBean.setTag(AVU_BEAN_TAG_MODEL_KEY, String.valueOf(unique_avu_id++));
+            return avuBean.as();
```
ash3rz: This may not be correct, but I'm a little confused since this looks like the code that's in the ModelKeyProvider in the `createAdditionalListStore` method.  It seems like at least some of the time you're using `setAvuModelKey` right before adding to the liststore.  In that case, I wonder if it makes sense to modify `createAdditionalListStore` to return the key/tag if it exists, and then if it doesn't exist to create/set it and then return it.  This could just be a preference type of change though.

ui/de-lib/src/main/java/org/iplantc/de/diskResource/client/views/metadata/DiskResourceMetadataViewImpl.java
```diff
@@ -408,14 +394,22 @@ private Widget getCollapseBtn(ContentPanel panel) {
...
-                if (storemd.getAttribute().equals(md.getAttribute())) {
-                    storemd.setValue(md.getValue());
-                    userMdListStore.update(storemd);
-                }
+            Avu model = userMdListStore.findModel(md);
+            if (model != null) {
+                model.setValue(md.getValue());
+                userMdListStore.update(model);
+            } else {
+                itemsToAdd.add(presenter.setAvuModelKey(md));
```
ash3rz: If you decide to go the route I mentioned earlier, then a ModelKeyProvider for userMdListStore could also do the same get/set routine.

ui/de-lib/src/main/java/org/iplantc/de/diskResource/client/presenters/metadata/MetadataPresenterImpl.java
sri2k1us: The biggest reason I see for not doing that is the method "getKey" of ModelKeyProvider will causing "side-effect" that is not expected. The method is intended for just getting the keys. 
Also If the key does not exist, I would like the code throw a exception because, as far the logic goes, the every item in the store should have a tag. So if indeed there is an assertion error because of no tag, it will show up here.

ash3rz: I guess I liked that the liststore's ModelKeyProvider would handle anything related to keys so you don't have to worry about it anywhere else in the code.  But your explanation makes sense too. 👍

ash3rz:
I would definitely recommend tests for these if they don't already exist, otherwise 🚢

ui/de-lib/src/main/java/org/iplantc/de/diskResource/client/presenters/metadata/MetadataPresenterImpl.java
psarando: I'm on the fence about where to set the model keys.

It seems like it's actually logic that belongs to the view, since the presenter could potentially use a different view that doesn't care about list stores or model keys.

Just food for thought and I don't think this needs to change for this PR.

psarando:
Looks good to me :shipit:

sri2k1us:
Thanks @ash3rz @psarando. I promise to add more unit tests. But i think its getting late for hot fix.I am gonna merge now. But will start working test soon.

PR #245, opened 2016-06-23
CORE-7736 fix refresh button for HT analyses drill down.

sri2k1us:


ui/de-lib/src/main/java/org/iplantc/de/analysis/client/presenter/AnalysesPresenterImpl.java
```diff
@@ -286,33 +286,40 @@ public void go(final HasOneWidget container, final List<Analysis> selectedAnalys
...
+            config.setLimit(200);
+            config.setOffset(0);
+            loader.load(config);
+        } else {
+            idParentFilter.setField(AnalysisSearchField.PARENT_ID);
+            idParentFilter.setValue(view.getParentAnalysisId());
+            config.getFilters().add(idParentFilter);
+            config.setLimit(200);
+            config.setOffset(0);
+            loader.load(config);
```
psarando: It looks like most of these are duplicated in both the `if` and `else` branches.
I think most of these can be factored outside the if/else block: the `config.getFilters().add(filterCb)` and `idParentFilter.setValue(...)` calls can stay inside and I think everything else can come after.

dennisroberts71:
This looks good to me. :shipit:

PR #246, opened 2016-06-24
CORE-7700

dennisroberts71:
This change adds `app-registration` and `permissions` to the lists of images to be tagged automatically and to be included in the QA drop manifest file. Because I'm using `go-swagger` in the permissions service, I can't easily add variables to the `main` package. Because of this, I created a new `version` package and modified all existing Go services and utilities to use it.

### DIFF
--- a/docker/images.txt
+++ b/docker/images.txt
@@ -1,4 +1,5 @@
 anon-files
+app-registration
 clockwork
 condor-log-monitor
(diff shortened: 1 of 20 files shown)


johnworth:
:shipit:

dennisroberts71:
Cool, thanks for the review.

PR #247, opened 2016-06-24
Split job-status-to-apps-adapter and job-status-recorder into their own hostgroups

ianmcorvidae:
(in de-deploy-services.yaml)

This is to accommodate the new split in prod. There's also a branch in private ansible named after the new server.

This branch, and that, are against their respective beta branches since we'll want them for release on Tuesday.


johnworth:
I checked both this and the branch in the internal repo.

:shipit:

ianmcorvidae:
Cool, thanks for the review!

PR #249, opened 2016-06-24
CORE-7645 App catalog refresh for the DE

ash3rz:
This is sort of a first pass at getting the different ontology hierarchy tabs to show in the DE (not fully complete for the ticket, but merging it won't break anything).

There's still a lot to do to make this PR complete, but I wanted to get eyes on it to make sure how I'm handling the tabs is a decent route.

This PR should allow users to:
1. See their private workspace in a tab and the HPC apps in a separate tab within the Apps window
2. See the "active" ontology's root hierarchies in separate tabs (e.g. Topic and Operation)
3. View the filtered list of apps within each hierarchy (filtered meaning apps they have permissions to view)
4. Create workflows while viewing the same tabs they saw in the Apps window

Still to come:
1. Updating the Public Submission form
2. Being able to filter app search results (right now all apps matching the search term get returned regardless of which root hierarchy the user is currently viewing)
3. Restructuring Agave authentication redirects in the DE
4. Handling Beta apps


ui/de-lib/src/main/java/org/iplantc/de/apps/client/views/AppsViewImpl.java
```diff
@@ -25,39 +30,42 @@
     }
 
     @UiField(provided = true) final AppsToolbarView toolBar;
-    @UiField(provided = true) final AppCategoriesView appCategoriesView;
+    @UiField TabPanel categoryTabs;
```
ash3rz: This is really the biggest change - I went from `AppsView` having the `AppCategoriesView` directly to having just a TabPanel.  The TabPanel then gets passed to `AppCategoriesPresenter` and `OntologyHierarchiesPresenter` to add whatever tabs they need to add.

psarando:
I don't think it's necessary to include Agave authentication redirect changes or new app search functionality in this PR.
Those changes also require separate service changes that aren't available yet.

ui/de-lib/src/main/java/org/iplantc/de/apps/client/presenter/hierarchies/OntologyHierarchiesPresenterImpl.java
```diff
@@ -0,0 +1,333 @@
...
+    }
+
+    @Override
+    public void onOntologyHierarchySelectionChanged(OntologyHierarchySelectionChangedEvent event) {
+        fireEvent(event);
+    }
+
+    @Override
+    public void onAppRatingDeselected(AppRatingDeselected event) {
+        final App appToUnRate = event.getApp();
```
ash3rz: Just to note: a good chunk of the methods here were removed from `AppCategoriesPresenter` and placed here instead.

ash3rz:
Makes sense - I guess I wanted to mention those items since I think those changes are part of the requirements list we got.  But yeah, I figured this PR was probably big enough to start now 👍

ui/de-lib/src/main/java/org/iplantc/de/apps/client/models/AppCategoryStringValueProvider.java
```diff
@@ -11,7 +11,7 @@
 
     @Override
     public String getValue(AppCategory object) {
-        return object.getName() + " (" + object.getAppCount() + ")";
+        return object.getName();
```
sri2k1us: So we won't display app count under each term ?

ash3rz: I thought I remembered us having a conversation about this.  Since we can't have counts for how many apps are in the hierarchies, we would get rid of the app counts entirely... something along those lines.

psarando: Yeah, I think we are leaving the counts out for now, since these categories can now display a mix of Agave and DE apps, and we don't currently have a count for Agave apps under each category.

sri2k1us:
I didn't find any issues with code. Since this is a big PR,I will be interested in testing it after the merge.

ash3rz:
Makes sense - thanks for the review!

ui/de-lib/src/main/java/org/iplantc/de/apps/client/OntologyHierarchiesView.java
```diff
@@ -0,0 +1,41 @@
...
+import com.google.gwt.user.client.ui.IsWidget;
+
+import com.sencha.gxt.widget.core.client.TabPanel;
+import com.sencha.gxt.widget.core.client.tree.Tree;
+
+/**
+ * This view contains the hierarchical {@link AppCategory} tree, and fires off selection events.
+ *
+ * Created by jstroot on 3/5/15.
+ * @author jstroot
```
psarando: This this be created by @ash3rz ?

ash3rz: Heh, too much copy-pasta.  Yes, I'll fix.

ui/de-lib/src/main/java/org/iplantc/de/apps/client/presenter/categories/AppCategoriesPresenterImpl.java
```diff
@@ -84,201 +82,150 @@ public int compare(AppCategory group1, AppCategory group2) {
...
-        final App appToRate = event.getApp();
-        appUserService.rateApp(appToRate,
-                               event.getScore(),
-                               new RateAppCallback(appToRate,
-                                                   eventBus));
+    void filterCategories(List<AppCategory> result) {
+        workspaceCategories.clear();
+        hpcCategories.clear();
+        for (AppCategory category : result) {
+            if (!category.isPublic()) {
```
psarando: Rather than filtering public categories from the response, the endpoint supports a `public` flag. If set to `false`, it will only return the user's private workspace categories.

It currently includes the public HPC category in the `public=false` response as well. When that logical error gets fixed in the future, then maybe the UI could just make the HPC category itself, instead of fetching it from the server, since its ID is now stored in the UI configs.

ash3rz: That works much better - I always forget to check if there are already filters in place.  I'll update as you suggested.

ui/de-lib/src/main/java/org/iplantc/de/client/util/OntologyUtil.java
```diff
@@ -101,4 +125,74 @@ public AvuList convertHierarchiesToAvus(List<OntologyHierarchy> selectedHierarch
...
+     * In order to achieve this, any time a key is defined for a node, its children must get updated with
+     * the parents key in a separate parent tag.
+     * @param hierarchy
+     * @return
+     */
+    public String treeStoreModelKeyProvider(OntologyHierarchy hierarchy) {
+        String key = getHierarchyPathTag(hierarchy);
+        setChildrenParentTag(key, hierarchy);
+        return key;
+    }
```
psarando: I'm not sure these method names are very intuitive for what they accomplish.
Perhaps a method name like `getOrCreateHierarchyPathTag` would be better?
Also, could the `setChildrenParentTag` method only be called if the model's tag needed to be created, instead of every time the key is fetched?

ui/de-lib/src/main/java/org/iplantc/de/apps/client/presenter/hierarchies/OntologyHierarchiesPresenterImpl.java
```diff
@@ -0,0 +1,333 @@
...
+        }
+
+        @Override
+        public void onFailure(final Throwable caught) {
+            announcer.schedule(new ErrorAnnouncementConfig(appearance.fetchAppDetailsError(caught)));
+        }
+
+        @Override
+        public void onSuccess(final App result) {
+            serviceFacade.getAppAVUs(result, new AppAVUCallback(dlg, result));
```
psarando: Not for this PR, but it might be better if the details endpoint returns the app's metadata, rather than requiring the UI to chain these calls.

ui/de-lib/src/main/java/org/iplantc/de/client/util/OntologyUtil.java
ash3rz: Agreed, not very intuitive.  I'll refactor.

psarando:
I reviewed the core logic of this PR, and it looks good to me! 👍

PR #250, opened 2016-06-24
CORE-7686

dennisroberts71:
This change adds a new feature to the bootstrap endpoint that lists the redirect URIs for all external APIs for which the user hasn't authenticated. A new endpoint was also added to the apps service to retrieve the list of redirect URIs. In addition to this, two endpoints were added to allow users and administrators to get OAuth access tokens for troubleshooting purposes.

I'm not overly concerned about this change; the new and modified endpoints seem to work fine. It would be helpful to have someone review it for bugs, however.


services/apps/src/apps/service/oauth.clj
```diff
@@ -32,3 +36,47 @@
...
+                :query {:response_type "code"
+                        :client_id     (:client-key server-info)
+                        :redirect_uri  (:redirect-uri server-info)
+                        :state         (op/store-authorization-request username state-info)}))))
+
+(defn get-redirect-uris
+  "Retrieves the redirect URIs that can be used for a user to authenticate to a remote API."
+  [{:keys [username]}]
+  (let [build-auth-uri (fn [[_ server-info-fn]] (authorization-uri (server-info-fn) username ""))]
+    (remove-vals nil? (into {} (map (juxt key build-auth-uri) server-info-fn-for)))))
```
dennisroberts71: I just realized that there's a problem with using `server-info-fn-for` as-is for this endpoint. The problem is that we'll still try to create the redirect URI even if Agave is disabled, which can cause problems. I'll make a commit to ensure that this won't happen momentarily.

services/terrain/src/terrain/clients/apps/raw.clj
```diff
@@ -582,6 +582,27 @@
...
+  (client/get (apps-url "oauth" "token-info" api-name)
+              {:query-params     (secured-params)
+               :as               :stream
+               :follow-redirects false}))
+
+(defn get-oauth-redirect-uris
+  []
+  (client/get (apps-url "oauth" "redirect-uris")
+              {:query-params     (secured-params)
+               :as               :stream
```
psarando: Since this is always being decoded as JSON in terrain, and the response isn't forwarded to the client, we could change this to `:as :json` here and remove the `service/decode-json` call on these results.

psarando:
Other than my minor refactor comment, I didn't catch any issues :shipit:

services/terrain/src/terrain/clients/apps/raw.clj
dennisroberts71: In general, I would agree with this. My only reservation for making the suggested change is that it reduces the consistency in the behavior of the functions in `terrain.clients.apps.raw` and also reduces the consistency in the implementation of the functions in `terrain.clients.apps`.

dennisroberts71:
Cool, thanks for the review.

dennisroberts71:
I did another rebase before merging  the pull request for good measure.

PR #251, opened 2016-06-24
CORE-7745 Add deleted flag to metadata db ontologies table.

psarando:
A deleted flag won't be necessary for the `ontology_classes` or `ontology_hierarchies` tables.
A hierarchy will be deleted directly, which will delete the associated `ontology_classes` (and `ontology_hierarchies` by cascade).
They can easily be added back with the "save hierarchies" endpoint, which will add the exact same classes and hierarchies under a hierarchy root, parsed from the saved ontology in the `ontologies` table.


dennisroberts71:
:shipit:

PR #252, opened 2016-06-25
CORE-7746 Add metadata admin DELETE ontology and hierarchy endpoints.

psarando:
These changes allow an admin to mark an ontology as deleted (so it no longer displays in the ontology listing), and to delete all classes under an ontology hierarchy for a given `ontology-version`.
This will allow the admin to add an ontology hierarchy root, and then selectively remove sub-hierarchies, or the entire hierarchy from the root.

Classes and sub-hierarchies removed with the new endpoint can easily be re-added by re-saving the hierarchy root with the `PUT /admin/ontologies/{ontology-version}/{root-iri}` endpoint (it will only save classes and hierarchies that are not already saved under the given root).

## New metadata service endpoints
* `DELETE /admin/ontologies/{ontology-version}`
* `DELETE /admin/ontologies/{ontology-version}/{root-iri}`

## New apps service endpoint
* `DELETE /admin/ontologies/{ontology-version}`
    * Returns an error if the admin tries to mark the `active` ontology deleted, otherwise it acts as a passthrough to the metadata service endpoint.

## New terrain passthrough endpoints
* `DELETE /admin/ontologies/{ontology-version}`
    * Passthrough to apps
* `DELETE /admin/ontologies/{ontology-version}/{root-iri}`
    * Passthrough to metadata

### DIFF
--- a/libs/iplant-clojure-commons/src/clojure_commons/exception_util.clj
+++ b/libs/iplant-clojure-commons/src/clojure_commons/exception_util.clj
@@ -22,6 +22,11 @@
   [reason & {:as ex-info}]
   (throw+ (assoc ex-info :type ::cx/bad-request :error reason)))
 
+(defn illegal-argument
+  "Throws an error indicating that that a bad request argument was received."
+  [reason & {:as ex-info}]
+  (throw+ (assoc ex-info :type ::cx/illegal-argument :error reason)))
+
 (defn not-found
   "Throws an error indicating that a resource could not be found."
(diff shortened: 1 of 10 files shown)


dennisroberts71:
what you have so far looks good. :shipit: unless you want to have more changes reviewed under this PR.

psarando:
I'm still going to include the terrain and apps service changes in this PR, which should be pretty minor, so I'll probably just merge after those changes are done.

PR #255, opened 2016-06-28
Switch dockerops package to engine-api

johnworth:
This code previously used go-dockerclient. We're switching to allow us to use the same packages that are used by the Docker CLI.

Most of this commit consists of the addition of the vendored engine-api dependency. The useful stuff is in golang/src/dockerops/containers.go.


dennisroberts71:
Looks good to me. :shipit:

PR #256, opened 2016-06-30
CORE-7759: made some changes to help to resolve a performance problem…

dennisroberts71:
… in the GET /analyses endpoint in the apps service

The problem originally appeared to be related either to serialization or coercion. This was a red herring caused by the evaluation of a lazy sequence, however.

### DIFF
--- a/services/apps/src/apps/service/apps/job_listings.clj
+++ b/services/apps/src/apps/service/apps/job_listings.clj
@@ -95,7 +95,7 @@
   [apps-client job-id]
   (let [job-info   (jp/get-job-by-id job-id)
         app-tables (.loadAppTables apps-client [(:app-id job-info)])]
-    (format-job apps-client app-tables job-info)))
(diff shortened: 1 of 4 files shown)


psarando:
Looks good!
Fetching all the steps at once instead of a separate query per job should really speed things up. 👍 

services/apps/src/apps/service/apps/job_listings.clj
```diff
@@ -85,16 +85,17 @@
         search-params    (util/default-search-params params :startdate default-sort-dir)
         types            (.getJobTypes apps-client)
         jobs             (list-jobs* user search-params types analysis-ids)
+        rep-steps        (group-by (some-fn :parent-id :job-id) (jp/list-representative-job-steps (mapv :id jobs)))
```
psarando: When I saw `some-fn` I couldn't help thinking of this:
http://imgs.xkcd.com/comics/cirith_ungol.png

PR #257, opened 2016-06-30
CORE-7758: schema validation when emptying an already-empty trash

ianmcorvidae:
This changes the schema to use OptionalPaths rather than Paths, since
that makes the key optional but more importantly allows the list of
paths to be empty.

n.b. I don't have any huge preference that this go directly into master, though it is an annoying thing which was reported by a user, so it might be good to have this be a hotfix. Perhaps it should go into beta rather than dev, but not right into master? I'd love to hear thoughts!

### DIFF
--- a/services/data-info/src/data_info/routes/domain/trash.clj
+++ b/services/data-info/src/data_info/routes/domain/trash.clj
@@ -4,7 +4,7 @@
   (:require [schema.core :as s]))
 
 (s/defschema Trash
-  (assoc Paths
+  (assoc OptionalPaths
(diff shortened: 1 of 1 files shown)


dennisroberts71:
Looks good to me. :shipit:

psarando:
Looks good to me too :shipit: 
I guess this could have gone into `beta` first, then `master` could have been merged with `beta` after this was tested and deployed to prod (that would also save us a merge commit).

psarando:
Cool, looks like git was smart enough not to do an extra merge-commit after all 👍

PR #258, opened 2016-07-01
Rename compojure-api model namespaces

psarando:
This PR renames our `*.routes.domain` namespaces to `*.routes.schemas`.

Our first swaggerized project followed the compojure-api examples and put the swagger `schema` models under a `domain` namespace. This would make sense if these models were only used to define the valid "inputs" of our endpoints.

Since a lot of these models also define the valid "outputs" of our endpoints, I don't think it makes sense to keep them under `*.routes.domain` namespaces (i.e. these also define the "range" of some of our endpoints).

I think `*.routes.schemas` makes more sense because these namespaces mainly define models with `defschema` macros.


services/apps/src/apps/routes.clj
```diff
@@ -1,4 +1,4 @@
-(ns apps.routes.api
+(ns apps.routes
```
psarando: The `apps.routes.api` namespace was renamed to `apps.routes` for consistency with our other services.

PR #259, opened 2016-07-01
CORE-7747 Allow admins to delete ontologies and root hierarchies in Belphegor

ash3rz:
This is the UI portion of PR #252 where Belphegor is giving admins the ability to delete ontologies they've uploaded and also to delete root hierarchies.

(Since the toolbar is getting to be pretty long, I also removed the Refresh Ontologies button as that didn't seem all that necessary)


ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/ontologies/views/OntologiesViewImpl.java
```diff
@@ -276,18 +300,14 @@ public void onSelection(SelectionEvent<Ontology> event) {
         return ontologySimpleComboBox;
     }
 
-    void updateButtonStatus() {
+    public void updateButtonStatus() {
         publishButton.setEnabled(selectedOntology != null && selectedOntology != activeOntology);
         saveHierarchy.setEnabled(selectedOntology != null);
+        deleteButton.setEnabled(selectedOntology != null && selectedOntology != activeOntology);
+        deleteHierarchy.setEnabled(selectedOntology != null && treeStore.getRootItems().size() > 0);
```
ash3rz: Right now, you're able to Save Hierarchies and Delete Hierarchies on the active ontology... not sure if we should change that.

psarando: I think this should be OK. Especially since the admins might only ever add 1 ontology, so they should be able to modify the hierarchies for the sole, active ontology.

ash3rz: True.  I guess my worry was that we're not allowing the admins to delete the active ontology outright, but if you accidentally deleted all the hierarchies in the active ontology, that's sort of the same?

psarando: Currently, once the admin marks an ontology as deleted, there's no way for them to get it back.
If they delete a hierarchy, they can just re-save the hierarchy's root and get back any classes that were deleted.

ash3rz: Also true.  Yeah, I guess ultimately there's no harm in leaving the capability there. 👍

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/ontologies/presenter/OntologiesPresenterImpl.java
```diff
@@ -449,4 +462,40 @@ public OntologyHierarchy getSelectedHierarchy() {
...
+                announcer.schedule(new SuccessAnnouncementConfig(appearance.ontologyDeleted(event.getOntologyVersion())));
+                getOntologies(false);
+            }
+        });
+    }
+
+    @Override
+    public void onDeleteHierarchy(final DeleteHierarchyEvent event) {
+        List<OntologyHierarchy> deletedHierarchies = event.getDeletedHierarchies();
+        for (final OntologyHierarchy hierarchy : deletedHierarchies) {
```
psarando: I think it would be best to restrict the admin to single selection and only allow them to delete 1 hierarchy at a time. This could also allow us to have a hierarchy delete button somewhere in the main view, instead of requiring a separate deletion dialog.

When a hierarchy is deleted, all classes under that hierarchy are deleted and those classes may reside under other classes in the root hierarchy.

So in one case, this loop opens the possibility for the admin to select a class plus all sub-classes in its hierarchy, then a request is sent to delete that class, followed by requests to delete the sub-classes that were already deleted in the first request.

In a less obvious scenario, the admin could select a class, then in a different branch of the root hierarchy, select a class that happens to be a sub-class of the first selection, but maybe the admin didn't notice.
For example, the admin might accidentally select class B and D for deletion in the following hierarchy, but D would already be removed by deleting B:
```
        A
       / \
      B   D
     / \    \
    C   F    E
   /
  D
 /
E
```

In fact, while reviewing this PR, I discovered a bug in the service when attempting to delete a class that doesn't exist 😖 (but also 👍 for PR reviews).

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/ontologies/views/dialogs/DeleteHierarchiesViewImpl.java
```diff
@@ -0,0 +1,90 @@
...
+        List<ColumnConfig<OntologyHierarchy, ?>> columns = Lists.newArrayList();
+        columns.add(colCheckBox);
+        columns.add(label);
+        columns.add(iri);
+        return new ColumnModel<>(columns);
+    }
+
+    public void addHierarchyRoots(List<OntologyHierarchy> roots) {
+        if (roots != null && roots.size() > 0) {
+            listStore.addAll(roots);
```
psarando: I'm not sure if I'm reading this right, but it looks like this dialog is only giving the admin the option to delete hierarchies from the root.

The endpoint allows any class anywhere in the hierarchy to be deleted. So a root may be deleted, but any sub-class may also be deleted, including 'leaf' classes or classes anywhere in the middle of the hierarchy (which will also remove sub-classes of the deleted class).

This lets the admin "prune" a hierarchy by selectively removing just some sub-classes.

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/ontologies/presenter/OntologiesPresenterImpl.java
ash3rz: Double 👍 for PR reviews

Interesting - so just to make sure my understanding is correct, are we allowing admins to delete hierarchies that are not the root?  If so, I'll have some refactoring to do anyway.

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/ontologies/views/dialogs/DeleteHierarchiesViewImpl.java
ash3rz: Ha, yes, I just made a comment about this.  I, for some reason, thought the requirement was only for the root hierarchies to have the option to be deleted.  I'll refactor this.

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/ontologies/presenter/OntologiesPresenterImpl.java
psarando: Yes 👍 I just posted a comment explaining this elsewhere seconds before you asked this question :)

I think the documentation for the hierarchy delete endpoint needs to be clearer about this, so I'll also update the docs when I fix the bug.

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/ontologies/views/dialogs/DeleteHierarchiesViewImpl.java
psarando: No problem, as I posted in my other comment, I think the documentation for this hierarchy delete endpoint needs to be clearer.

ash3rz:
I just pushed a change that should allow admins to be able to delete any hierarchy.  On the flip side, it looks like if they accidentally delete a hierarchy and want to add it back, they would just have to go to the `Save Hierarchies` dialog and enter in that hierarchy's parent's IRI to get it back.  So I think that's reasonable.

psarando:
Yes, pretty much, except they shouldn't use the `Save Hierarchy` dialog unless they want to create a new "root" hierarchy (so saving is just for roots, but deleting can be for any sub-class).

Instead, they just need to re-save the root class of the hierarchy they want to re-add. See step 3 of this overview: https://github.com/cyverse/DE/blob/gh-pages-dev/api/endpoints/app-ontologies.md#categorizing-apps-with-ontology-hierarchies

ash3rz:
OK cool, I guess I hadn't looked at these docs in a while, but they're very clear. 👍

ui/de-lib/src/main/java/org/iplantc/de/theme/base/client/admin/ontologies/OntologyDisplayStrings.properties
```diff
@@ -23,4 +25,10 @@ categorizeApp = Select tags for {0}
...
 appClassifiedList = {0} tagged with {1,list}
-emptyDEOntologyLabel = <NONE>
\ No newline at end of file
+emptyDEOntologyLabel = <NONE>
+hierarchyDeleted = {0} successfully removed
+confirmDeleteOntology = Are you sure you want to delete ontology version {0}?
+ontologyDeleted = Deleted {0} successfully
+rootColumnLabel = Iri(s)
+hierarchyColumnLabel = Root(s)
+confirmDeleteHierarchy = Are you sure you want to delete the {0} hierarchy?
```
psarando: Would it be worth having a message with this confirmation letting the admin know that they can re-add the deleted sub-classes by re-saving the affected `root` hierarchy?
Although I'm not exactly sure how to word that more clearly.

psarando:
Other than my minor confirmation message comment, this looks good to me now :shipit:

ui/de-lib/src/main/java/org/iplantc/de/theme/base/client/admin/ontologies/OntologyDisplayStrings.properties
ash3rz: Yeah, I think it's worth it.  Even I was curious about how I would get the deleted hierarchy back.  I'll add a little blurb in the message before merging.

ash3rz:
Sweet - thanks for the review!  We caught some legit showstoppers 👍

PR #262, opened 2016-07-05
CORE-7760 Add ‘description’ column to metadata ‘templates’ table.

psarando:
Adds metadata db conversion 2.8.0:20160705.01 and updates the metadata db SQL init scripts.


dennisroberts71:
Looks good to me. :shipit:

PR #263, opened 2016-07-05
CORE-7759: updated one more call to format-job

dennisroberts71:


psarando:
Looks good to me :shipit:

PR #264, opened 2016-07-06
CORE-7767 Add warning for importing irods metadata into DE metadata

sri2k1us:


ui/de-lib/src/main/java/org/iplantc/de/diskResource/client/presenters/metadata/MetadataPresenterImpl.java
```diff
@@ -23,6 +23,7 @@
 import com.google.web.bindery.autobean.shared.AutoBeanUtils;
 
 import com.sencha.gxt.widget.core.client.Dialog.PredefinedButton;
+import com.sencha.gxt.widget.core.client.box.AlertMessageBox;
```
ash3rz: This looks like a stray import.

ash3rz:
(Github is taking a long time to verify whether this can be merged into dev or not, at least for me, so might want to check on that)

But otherwise, this looks straightforward 🚢

PR #266, opened 2016-07-06
Core 7764

dennisroberts71:
This change allows users from iPlant networks (or, really, any IP address range we want) to log into the DE when the DE is under maintenance. We have a new configuration setting that allows us to specify which IP address ranges to allow into the DE when it's under maintenance. The DE determines the original IP address of the request from the `X-Forwarded-For` header if it's present in the request or the remote IP address otherwise. Please see the Jira issue for more details.


sri2k1us:
Looks good to me!

dennisroberts71:
Cool, thanks for the review.

PR #267, opened 2016-07-07
CORE-7762 Update metadata templates in Belphegor with 'description' field and name limit

ash3rz:
This PR addresses a request for there to be a limit on how long the name of a metadata template can be (20 chars) and also adds an optional description field.  The metadata template names currently are too long because the description is also being included in the name.

I noticed some NPEs happening when clicking on the Edit and Delete buttons so I patched those up.  Then I noticed that the view was calling presenter methods directly, so I went ahead and refactored the view to fire events instead.

### DIFF
--- a/ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/metadata/view/TemplatesListingViewImpl.java
+++ b/ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/metadata/view/TemplatesListingViewImpl.java
@@ -110,8 +110,11 @@ public class TemplatesListingViewImpl extends Composite implements IsWidget, Tem
     @UiFactory
     ColumnModel<MetadataTemplateInfo> createColumnModel() {
         ColumnConfig<MetadataTemplateInfo, String> nameCol = new ColumnConfig<>(props.name(),
-                                                                                300,
+                                                                                150,
                                                                                 appearance.nameColumn());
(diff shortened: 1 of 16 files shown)


ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/metadata/view/TemplatesListingViewImpl.java
```diff
@@ -123,6 +150,7 @@ public String getKey(MetadataTemplateInfo item) {
                                                                                     appearance.deleted());
         List<ColumnConfig<MetadataTemplateInfo, ?>> columns = new ArrayList<ColumnConfig<MetadataTemplateInfo, ?>>();
         columns.add(nameCol);
+        columns.add(descriptionCol);
```
sri2k1us: Descriptions can be lengthy. I am not sure if displaying that as a column is good a idea.

ash3rz: Sure, I can remove the column.  It looks like if you do have a lengthy description it gets truncated in the column view after however many characters, at least.

Either way, this made me realize I should use `TextArea` in the edit window for description instead of `TextField` so I'll update that real quick.

sri2k1us:
Looks good to me!

ash3rz:
Cool - thanks for the review!

PR #268, opened 2016-07-09
CORE-7764: add local request checking to the DE controller

dennisroberts71:


ui/de-webapp/src/main/java/org/iplantc/de/server/controllers/DeController.java
```diff
@@ -46,12 +49,30 @@ public String redirectToDe(final HttpServletRequest request) throws MalformedURL
...
+
+    private boolean isLocalRequest(HttpServletRequest req) {
+        IpRanges ipRanges = new IpRanges(localIpRanges);
+        String forwardedFor = req.getHeader("X-Forwarded-For");
+        if (forwardedFor != null && forwardedFor.length() != 0) {
+            return ipRanges.matches(forwardedFor.split("\\s*,\\s*")[0]);
+        } else {
+            return ipRanges.matches(req.getRemoteAddr());
+        }
+    }
```
psarando: This code looks similar to the DeLandingPage#localRequest method.
Could this logic be refactored into a common place?

dennisroberts71: Yeah, I could probably create a static function somewhere that takes a request and a set of IP ranges and checks them.

ui/de-webapp/src/main/java/org/iplantc/de/server/controllers/DeController.java
```diff
@@ -46,12 +49,30 @@ public String redirectToDe(final HttpServletRequest request) throws MalformedURL
         return "redirect:de/";
     }
 
+    private DiscoveryEnvironmentMaintenance getDeMaintenance() {
+        return new DiscoveryEnvironmentMaintenance(maintenanceFile);
```
psarando: In the current code the `DiscoveryEnvironmentMaintenance` is constructed with `System.getProperty("user.dir") + "/" + maintenanceFile`.
Does just passing `maintenanceFile` work here?

ui/de-webapp/src/main/java/org/iplantc/de/server/controllers/DeController.java
dennisroberts71: Heh, or for that matter, I could probably make it an instance method of IpRanges.

psarando: Good point, I was wondering that myself :)

ui/de-webapp/src/main/java/org/iplantc/de/server/controllers/DeController.java
dennisroberts71: Yeah, `user.dir` resolves to the current working directory, so it's effectively the same as long as the path doesn't start with a path separator. When the path does start with a path separator, the current code will treat the path as absolute whereas the previous code will treat it as relative to the current working directory. The drop page code has always used just the maintenance file property name, so I went with that option.

psarando:
My comments were pretty minor, so otherwise this looks good to me :shipit:

PR #269, opened 2016-07-11
Core 7763 Display template description 

sri2k1us:


ui/de-lib/src/main/java/org/iplantc/de/diskResource/client/views/metadata/dialogs/DownloadTemplateCell.java
```diff
@@ -45,7 +45,7 @@ public void onBrowserEvent(Cell.Context context,
         }
 
         Element eventTarget = Element.as(event.getEventTarget());
-        if(eventTarget.getId().equals(value.getId())){
+        if(eventTarget.getClassName().contains("download")){
```
ash3rz: I noticed in the [Apps cells](https://github.com/cyverse/DE/blob/master/ui/de-lib/src/main/java/org/iplantc/de/apps/client/views/grid/cells/) we typically also include a check like `parent.isOrHasChild(eventTarget)`, I wonder if it might be a good idea to include that here as well on top of checking the ClassName?

ash3rz:
I had a minor comment, but otherwise looks straightforward :shipit:

sri2k1us:
Addressed all PR comments.

PR #272, opened 2016-07-12
CORE-7752: switched to Clojure 1.8

dennisroberts71:
This one should be short and sweet. The obvious change was to update the Clojure dependencies. The `clj-http` dependencies were updated because ancient versions relied on a version of `potemkin` that was incompatible with Clojure 1.8. The `elastisch` dependencies were updated because they relied on an ancient version of `clj-http`.


psarando:
I don't see any obvious issues with this PR.
Better :shipit: sooner rather than later, to catch any issues before the next release.

dennisroberts71:
Cool, thanks for the review.

PR #273, opened 2016-07-12
CORE-7773 Save AVUS instead of category IDs in apps publish endpoint.

psarando:
The `/apps/:id/publish` endpoint now accepts a list of AVUS instead of category IDs, which it will publish immediately to the metadata service.
This means apps published by users are categorized immediately into the ontology hierarchies the user chooses, rather than requiring admins to categorize the apps.

* This also makes the `suggested_groups` table obsolete, and it can be removed in a future db conversion.
* In a following PR, this endpoint will also tag the app with a "Beta" AVU, which will be the indicator to users that the app has not been curated by the admins yet.


dennisroberts71:
Looks good to me! :shipit:

psarando:
Thanks for the review 👍

PR #274, opened 2016-07-13
CORE-7645 Update Public Submission Form for Ontology Hierarchies

ash3rz:
These are the UI changes that go along with PR #273.  Basically, since we're moving away from app categories, when a user wants to publish an app, they now have the active ontology's hierarchies to choose from to categorize their app.  Those hierarchies then get turned into AVUs for metadata.

### DIFF
--- a/ui/de-lib/src/main/java/org/iplantc/de/apps/client/presenter/submit/SubmitAppForPublicPresenter.java
+++ b/ui/de-lib/src/main/java/org/iplantc/de/apps/client/presenter/submit/SubmitAppForPublicPresenter.java
@@ -2,28 +2,30 @@ package org.iplantc.de.apps.client.presenter.submit;
 
 import org.iplantc.de.apps.client.SubmitAppForPublicUseView;
 import org.iplantc.de.apps.client.events.AppPublishedEvent;
-import org.iplantc.de.apps.client.presenter.categories.proxy.PublicAppCategoryProxy;
 import org.iplantc.de.client.events.EventBus;
-import org.iplantc.de.shared.DEProperties;
 import org.iplantc.de.client.models.apps.App;
-import org.iplantc.de.client.models.apps.AppCategory;
 import org.iplantc.de.client.models.apps.AppRefLink;
+import org.iplantc.de.client.models.apps.PublishAppRequest;
+import org.iplantc.de.client.models.ontologies.OntologyHierarchy;
 import org.iplantc.de.client.services.AppUserServiceFacade;
+import org.iplantc.de.client.services.OntologyServiceFacade;
 import org.iplantc.de.client.util.JsonUtil;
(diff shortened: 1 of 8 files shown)


ui/de-lib/src/main/java/org/iplantc/de/apps/client/presenter/submit/SubmitAppForPublicPresenter.java
```diff
@@ -33,56 +35,70 @@
     interface SubmitAppPresenterBeanFactory extends AutoBeanFactory {
         AutoBean<AppRefLink> appRefLink();
     }
-    @Inject PublicAppCategoryProxy appGroupProxy;
+
+    private class HierarchiesCallback implements AsyncCallback<List<OntologyHierarchy>> {
```
psarando: The logic in these callback methods are pretty much duplicated in `OntologiesPresenterImpl` and `OntologyHierarchiesPresenterImpl`.
This logic should be refactored into a common place at some point (but not necessary for this PR).

ui/de-lib/src/main/java/org/iplantc/de/apps/client/views/submit/SubmitAppForPublicUseViewImpl.java
```diff
@@ -445,14 +417,14 @@ public static native String generateMarkDown(String name,
     ;
 
     @Override
-    public JSONObject toJson() {
-        JSONObject json = new JSONObject();
+    public PublishAppRequest getPublishAppRequest() {
+        PublishAppRequest appRequest = factory.publishAppRequest().as();
```
psarando: Excellent refactor 👍

ui/de-lib/src/main/java/org/iplantc/de/client/models/apps/PublishAppRequest.java
```diff
@@ -0,0 +1,42 @@
...
+
+import com.google.gwt.user.client.ui.HasName;
+import com.google.web.bindery.autobean.shared.AutoBean.PropertyName;
+
+import java.util.List;
+
+/**
+ * @author aramsey
+ */
+public interface PublishAppRequest extends HasId, HasName, HasDescription {
```
psarando: Note that the service endpoint's `PublishAppRequest` schema allows the `id`, `integration_date`, and `edited_date` fields, but these were only allowed as a convenience to the UI, which was sending an app object that included these fields to the service.

The endpoint doesn't check these fields (the ID is already included as part of the endpoint's URL), so they may be safely removed from this new `PublishAppRequest` model. In a future update, the endpoint may distinguish between app creator and app publisher, so we can update both this model and the endpoint at that time to remove these extraneous fields.

psarando:
I didn't catch any issues that should prevent this PR from being merged before the next QA drop.
We can :shipit: after #273 is merged.

PR #275, opened 2016-07-13
CORE-7755: download zip of blank+guide template CSVs

ianmcorvidae:
Plus, some refactoring. Mostly just making it so we don't hit the database multiple times for this endpoint which does several things with the same template data.

This uses a sanitize-filename library I found to generate the folder name inside the zip file from the template name. It seems to be pretty sane, though it does choose to replace slashes with dollar signs which could be mildly confusing. Still, it should be pretty readable in general. For making zip files, apparently `me.raynes.fs` has tools for this already! Woo!

### DIFF
--- a/services/metadata/src/metadata/routes/templates.clj
+++ b/services/metadata/src/metadata/routes/templates.clj
@@ -5,13 +5,17 @@
         [ring.util.http-response :only [ok]])
   (:require [metadata.services.templates :as templates]))
 
-(defn- csv-download-resp
-  [attachment filename body]
+(defn- filetype-download-resp
+  [filetype attachment filename body]
   (let [attachment? (or (nil? attachment) attachment)
         disposition (str (if attachment? "attachment; " "") "filename=\"" filename "\"")]
     (assoc (ok body)
-           :headers {"Content-Type" "text/csv; charset=utf-8"
+           :headers {"Content-Type" filetype
                      "Content-Disposition" disposition})))
+
+(defn- csv-download-resp
+  [attachment filename body]
+  (filetype-download-resp "text/csv; charset=utf-8" attachment filename body))
 
 (defroutes* templates
   (context* "/templates" []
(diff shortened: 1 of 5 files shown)


services/metadata/src/metadata/routes/templates.clj
```diff
@@ -57,7 +61,14 @@
...
                      filling out a file from the blank-csv endpoint for the same template."
-        (csv-download-resp attachment "guide.csv" (templates/view-template-guide template-id))))))
+        (csv-download-resp attachment "guide.csv" (templates/view-template-guide template-id)))
+
+      (GET* "/zip-csv" []
+        :query [{:keys [attachment]} CSVDownloadQueryParams]
+        :summary "Get a ZIP file containing both the guide and blank CSV for a template."
+        :description "This endpoint returns a zip file containing both the
+                     blank CSV and the guide CSV for the same template."
+        (filetype-download-resp "application/zip" attachment "template.zip" (templates/view-template-zip template-id))))))
```
ianmcorvidae: The changes in this file, aside from the obvious adding of a new endpoint, are just to make it so we can return the proper `application/zip` MIME type.

services/metadata/src/metadata/services/templates.clj
```diff
@@ -16,24 +18,39 @@
...
+    (->> template-data
       :attributes
       (map (juxt :name :description :required :type #(if (:values %) (string/join ", " (map :value (:values %))) "")))
       (cons ["attribute name", "attribute description",
              "required (If you cannot provide,  enter 'not collected', 'not applicable' or 'missing'.)",
              "value type definition", "enum value options (you must enter one of these values)"]))))
 
+(defn view-template-guide
+  [template-id]
+  (format-template-guide (view-template template-id)))
```
ianmcorvidae: Stuff above here is all refactoring to separate the formatting of the CSV files from template data from the fetching of the data, for the existing/prior endpoints.

services/metadata/src/metadata/services/templates.clj
```diff
@@ -16,24 +18,39 @@
...
+
+(defn view-template-zip
+  [template-id]
+  (let [template-data (view-template template-id)
+        template-name (sf/sanitize (:name template-data))]
+    (cmp/make-zip-stream [(str template-name "/blank.csv") (format-template-csv template-data)]
+                         [(str template-name "/guide.csv") (format-template-guide template-data)])))
```
ianmcorvidae: This function is what's doing all the work.

psarando:
Looks straight forward to me 👍
Good finds in these libraries too :)
:shipit:

ianmcorvidae:
Cool. Thanks for the review!

PR #276, opened 2016-07-14
CORE-7934 Update publish endpoint to include beta AVU in app metadata.

psarando:
Since the Beta category is going away with all the old app categories, when an app is made public from now on, it will be "tagged" with a "Beta" AVU, rather than added to the old Beta category.

These changes remove references in the apps service to the old apps.workspace.beta-app-category-id config and add new settings for the beta AVU’s attr, value, and the attr's label.
These new configs are used to automatically add the beta AVU to apps published with the `/apps/:id/publish` endpoint.


dennisroberts71:
Looks good to me! :shipit:

PR #277, opened 2016-07-14
core-7729 update end-point

sri2k1us:
This PR is just to update end-point to download the csv template and guide in a single zip file.


ianmcorvidae:
I'm not sure what status the rest of the code is in de-2 (I can't get this to work through the UI), but this is correct and I can at least validate the `/de/secured/mdTemplateDownload` currently downloads the blank csv so it would be changed to the zipfile. So as far as this PR, I'd say :shipit:

sri2k1us:
Selecting metadata -> Download template works for me in de-2. Clicking on the download icon in the select template dialog didn't work. But i double checked that both works in my branch!

PR #279, opened 2016-07-14
CORE-7677: upgrade jargon version and return to using AVU query code

ianmcorvidae:
Jargon 4.0.2.6 has been released, which includes my fixes for the AVU query code, so we can now return to the code I'd previously written for AVU lookups by specific attributes and values, which is more efficient.


psarando:
Looks good, I didn't catch any issues :shipit:

PR #280, opened 2016-07-14
CORE-7935 Update apps listing endpoints to include a beta flag.

psarando:
The `/apps/`, `/apps/categories/:id`, and `/apps/hierarchies/:iri/apps` endpoints now include a `beta` flag for each app, which is `true` for apps that have the `beta` metadata AVU attached.


ianmcorvidae:
Changes look good to me. I haven't worked with this code much, so the main thing I didn't check/might not be equipped to check efficiently is if everywhere that this logic should appear it does. I'm assuming you've checked that in producing the list in your description, though, so I'd say :shipit:

PR #281, opened 2016-07-16
CORE-7936 Add an apps-beta-tagger database conversion tool

psarando:
This PR adds the apps-beta-tagger database conversion tool, which attaches the "beta" metadata AVU to all apps found in the current "Beta" apps category.


ansible/roles/de-TMP-RUN-ONCE-app-beta-tagger/tasks/main.yml
```diff
@@ -0,0 +1,31 @@
+---
+- debug:
+    msg: "This role should only be RUN ONCE against the metadata database. It will fail if run more than once."
```
psarando: The tool could probably be improved to check for existing beta AVUs, and then not attempt to insert one if found. This would be similar to the logic in the metadata POST /avus endpoint.

ansible/roles/de-TMP-RUN-ONCE-app-beta-tagger/tasks/main.yml
```diff
@@ -0,0 +1,31 @@
...
+
+- name: update docker-py
+  pip: name="docker-py" state="latest"
+
+- name: pull the apps beta tagger docker image
+  docker_image: name="{{ app_beta_tagger_repo }}" tag="{{ docker_tag }}"
+
+- name: tag DE beta apps
+  shell: "docker run --rm
```
psarando: I resorted to a `shell` command here because the `docker_container` version of this command was not running the conversion for some reason, but still returning success to the task.

I was also not sure how to capture the `docker_container` command's output to debug the issue.

ansible/roles/de-TMP-RUN-ONCE-app-beta-tagger/tasks/main.yml
```diff
@@ -0,0 +1,31 @@
...
+    --name apps-beta-tagger
+    -v ~/.pgpass:/root/.pgpass:ro
```
psarando: It might be better to pass the db passwords to the tool on the command line, but I had already written the tool to use `kameleon.pgpass/get-password` by the time I got to this playbook and the tool's docker container.

psarando:
A Jenkins job has already been setup for this tool, so it should automatically trigger once this PR is merged into `dev`.

dennisroberts71:
This looks good to me. I'm merging it now since @psarando is out of the office.

PR #285, opened 2016-07-25
CORE-7261: update iplant-groups exception handling to match other services

ianmcorvidae:
This should just be consistent with other such exception handling updates.


cyverse-org:
LGTM
:shipit:

PR #286, opened 2016-07-25
Use the correct name in the context map for preventing temporary password creations

ianmcorvidae:
I didn't test my changes in #139 well enough, clearly, since it didn't actually work at all :confounded: 

I've also added some debug-level logging, which won't appear in most cases but could hopefully be useful for other debugging in the future.


tedgin:
It looks straight forward to me. :shipit:

ianmcorvidae:
Cool, thanks for the review!

PR #287, opened 2016-07-26
Use explicit column names in a notificationagent insert statement

ianmcorvidae:
This happened to be erroring in a QA environment, I think for unrelated reasons, but I think it's better to be explicit about the columns we're inserting into.

So, minor, but may as well be fixed. I'd like a double-check that I've put the right columns in the right order, though.


dennisroberts71:
Looks good to me. :shipit:

ianmcorvidae:
Thanks for the review!

PR #288, opened 2016-07-27
Core 8087 - Ansible Elasticsearch host/group refactoring

strootman:
The current host groups which call out the hosts which host elasticsearch clusters for our data-store- clusters are in the `[elasticsearch]` host groups. 
Currently, the DE only runs one cluster, but we cannot preclude that we might run multiple clusters in the future. 

Therefore, it is desirable to separate variables related to accessing the data-store- clusters from variables related to the installation of elasticsearch.

Variables related to elasticsearch installation are kept in the private ansible repo.

### DIFF
--- a/ansible/inventories/example.cfg
+++ b/ansible/inventories/example.cfg
@@ -170,10 +170,12 @@ amqp.example.com
 docker-registry.example.com
 
 ###############################################################################
-# elasticsearch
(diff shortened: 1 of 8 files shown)


dennisroberts71:
These changes look good to me. :shipit:

PR #289, opened 2016-07-27
CORE-7898 Make apps tagged as Beta appear as blue

ash3rz:
This applies to both Belphegor and the DE.  Belphegor just doesn't have the underlined hyperlink app names.

### DIFF
--- a/ui/de-lib/src/main/java/org/iplantc/de/client/models/apps/App.java
+++ b/ui/de-lib/src/main/java/org/iplantc/de/client/models/apps/App.java
@@ -140,4 +140,6 @@ public interface App extends HasId,
     void setPermission(PermissionValue value);
 
     PermissionValue getPermission();
+
+    Boolean isBeta();
 }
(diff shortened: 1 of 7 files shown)


ui/de-lib/src/main/java/org/iplantc/de/client/models/apps/App.java
```diff
@@ -29,7 +29,7 @@
     String IS_PUBLIC_KEY = "is_public";
     String PIPELINE_ELIGIBILITY_KEY = "pipeline_eligibility";
     String STEP_COUNT_KEY = "step_count";
-    String SUGGESTED_GROUPS_KEY = "suggested_groups";
+    String SUGGESTED_GROUPS_KEY = "suggested_categories";
```
psarando: Bug-fix rider 👍

psarando:
Looks good to me :shipit:

ash3rz:

PR #290, opened 2016-07-27
CORE-7942 Add ability to delete apps in Belphegor App Catalog tab

ash3rz:


ui/de-lib/src/main/java/org/iplantc/de/admin/apps/client/views/grid/AdminAppsGridImpl.java
```diff
@@ -140,4 +140,12 @@ public App getAppFromElement(Element as) {
     public void deselectAll() {
         grid.getSelectionModel().deselectAll();
     }
+
+    @Override
+    public void removeApp(App selectedApp) {
+        App app = listStore.findModelWithKey(selectedApp.getId());
```
ash3rz: Just to note - this looks redundant, but I'm actually re-fetching the app from the listStore because I have two separate grids with their own listStores which may have their own copies of the same app.

psarando:
Looks good to me :shipit:

ash3rz:

PR #291, opened 2016-07-27
CORE-7946: Update apps and tool listing 'is_public' flags

dennisroberts71:
Prior to this change, the `apps` service used app categories to determine whether or not an app was public. If the app was in any of the public categories then it was considered to be public. Otherwise, it wasn't. The recent app catalog revamp eliminates the need for public app categories altogether, so we need a different way to determine whether or not an app is public.

Fortunately, the implementation of the permissions service makes this relatively easy. If the `de-users` group, which is the group containing all DE users, has access to the app then the app is public by definition. Otherwise, it isn't.

The way that apps are listed for the `Shared With Me` category currently relies on the fact that we only support sharing apps directly with other users. This will have to be changed when we add support for sharing apps with groups. CORE-8092 has been created to address this.


databases/permissions/src/main/conversions/v2.8.0/c280_2016072501.clj
```diff
@@ -0,0 +1,18 @@
...
+  "Adds a version table to the database."
+  []
+  (println "\t* adding a version table to the database.")
+  (load-sql-file "tables/version.sql"))
+
+(defn convert
+  "Performs the conversion for this database version."
+  []
+  (println "Performing the conversion for" version)
+  (add-version-table))
```
dennisroberts71: I added this conversion because I was expecting to have to make other changes to the permissions database. The other changes turned out not to be necessary. This change is necessary, however, so I left it in.

services/apps/src/apps/app_listings.clj
```diff
@@ -1,88 +0,0 @@
-(ns apps.app-listings
```
dennisroberts71: Another zombie namespace vaquished!

services/apps/src/apps/app_validation.clj
```diff
@@ -1,119 +0,0 @@
-(ns apps.app-validation
```
dennisroberts71: And another! No brains for you!

services/apps/src/apps/clients/permissions.clj
```diff
@@ -140,3 +140,19 @@
 (def unshare-app (partial unshare-resource (rt-app)))
 (def share-analysis (partial share-resource (rt-analysis)))
 (def unshare-analysis (partial unshare-resource (rt-analysis)))
+
+(defn- get-public-resource-ids [resource-type]
+  (->> (pc/get-subject-permissions-for-resource-type (client) "group" (ipg/grouper-user-group-id) resource-type false)
```
dennisroberts71: This works by setting the `lookup` flag to `false` in the call to `get-subject-permissions-for-resource-type`. Having the `lookup` flag set to `false` means that only permissions that are assigned directly to the subject will be listed. [To be honest, this is the only available permission search for groups. I thought it was best to have the lookup type explicitly set to `false` anyway.]

services/apps/src/apps/clients/permissions.clj
```diff
@@ -140,3 +140,19 @@
...
+       :permissions
+       (map (comp uuidify :name :resource))
+       set))
+
+(def get-public-app-ids (partial get-public-resource-ids "app"))
+
+(defn- get-directly-accessible-resource-ids [resource-type user]
+  (->> (pc/get-subject-permissions-for-resource-type (client) "user" user resource-type false)
```
dennisroberts71: Similarly, we have the `lookup` flag set explicitly to `false` here. Since the subject is a `user` in this case, however, the `lookup` flag does have an effect on the result set.

services/apps/test/apps/test_fixtures.clj
```diff
@@ -26,16 +15,7 @@
   (let [config-path (getenv "APPS_CONFIG_PATH" default-config-path)]
     (require 'apps.util.config :reload)
     (apps.util.config/load-config-from-file config-path {:log-config? false})
-    (f)))
-
-(defn with-service-uris [f]
```
dennisroberts71: This fixture wasn't being used, and I couldn't think of a good reason to keep it around now that we have `with-config`.

services/apps/test/apps/test_fixtures.clj
```diff
@@ -26,16 +15,7 @@
...
-  (binding [jex-uri                (getenv "JEX_URI" default-jex-uri)
-            data-info-uri          (getenv "DATA_INFO_URI" default-data-info-uri)
-            notification-agent-uri (getenv "NOTIFICATION_AGENT_URI" default-notification-agent-uri)
-            iplant-groups-uri      (getenv "IPLANT_GROUPS_URI" default-iplant-groups-uri)
-            metadata-uri           (getenv "METADATA_URI" default-metadata-uri)
-            permissions-uri        (getenv "PERMISSIONS_URI" default-permissions-uri)]
-    (f)))
+    (metadata-client/with-metadata-base (config/metadata-base) (f))))
```
dennisroberts71: This change compensates for the fact that the integration tests don't go through the Ring middleware.

services/apps/src/apps/service/apps/de/metadata.clj
```diff
@@ -95,7 +95,7 @@
       (throw+ {:type  :clojure-commons.exception/bad-request-field
                :error (str "Rating must be an integer between 1 and 5 inclusive."
                                 " Invalid rating (" rating ") for App ID " app-id)}))
-    (when-not (:is_public app)
+    (when-not (contains? (perms-client/get-public-app-ids) app-id)
       (throw+ {:type  :clojure-commons.exception/bad-request-field
                :error (str "Unable to rate private app, " app-id)}))
```
psarando: Now that app rating isn't tied to confluence wiki pages for public apps, I wonder if it's time to revisit this policy?

psarando:
I didn't catch any issues.
Looks good! :shipit:

dennisroberts71:
Cool, thanks for the review.

PR #292, opened 2016-07-28
DE-1133/CORE-8094: handle spaces in paths in anon-files properly

ianmcorvidae:
We weren't URL decoding... well, anything, in anon-files. So this would probably have also failed with other url-unsafe characters that'll end up in the URI as decoded.

Tested this manually and it does work with the test file listed in the ticket.


psarando:
Looks good to me :shipit:

ianmcorvidae:
Cool, thanks for the review!

PR #293, opened 2016-07-28
CORE-8093 Add static IDs to the tabs within Apps window

ash3rz:
This is adding static IDs to the new hierarchy/category tabs within the Apps window.


psarando:
Looks pretty straight forward :shipit:

ash3rz:
Thanks for the review!

PR #294, opened 2016-07-28
Consul checks expecting specific services (Clojure)

ianmcorvidae:
This pull request (along with a branch on the private ansible of the same name) improves the resiliency of consul service checks in certain sorts of error conditions and our ability to notice issues from the consul service overview page. Particularly, these changes will make it so that if an nginx is still serving requests, but to the wrong service, it's more likely to correctly show as erroring.

It does this by adding an `expecting` parameter to the status endpoints of services, which if provided and equal to something other than the actual service that's being requested, will return a 500 error code (though otherwise the same contents). Thus, if the anon-files nginx has ended up proxying to the apps service for some reason, its check will fail because the apps service endpoint will get `?expecting=anon-files` and return an error.

This PR adds this handling to the clojure services, i.e. where I knew how to add it without needing to do much research. The corresponding branch in private ansible adds this parameter to the service and nginx checks for all services.

The fact this is only clojure services means that consul won't automatically detect the error condition if an nginx is mistakenly pointing at any of the go services it shouldn't be. But, generally in the past when we've had this issue it's been a whole bunch of stuff broken at once, so it's likely we'll know something's generally wrong. It's undoubtedly an improvement even without being perfect, I think!


ianmcorvidae:
nitpicking welcome here, this isn't high-priority enough to leave any minor itches/future technical debt

psarando:
I didn't catch any issues (as long as the data-info endpoint still works as expected using the common `trap`, rather than its own `trap`, but I don't see why it wouldn't)
:shipit:

PR #295, opened 2016-07-29
CORE-8090: Update apps details endpoint to include an app's category hierarchy

psarando:
These changes update the `/apps/:id/details` endpoint to include a `hierarchies` field, populated by a new metadata service `POST /ontologies/{ontology-version}/filter` endpoint, which returns all hierarchies the given target is tagged under.

### DIFF
--- a/ansible/roles/util-cfg-service/templates/apps.properties.j2
+++ b/ansible/roles/util-cfg-service/templates/apps.properties.j2
@@ -27,6 +27,7 @@ apps.workspace.default-app-categories       = {{ apps.user_subs }}
 apps.workspace.dev-app-category-index       = 0
 apps.workspace.favorites-app-category-index = 1
 apps.workspace.public-id                    = 00000000-0000-0000-0000-000000000000
+apps.workspace.metadata.category.attrs      = rdf:type, http://edamontology.org/has_topic
 apps.workspace.metadata.beta.attr.iri       = n2t.net/ark:/99152/h1459
 apps.workspace.metadata.beta.attr.label     = releaseStatus
 apps.workspace.metadata.beta.value          = beta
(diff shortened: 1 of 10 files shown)


ansible/roles/util-cfg-service/templates/apps.properties.j2
```diff
@@ -27,6 +27,7 @@ apps.workspace.default-app-categories       = {{ apps.user_subs }}
 apps.workspace.dev-app-category-index       = 0
 apps.workspace.favorites-app-category-index = 1
 apps.workspace.public-id                    = 00000000-0000-0000-0000-000000000000
+apps.workspace.metadata.category.attrs      = rdf:type, http://edamontology.org/has_topic
```
psarando: I'm not sure about using the configs for storing these metadata AVU attributes.
@ash3rz and I were discussing the possibility of allowing the admins to configure which attributes should be used for storing hierarchy classes as AVU values, but that would require more design discussion and many more changes (including database conversions).

dennisroberts71: That sounds like a good idea to me, but we should be able to add that later without too much trouble.

dennisroberts71:
Looks good to me! :shipit:

ansible/roles/util-cfg-service/templates/apps.properties.j2
ash3rz: It sounds like giving the flexibility to the admins to choose the attributes might take a lot of changes.  I wonder, since we're already adding the attributes to the configs here, if it would make sense to have the IRI-to-attribute mapping in the configs as well somehow?  The UI in that case could get rid of the hard-coded IRI-to-attribute mapping and could be updated to understand the mapping provided in the configs.  (Either way, a design discussion would probably be good)

psarando: I think that's a good idea: we can move these attributes into a couple of group vars that both the UI and this apps config can use.
I'll merge these changes, then we can open another PR for the UI config changes.

PR #296, opened 2016-07-30
CORE-8085: Update ontology hierarchy apps listing endpoint to include all apps in all sub-categories of the hierarchy

psarando:
These changes update the apps service `GET /apps/hierarchies/:iri/apps` endpoint to now include all apps tagged with any class under the hierarchy for the given root IRI.
Also a new apps service `GET /admin/ontologies/:version/:root-iri/apps` endpoint was added for use by admins to preview app listings for different ontology versions (other than the apps service's "active" version).

These endpoints filter their apps listings with a new metadata service `POST /ontologies/:version/:root/filter-targets` endpoint.


services/apps/src/apps/routes/admin.clj
```diff
@@ -243,7 +259,7 @@
   "metadata"
   "POST /ontologies/{ontology-version}/{root-iri}/filter-unclassified"))
         (ok (coerce! AppListing
-                     (listings/get-unclassified-app-listing current-user ontology-version root-iri params)))))
+                     (listings/get-unclassified-app-listing current-user ontology-version root-iri attr params)))))
```
psarando: The logic of the `get-unclassified-app-listing` function has not changed. Its params were only changed to be consistent with the new `list-apps-under-hierarchy` functions.

dennisroberts71:
Looks good to me. :shipit:

psarando:
Thanks for the review!

PR #297, opened 2016-08-03
CORE-8091 Switch to tree paths in AppDetails view

ash3rz:
Since we're moving to hierarchies in 2.8, the app details window has become cluttered with the list of all the paths where an app can be found.  Instead of having a textual path in the app details window, we will have a slimmed down tree to show only the hierarchies where the app resides.

If the app is an HPC app, the details window will show a tree that only has the HPC category.
If the app has not been tagged with any hierarchies, only the Unclassified hierarchies will be listed.

Additionally, clicking on any node in the tree will open up that hierarchy in the Apps window automatically.


strootman:
LGTM

With regard to testing, much of the presenter is covered, but I would put tests in the `getFilteredHierarchies(..)` method. Specifically, testing the view masking. That is a _behavior_ which strongly affects UX.
Not needed for this PR, but should occur.

ash3rz:
Cool, I'll add some more tests before merging.  Thanks for the review!

PR #298, opened 2016-08-03
CORE-7927 Disable actions in App Catalog for external apps

ash3rz:
Once Agave apps are tagged with EDAM ontology terms, agave apps will get listed in the admin App Catalog tab mixed in with DE apps.  This PR makes sure that app categorization is disabled for the admins on any of those external apps.


ash3rz:
I'm going to go ahead and merge this PR since there aren't any crazy changes.

PR #299, opened 2016-08-04
CORE-8084 Add duplicate app name validation for ontology hierarchies

psarando:
These changes update the apps service to validate duplicate app names under ontology hierarchies.

The metadata `POST /avus/filter-targets` endpoint now accepts a list of AVUs (just attr/value pairs for now) rather than a single `attr` and `value`.

The following app edit endpoints now validate duplicate app names in the app's current ontology hierarchy:
* PATCH /admin/apps/:app-id
* PATCH /apps/:app-id
* PUT /apps/:app-id

The following app metadata endpoints now validate duplicate app names in ontology hierarchy AVUs given in the request:
* POST /apps/:app-id/metadata
* PUT /apps/:app-id/metadata

### DIFF
--- a/services/apps/src/apps/service/apps/de/categorization.clj
+++ b/services/apps/src/apps/service/apps/de/categorization.clj
@@ -5,8 +5,42 @@
         [kameleon.entities]
         [apps.validation]
         [slingshot.slingshot :only [throw+]])
-  (:require [apps.persistence.app-metadata :as ap]
-            [apps.service.apps.de.validation :as av]))
+  (:require [apps.clients.permissions :as perms-client]
+            [apps.persistence.app-metadata :as ap]
+            [apps.persistence.categories :as db-categories]
+            [apps.service.apps.de.validation :as av]
+            [apps.util.config :as config]
+            [metadata-client.core :as metadata-client]))
+
+(defn get-active-hierarchy-version
+  []
+  (let [version (db-categories/get-active-hierarchy-version)]
+    (when-not version
+      (throw+ {:type  :clojure-commons.exception/not-found
+               :error "An app hierarchy version has not been set."}))
+    version))
+
+(defn validate-app-name-in-hierarchy-avus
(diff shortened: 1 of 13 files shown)


dennisroberts71:
Looks good to me! :shipit:

psarando:
Thanks for the review!

services/apps/src/apps/service/apps/de/categorization.clj
```diff
@@ -5,8 +5,42 @@
...
+               :error "An app hierarchy version has not been set."}))
+    version))
+
+(defn validate-app-name-in-hierarchy-avus
+  [username app-id app-name avus]
+  (let [category-attrs    (set (config/workspace-metadata-category-attrs))
+        hierarchy-avus    (->> avus
+                               (filter #(contains? category-attrs (:attr %)))
+                               (map #(select-keys % [:attr :value])))
+        other-app-ids     (disj (set (keys (perms-client/load-app-permissions username))) app-id)
```
psarando: I wonder if it would be better to just pass the app's ID along with its name to the [database query](https://github.com/cyverse/DE/pull/299/files#diff-e8fd70ff48696b547e301e34fb2370c3R854), and let SQL exclude this ID?
Maybe with something like
```sql
WHERE id IN ($app_id_set) AND id != $app_id
```

dennisroberts71: Either works as far as I'm concerned. The penalty for a set disjoin in Clojure is small, so you're not likely to get much of a performance boost from excluding the ID of the current app in SQL. I've done most of my searches for duplicate apps using SQL if I remember correctly, but either option works.

psarando: Sounds good. Since I also think the intent of `list-duplicate-apps-by-id` is still clear, I'll leave it this way for now.
Thanks 👍

PR #300, opened 2016-08-04
CORE-8104 Enable admins to add/remove Beta tag to apps

ash3rz:


ash3rz:
I'll merge this PR too, I think it's straightforward.

ui/de-lib/src/main/java/org/iplantc/de/admin/apps/client/presenter/grid/AdminAppsGridPresenterImpl.java
```diff
@@ -70,6 +74,47 @@ public void onSuccess(AppDoc result) {
...
+
+        @Override
+        public void onSuccess(List<Avu> result) {
+            app.setBeta(false);
+            listStore.update(app);
+            announcer.schedule(new SuccessAnnouncementConfig(appearance.betaTagRemovedSuccess()));
+        }
+    }
+
+    class AddAppAvusCallback implements AsyncCallback<List<Avu>> {
```
psarando: Maybe these callbacks could be named something more specific for setting and removing the Beta AVU.
When I first looked at these callbacks, I thought they were generic metadata AVU endpoint callbacks, until I saw them setting the beta flag of the app, which was a little confusing at first.

ui/de-lib/src/main/java/org/iplantc/de/client/util/OntologyUtil.java
```diff
@@ -41,6 +41,11 @@ public int compare(OntologyHierarchy o1, OntologyHierarchy o2) {
     final String UNCLASSIFIED_LABEL = "Unclassified";
     final String UNCLASSIFIED_IRI_APPEND = "_unclassified";
 
+    final String BETA_ATTR = "n2t.net/ark:/99152/h1459";
+    final String BETA_VALUE = "beta";
+    final String BETA_SUB_VALUE = "releaseStatus";
+    final String BETA_SUB_UNIT = "attr";
```
psarando: These values should be probably be stored as config values in the UI's properties file, rather than hard-coded.
In fact, these values are already duplicated in the apps service configs, so they could be refactored into the group_vars for use in both properties file templates.

ash3rz: Oh awesome, I didn't realize they were in the configs already.  Yeah, I'll use the configs for these instead.

ui/de-lib/src/main/java/org/iplantc/de/admin/apps/client/presenter/grid/AdminAppsGridPresenterImpl.java
ash3rz: Ah true, yeah, I'll give them better names.

PR #301, opened 2016-08-04
CORE-7782 Add Agave Auth on log in and App window launch

ash3rz:
In an effort to make it more obvious to our users that HPC apps will not show up without Agave authentication, we are adding an Agave auth prompt in 3 places:

1. When the user first logs in
2. When the user opens the Apps window
3. When the user clicks on the HPC tab/category in the Apps window

Users have the option to authenticate at any of the 3 stages and they also have the option to postpone authentication for another time.


ui/de-lib/src/main/java/org/iplantc/de/desktop/client/presenter/DesktopPresenterImpl.java
```diff
@@ -654,6 +655,11 @@ void restoreWindows(List<WindowState> windowStates) {
         for (WindowState ws : windowStates) {
             desktopWindowManager.show(ws);
         }
+        if (userInfo.hasAgaveRedirect()) {
+            AgaveAuthPrompt prompt = AgaveAuthPrompt.getInstance();
+            prompt.show();
+            desktopWindowManager.stickWindowToTop(prompt);
```
ash3rz: I added this specifically for the case when a user is saving their sessions.  When they first log in, all the windows they had open will reopen and can hide the prompt.  This makes sure the prompt is in focus for the user to interact with first.

ui/de-lib/src/main/java/org/iplantc/de/apps/client/presenter/categories/AppCategoriesPresenterImpl.java
```diff
@@ -164,28 +167,41 @@ public void onFailure(Throwable caught) {
             @Override
             public void onSuccess(List<AppCategory> result) {
                 filterCategories(result);
-                createWorkspaceTab(selectedAppCategory);
-                createHPCTab();
+                addCategoriesToWorkspaceTree(selectedAppCategory);
+                addCategoriesToHPCTree();
+                autoSelectHPCCategory();
```
psarando: Could this be named something like `addHPCTabSelectionHandler`?
I initially thought the HPC category would auto-select every time the apps window was opened 😜

ash3rz: 😝 I'm terrible at naming things.  Yes, your suggestion is much more clear, I'll update it

ui/de-lib/src/main/java/org/iplantc/de/desktop/client/presenter/DesktopPresenterImpl.java
psarando: So does this add a 4th case where the prompt will show?
Since the prompt is already shown when the bootstrap endpoint returns, and the sessions load after the bootstrap call returns.
Could this `stickWindowToTop` call happen when the prompt is shown in the bootstrap callback, instead?

ui/de-lib/src/main/java/org/iplantc/de/apps/client/presenter/categories/AppCategoriesPresenterImpl.java
psarando: heh, I'm not so great at naming things in my own code. Sometimes it just takes a second set of 👀

ui/de-lib/src/main/java/org/iplantc/de/desktop/client/presenter/DesktopPresenterImpl.java
ash3rz: Actually yes, let me update this - good catch!

psarando:
Look good! 👍

ash3rz:
Cool, thanks for the second pair of 👀 !

PR #302, opened 2016-08-05
CORE-8099 Add mask for deprecated apps upon launch

ash3rz:
In 2.8, since the Archive category will be gone, admins will need to delete/deprecate apps in order to keep the app catalog clean.  Since apps may have been used in publications with links to the app, it would be helpful to present users with a message that lets them know the app has been deprecated and to contact support if they need access.


ui/de-lib/src/main/java/org/iplantc/de/apps/widgets/client/view/AppLaunchViewImpl.java
```diff
@@ -42,15 +43,22 @@
     private final EditorDriver editorDriver = GWT.create(EditorDriver.class);
 
     private final LaunchAnalysisView law;
+    private CustomMask customMask;
```
ash3rz: The `CustomMask` and its appearance files are all basically so I can mask the view and NOT have the loading cursor.  I thought having the loading cursor might confuse users into thinking they might still be able to use the app if they wait for the "load" to finish.

psarando:

ash3rz:

PR #303, opened 2016-08-06
CORE-8102 Fix terrain's Permanent ID Requests metadata parsing

psarando:
This fixes the terrain permanent-id-requests service's expected format of data-info's metadata endpoint response (the metadata AVUs are no longer nested under a `metadata` key, after one of the last commits of #196).

Initial integration tests have also been added for terrain.services.permanent-id-requests, with more to be added before merging this PR.

### DIFF
--- a/services/terrain/src/terrain/clients/data_info/raw.clj
+++ b/services/terrain/src/terrain/clients/data_info/raw.clj
@@ -104,14 +104,15 @@
 ;; CREATE
 
 (defn upload-file
-  [user dest-path filename content-type istream]
+  [user dest-path filename content-type istream & {:keys [as] :or {as :stream}}]
   (http/post (str (url/url (cfg/data-info-base) "data"))
              {:query-params {:user user
                              :dest dest-path}
               :multipart [{:part-name "file"
                            :name filename
                            :mime-type content-type
-                           :content istream}]}))
+                           :content istream}]
+              :as as}))
 (defn create-dirs
   "Uses the data-info directories endpoint to create several directories."
   [user paths]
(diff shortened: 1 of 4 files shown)


ianmcorvidae:
The changes and the tests that are here both look good to me, so I'd say :shipit:

I'd love your opinion on something though: while I was looking at this I noticed the data-info swagger docs don't show the fact the metadata endpoints include stuff from the downstream request, since compojure-api doesn't generate anything in swagger for a schema `{s/Keyword s/Any}`. Now that I look again, I see that this is documented in the text, but I wonder if it'd make sense to add a documentation-only schema which puts in something there so it'll appear in the generated documentation. Thoughts?

psarando:
This sounds like a good idea. I'm just not sure how to document this without giving the false sense that this mock-documentation should be referenced over the metadata service documentation.

For example, I don't want to cause confusion (or QA bug reports) if the metadata endpoint changes (a key is added/removed/renamed), but the data-info docs don't get updated to reflect the change. This also might violate the DRY principle, unless the metadata response schemas are stored in the common compojure-api library and are shared between these metadata and data-info endpoints?

ianmcorvidae:
Yeah, that was also my concern. My thought was to put in some fairly-dumb key-value sets, like have it show up in the documentation as `"...": "more keys and values from the metadata service response"`, forcing a look at the real documentation for details.

ianmcorvidae:
I don't like the idea of sharing the schemas, to be honest, because it feels like creating a false dependency -- but, since the actual responses from the data-info service do in fact rely on the specifics of the metadata service response, perhaps it's not a false dependency.

Another option would be for us to force the data-info return value into a particular mold to ensure we can define its schema -- which would probably mean wrapping a `select-keys` or a few around what we get, and possibly doing some sort of validation on what we get from the metadata service. That way we can precisely define both responses and have them operate independently except where they're not independent, i.e. for the specific values data-info cares about.

psarando:
Right, unless we want data-info to validate the response info from the metadata service, then they probably shouldn't use a shared schema; and I'm not sure about having metadata validate its response, and then data-info validating the same info again for its response.

I'm leaning towards your blatantly-obvious-fake-documentation example, like `"...": "more keys and values from the metadata service response"`.

PR #304, opened 2016-08-08
CORE-8114 Update Apps window to parse ontology attrs from configs

ash3rz:
This updates the UI so that instead of hard-coding the attribute values for metadata into the UI, the UI can read the mapping between IRIs and the corresponding attribute from the configs.  The value is the attribute and the key is a regex that should match all IRIs that should be associated with that attribute.  Right now the UI needs to know which attributes apply to which hierarchies in order to update any metadata.

[This is a temporary solution until changes can be made so that admins can select which attributes they want associated with which hierarchies, which would be too many changes to make the 2.8 release.]

### DIFF
--- a/ansible/roles/util-cfg-service/templates/ui/de.properties.j2
+++ b/ansible/roles/util-cfg-service/templates/ui/de.properties.j2
@@ -130,6 +130,7 @@ org.iplantc.discoveryenvironment.workspace.defaultAppCategories      = {{ apps.u
 org.iplantc.discoveryenvironment.workspace.defaultBetaAppCategoryId  = {{ apps.beta_category }}
 org.iplantc.discoveryenvironment.workspace.defaultHpcAppCategoryId   = {{ apps.hpc_category }}
 org.iplantc.discoveryenvironment.workspace.defaultTrashAppCategoryId = {{ apps.trash_category }}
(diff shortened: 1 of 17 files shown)


ansible/roles/util-cfg-service/templates/ui/de.properties.j2
```diff
@@ -130,6 +130,7 @@ org.iplantc.discoveryenvironment.workspace.defaultAppCategories      = {{ apps.u
 org.iplantc.discoveryenvironment.workspace.defaultBetaAppCategoryId  = {{ apps.beta_category }}
 org.iplantc.discoveryenvironment.workspace.defaultHpcAppCategoryId   = {{ apps.hpc_category }}
 org.iplantc.discoveryenvironment.workspace.defaultTrashAppCategoryId = {{ apps.trash_category }}
+org.iplantc.discoveryenvironment.workspace.ontologyAttrs             = {{ apps.ontology_attrs }}
```
psarando: Since this value is only needed in this properties file, we don't need to create a group var for it.
I would just put the literal value here.

ash3rz: Ha, I just did that in my other PR.  Not sure why I spaced out on this PR.  Thanks for catching!

psarando:
Looks good :shipit:

ash3rz:
Cool, thanks for the review!

PR #306, opened 2016-08-10
CORE-8126 prep: support fetching groups for a subject with a folder filter in iplant-groups

ianmcorvidae:
This is in preparation for fetching groups for a subject when an analysis is submitted -- we want to only fetch groups which are within the folder corresponding to the current environment, then remove the environment-folder part of the name and pass it down to jex-adapter.


dennisroberts71:
Looks good to me! :shipit:

ianmcorvidae:
cool, thanks for the review!

PR #307, opened 2016-08-10
CORE-8126: put user groups into the jex submission in apps

ianmcorvidae:
This time actually fixing the ticket! This:

* adds a client function for fetching a subject's groups
* adds a client function for trimming off the environment name
* uses them to add information to the submission that will be sent to the jex in the format I was told to

I've tried this against our dev environment with the newly-deployed jex-adapter and the `+IpcUserGroups` entry I'd expect (`{"users:de-users"}` only -- I'm not in anything else) appeared.


ianmcorvidae:
thanks as always for the quick review!

PR #309, opened 2016-08-12
Core 8124

dennisroberts71:
This change adds some endpoints for managing the workspace group.

### DIFF
--- a/services/apps/src/apps/routes/schemas/groups.clj
+++ b/services/apps/src/apps/routes/schemas/groups.clj
@@ -26,3 +26,31 @@
 
    :id
    (describe String "The group ID.")})
+
+(s/defschema Subject
+  {:id
+   (describe String "The subject ID.")
+
+   (s/optional-key :name)
+   (describe String "The subject name.")
+
+   (s/optional-key :first_name)
+   (describe String "The subject's first name.")
+
+   (s/optional-key :last_name)
+   (describe String "The subject's last name.")
+
+   (s/optional-key :email)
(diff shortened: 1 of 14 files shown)


dennisroberts71:
I'm merging this into dev now because the upcoming work that I have to do depends on this. Of course, anyone who wants to review the code is welcome to do so.

services/apps/src/apps/routes/schemas/groups.clj
```diff
@@ -0,0 +1,72 @@
+(ns apps.routes.schemas.groups
+  (:use [common-swagger-api.schema :only [describe NonBlankString]])
+  (:require [schema.core :as s]))
+
+(s/defschema Group
```
psarando: Some of the schemas in this namespace seem like duplicates of [schemas in iplant-groups](https://github.com/cyverse/DE/tree/8fba9b30622796c94e8003a051e77422aef1b931/services/iplant-groups/src/iplant_groups/routes/schemas).
These common schemas could be moved into the `common-swagger-api` lib so that iplant-groups and apps can share them.
For example, [some ontology schemas](https://github.com/cyverse/DE/tree/8fba9b30622796c94e8003a051e77422aef1b931/libs/common-swagger-api/src/common_swagger_api/schema/ontologies.clj) are shared between the apps and metadata services.

dennisroberts71: Thanks, yeah, that's probably a good idea. I'll create a Jira issue for it.

PR #312, opened 2016-08-15
CORE-8113 Add search to App Catalog tab in Belphegor

ash3rz:
This PR just adds the search bar to the App Catalog tab in Belphegor (and due to space, moves the ontology dropdown inside of the content panel holding the hierarchies on the left side).  

I'll probably end up rearranging the view a little bit in a later PR since this makes things look a bit uneven between the two grids, but I won't make the change in this PR.


ash3rz:
This PR is mostly copy-paste from the Apps code and I know this will have merge conflicts with at least one of my other PRs that I want to take care of sooner rather than later.  I'll go ahead and merge this, but anyone can feel free to look it over afterwards.

ui/de-lib/src/main/java/org/iplantc/de/theme/base/client/admin/ontologies/OntologyDisplayStrings.properties
```diff
@@ -6,7 +6,7 @@ fileUploadMaxSizeWarning = Maximum size of each file is 1.9GB when using simple
 reset = Reset
 fileUploadFailed = The file {0} failed to upload.
 fileUploadSuccess = The file {0} was uploaded successfully.
-publishOntology = Publish this Ontology to the DE
+publishOntology = Publish
```
psarando: Since this string doesn't say anything about ontologies anymore, I'd rename this property to just `publish`.
And it's now generic enough, maybe it can be moved to `de-lib/src/main/resources/org/iplantc/de/resources/client/messages/IplantDisplayStrings.properties`, if it's not too much trouble.

ash3rz: Ah, good point. I'll refactor it into IplantDisplayStrings.

PR #313, opened 2016-08-15
CORE-8110: add a wrapping macro to catch IOExceptions thrown as JargonExceptions in data-info

ianmcorvidae:
NOTE: most of this PR is indentation changes. I'd recommend adding `w=1` to the URL to exclude those lines.

This adds a macro which is intended for wrapping `with-jargon` and which catches `JargonException`s and adds some extra info/a sensible error code when the error is an `IOException`, which I believe should be what happens in these cases (and, hopefully, not in other cases). When it catches these errors, the error that comes out looks like: 

```json
{
  "error_code": "ERR_UNAVAILABLE",
  "reason": "iRODS is unavailable: org.irods.jargon.core.exception.JargonException: java.net.UnknownHostException: ..."
}
```
(UnknownHostException is a type of IOException)

I'm not that happy with this right now because of how many changes it seems to require. But, I'm not happy with putting it into `clj-jargon` directly (too low-level, using our error codes would be adding dependencies to `clj-jargon` and we aren't entirely consistent about how we throw errors right now anyway), nor with what I'd need to do to create a single macro that does both this and what `with-jargon` does (even calling out to `with-jargon`, I'd need to copy a decent chunk of its argument-handling, symbol-manipulating, logic, which is complex enough I'd rather not). I'd appreciate if others have clever ideas on how to do this better, because I bet I'm missing something.

### DIFF
--- a/services/data-info/src/data_info/services/entry.clj
+++ b/services/data-info/src/data_info/services/entry.clj
@@ -4,7 +4,6 @@
   (:require [me.raynes.fs :as fs]
             [clj-icat-direct.icat :as icat]
             [clj-jargon.by-uuid :as uuid]
-            [clj-jargon.init :as init]
             [clj-jargon.item-info :as item]
             [clj-jargon.item-ops :as ops]
             [clj-jargon.permissions :as perm]
@@ -24,15 +23,14 @@
 (defn id-entry
   [url-id user]
   (try
-    (irods/catch-jargon-io-exceptions
-      (init/with-jargon (cfg/jargon-cfg) [cm]
-        (if-not (user/user-exists? cm user)
-          (http-response/unprocessable-entity)
-          (if-let [path (uuid/get-path cm url-id)]
-            (if (perm/is-readable? cm user path)
-              (http-response/ok)
-              (http-response/forbidden))
-            (http-response/not-found)))))
+    (irods/with-jargon-exceptions [cm]
+      (if-not (user/user-exists? cm user)
+        (http-response/unprocessable-entity)
+        (if-let [path (uuid/get-path cm url-id)]
+          (if (perm/is-readable? cm user path)
+            (http-response/ok)
+            (http-response/forbidden))
+          (http-response/not-found))))
     (catch IllegalArgumentException _
       (http-response/unprocessable-entity))))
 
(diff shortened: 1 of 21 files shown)


services/data-info/src/data_info/util/irods.clj
```diff
@@ -11,8 +11,19 @@
             [data-info.util.config :as cfg])
   (:import [clojure.lang IPersistentMap]
            [java.util UUID]
+           [java.io IOException]
+           [org.irods.jargon.core.exception JargonException]
            [org.apache.tika Tika]))
 
+(defmacro catch-jargon-io-exceptions
```
ianmcorvidae: (here's where you're looking for the implementation of the new macro, if you're having trouble digging it out of the whitespace changes)

services/data-info/src/data_info/util/irods.clj
```diff
@@ -11,8 +11,19 @@
...
+  [& body]
+  `(try+
+     (do ~@body)
+     (catch JargonException e#
+       (if (instance? IOException (.getCause e#))
+         (throw+ {:error_code error/ERR_UNAVAILABLE
+                  :reason (str "iRODS is unavailable: " e#)})
```
ianmcorvidae: this "reason" is what will be passed in the JSON response, but the entire stacktrace will be in the logs, in case that's concerning anyone

tedgin:
I think the general concept is about as good as we can do.  Maybe you add could add the call to the with-jargon macro as part of the definition of the catch-jargon-io-exceptions macro. This would reduce a lot of code duplication. Of course, you'd want to rename catch-jargon-io-exceptions to something more descriptive like with-jargon-exn-mapped or something.

ianmcorvidae:
Hm, yeah, I guess that's acceptable-enough since nothing in data-info, at least right now, uses anything but the context-map from with-jargon. I'll try doing that.

services/data-info/src/data_info/services/entry.clj
```diff
@@ -39,11 +38,10 @@
 ;; file specific
 
 (defn- get-file
-  [path]
-  (init/with-jargon (cfg/jargon-cfg) [irods]
-    (if (zero? (item/file-size irods path))
-      ""
-      (ops/input-stream irods path))))
```
ianmcorvidae: (this change is mildly unrelated but there wasn't any reason to have the extra with-jargon here so I killed it)

tedgin:
Looks good to me! :shipit:

PR #314, opened 2016-08-15
CORE-8128: Update /apps search endpoint to also search by category name

psarando:
This PR updates the /apps search endpoint to also search apps by category name.
So the search endpoint should still include the same search results as before, but now it will also include apps with ontology class AVUs attached whose label matches the search term (using the same wildcard logic to match the search term in other app fields).

The /apps search endpoint uses a new metadata service `POST /ontologies/{ontology-version}/filter-targets` endpoint to search apps with matching category labels to include in the /apps search results.


dennisroberts71:
I didn't notice any issues. :shipit:

psarando:
Cool, thanks for the review!

PR #315, opened 2016-08-16
Use futures to parallelize/speed up the terrain bootstrap endpoint

ianmcorvidae:
This probably isn't _that_ important but it did speed things up a noticeable amount in my local testing, from 4.5 to 1.2s for first call, and from ~0.4 to ~0.15s for subsequent calls. It'll hit the apps service a tiny bit harder since it means three simultaneous calls rather than three serial calls, but the get-workspace one is the hardest hitting call anyway.

### DIFF
--- a/services/terrain/src/terrain/services/metadata/apps.clj
+++ b/services/terrain/src/terrain/services/metadata/apps.clj
@@ -48,13 +48,14 @@
   (assert-valid user-agent "Missing or empty request parameter: user-agent")
   (let [username    (:username current-user)
         user        (:shortUsername current-user)
-        workspace   (dm/get-workspace)
-        preferences (user-prefs (:username current-user))
-        login-time  (:login_time (dm/record-login ip-address user-agent))]
+        workspace   (future (dm/get-workspace))
+        preferences (future (user-prefs (:username current-user)))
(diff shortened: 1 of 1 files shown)


services/terrain/src/terrain/services/metadata/apps.clj
```diff
@@ -48,13 +48,14 @@
   (assert-valid user-agent "Missing or empty request parameter: user-agent")
   (let [username    (:username current-user)
         user        (:shortUsername current-user)
-        workspace   (dm/get-workspace)
-        preferences (user-prefs (:username current-user))
-        login-time  (:login_time (dm/record-login ip-address user-agent))]
+        workspace   (future (dm/get-workspace))
+        preferences (future (user-prefs (:username current-user)))
```
ianmcorvidae: There's actually a lot buried in this `user-prefs` call, since it's also in charge of determining if output directories need to be created/doing so/etc.

psarando:
I guess this should be closed and reopened against https://github.com/cyverse-de/terrain

ianmcorvidae:
I'm closing this in favor of https://github.com/cyverse-de/terrain/pull/3

PR #316, opened 2016-08-16
CORE-8125 Switch to DE hierarchy preview instead of AppCategory listing in App Catalog tab

ash3rz:
The App Catalog tab was initially created to serve the transition period from 2.7 to 2.8 where the science team was tagging apps from the old app categories into the new ontology hierarchies.

This PR updates the App Catalog tab into a state where you no longer see the old app categories, instead you get a "preview" of what hierarchies the DE users would see.  Basically now the admins can edit an ontology and have a side-by-side "preview" of that ontology in the DE (which is basically the filtered-down list of hierarchies).


psarando:

ash3rz:

Now I get to rebase all my other PRs on top of this one 😅

PR #317, opened 2016-08-16
CORE-8133: add :username sort to analyses translate-sort-field

ianmcorvidae:
This should stop the UI from throwing an error when you try to sort by this column!


dennisroberts71:
Looks good to me. :shipit:

ianmcorvidae:
Thanks for the review! I'll merge it later today or tomorrow, once deployments and merges and such are done.

PR #318, opened 2016-08-18
CORE-8134 Filter hierarchies for deleted apps.

psarando:
This change updates the `GET /apps/hierarchies/{root-iri}` and `GET /admin/ontologies/{ontology-version}/{root-iri}` endpoints to filter out hierarchies which only contain deleted apps.

### DIFF
--- a/services/apps/src/apps/persistence/app_metadata.clj
+++ b/services/apps/src/apps/persistence/app_metadata.clj
@@ -882,3 +882,12 @@
          category-ids
          (app-category-id-subselect app-id))
        (list-duplicate-apps* app-name app-id)))
+
+(defn filter-visible-app-ids
+  "Filters the given list of app IDs, returning only those not marked as deleted."
+  [app-ids]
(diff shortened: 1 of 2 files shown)


ianmcorvidae:
This looks good if I'm understanding it properly, but I'd like to make sure:

It looks like this is an error condition that happens when either: an app that you had non-public access to is deleted (but still exists) or a public app is deleted (but still exists); theoretically, permissions could be revoked in some fashion in the permissions database, but that might prevent undeleting the apps in the future if desired, so they're retained. But, the permissions service doesn't have any idea which apps are and aren't deleted, so this PR adds a query into the apps database to filter them out.

Assuming I'm correct, LGTM :ship: :boat: :shipit: etc

psarando:
Yes, that is correct. When an app is deleted its permissions aren't changed, and this is currently convenient so that permissions don't need to be modified if an admin un-deletes an app.

Thanks for the review! 👍

PR #319, opened 2016-08-18
CORE-7644 Show non-public apps with green text

ash3rz:
This PR is to address an issue where private apps might have the same name as public apps, which could be confusing for users to figure out which app they intended to use.

Private apps will now be green.

Also, this makes sure that beta apps (blue) and private apps (green) are the same size font as public apps (they were slightly smaller).


psarando:

ash3rz:
Thanks for the review!

PR #320, opened 2016-08-18
CORE-8111 Visually removing the App tab from Belphegor

ash3rz:
This is getting rid of the Apps tab in Belphegor and not initializing the Apps presenter.  

Removing the old apps code will be a PR for another time 😄


psarando:
This looks like it will work for now.

So removing the injected apps presenter from the BelphegorViewImpl constructor, and `org.iplantc.de.shared.DEProperties#getDefaultBetaCategoryId` could be cleaned up when the old apps code is cleaned up (maybe add a `FIXME` comment to those spots as a reminder that they're obsolete).

ash3rz:
Those were easy enough to remove, so I took care of them. 👍

PR #321, opened 2016-08-18
CORE-8130: workshop admin panel in Belphegor

dennisroberts71:
This change adds a UI to Belphegor to allow administrators to manage the list of workshop participants.


services/apps/src/apps/routes/schemas/groups.clj
```diff
@@ -58,15 +58,6 @@
...
-   (describe Boolean "True if the user was added successfully")
-
-   :subject_id
-   (describe NonBlankString "The subject ID.")
-
-   (s/optional-key :subject_name)
-   (describe NonBlankString "The subject name.")})
-
 (s/defschema GroupMembersUpdateResponse
-  {:results (describe [GroupMemberSubjectUpdateResponse] "The list of membership update results.")})
```
dennisroberts71: I changed the response body of the `PUT /admin/groups/workshop/members` endpoint in order to prevent the UI from having to chain service calls to refresh the member list.

services/terrain/src/terrain/clients/iplant_groups.clj
```diff
@@ -10,6 +10,7 @@
   {:username (:id response)
    :firstname (:first_name response)
    :lastname (:last_name response)
+   :name (:name response)
```
dennisroberts71: I added the user's full name (from the `CN` field in LDAP) to the response so that we can avoid having to reassemble the user's name.

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/workshopAdmin/presenter/WorkshopAdminPresenterImpl.java
```diff
@@ -0,0 +1,178 @@
...
+                return;
+            }
+
+            // Add the user to the list if not there already.
+            Member member = memberFromCollaborator(event.getCollaborator());
+            if (!listContainsMember(listStore.getAll(), member)) {
+                listStore.add(member);
+            }
+        }
+    }
```
dennisroberts71: I've seen other cases where the presenter implements all of the event handler interfaces instead. I considered using that approach here, but opted for inner classes instead just because inner classes give me more control over which methods can be called.

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/workshopAdmin/view/WorkshopAdminViewImpl.java
```diff
@@ -0,0 +1,122 @@
...
+        this.listStore = listStore;
+        initWidget(uiBinder.createAndBindUi(this));
+        grid.getSelectionModel().setSelectionMode(Style.SelectionMode.MULTI);
+    }
+
+    @Override
+    public <H extends EventHandler> void addGlobalEventHandler(GwtEvent.Type<H> type, H handler) {
+        HandlerRegistration registration = EventBus.getInstance().addHandler(type, handler);
+        globalHandlerRegistrations.add(registration);
+    }
```
dennisroberts71: The only reason we needed the global event bus here was because the `UserSearchField` uses it.

ui/de-lib/src/main/java/org/iplantc/de/collaborators/client/views/ManageCollaboratorsViewImpl.java
```diff
@@ -63,7 +63,7 @@
     public ManageCollaboratorsViewImpl(final ListStore<Collaborator> store,
                                        final MODE mode) {
         this.listStore = store;
-        searchField = new UserSearchField(USER_SEARCH_EVENT_TAG.MANAGE);
+        searchField = new UserSearchField(USER_SEARCH_EVENT_TAG.WORKSHOP_ADMIN);
```
dennisroberts71: Heh, this was an error. I'll fix it.

services/apps/src/apps/routes/schemas/groups.clj
ianmcorvidae: This is minor, but it might be nice to implement this schema as an addition to the existing `GroupMembers` schema rather than something separate (i.e. something like `(assoc GroupMembers :failures (...))`). I tend to think this is better since it conveys the intent better -- it's a list of the new group members with an extra field added, plus it makes it a bit more DRY.

(but, definitely not a bug or something that has to be fixed, just food for thought)

dennisroberts71: Good point. I'll update it.

services/terrain/src/terrain/clients/iplant_groups.clj
ianmcorvidae: Always in favor of being a little bit less wrong as far as the entries in https://www.kalzumeus.com/2010/06/17/falsehoods-programmers-believe-about-names/ :D

ianmcorvidae:
sadly I probably can't do much meaningful review of the UI, so I'll leave that to someone with a better grasp. But the parts that aren't the UI look good to me!

dennisroberts71:
Cool, thanks for the review.

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/workshopAdmin/presenter/WorkshopAdminPresenterImpl.java
```diff
@@ -0,0 +1,178 @@
...
+
+    private final class DeleteMembersClickedEventHandler
+            implements DeleteMembersClickedEvent.DeleteMembersClickedEventHandler {
+
+        @Override
+        public void onDeleteMembersClicked(DeleteMembersClickedEvent event) {
+            for (Member member : event.getMembers()) {
+                listStore.remove(member);
```
ash3rz: (More curiosity about workshops than a comment on your code since I'm not familiar with how the workshops will be run) 

When members are deleted, here we're deleting them from the listStore with no calls to the service.  So this makes me think the Workshop Admin is intended to only handle one ongoing workshop at a time, or there could be other things going on with the service when saving members so I'm not seeing the full picture here.

dennisroberts71: We normally don't have multiple simultaneous workshops, but there's no reason that we couldn't. The purpose of this group is to direct jobs submitted by users in workshops to a dedicated pool of Condor execute nodes, so we can accommodate simultaneous workshops by adding the users in all of the workshops to the workshop group.

I chose not to call the service when users are added to or removed from the list store because the service call is fairly slow. No changes are made until the user clicks the Save button. Hopefully, that won't be too confusing for workshop administrators.

ash3rz: Gotcha.  I misunderstood the setup of the workshops. I thought the workshop group was more complicated in that members of workshop A get to use Group A of Condor nodes which are exclusive to them, while members of workshop B use Group B, etc.

dennisroberts71: Ah, yeah, we don't have to support that capability right now. Hopefully we won't have to support it in the future either. 😄

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/workshopAdmin/presenter/WorkshopAdminPresenterImpl.java
ash3rz: I think this will be fine.  The interfaces are particularly nice when you need multiple handlers for the same event, then they only need to each implement the interface.

ash3rz:
I didn't see any showstoppers and we already have plans for refactoring out the global event bus when there's more time 👍

I'll have some small changes here and there so that the Mockito tests will work, but I think I'll create a branch off your branch for that.

dennisroberts71:
Cool, thanks for the review. I think I'll merge this branch in now so that we can get it into our development deployments as soon as possible.

ui/de-lib/src/main/java/org/iplantc/de/client/models/groups/Member.java
```diff
@@ -0,0 +1,52 @@
...
+    @AutoBean.PropertyName("email")
+    String getEmail();
+
+    @AutoBean.PropertyName("email")
+    void setEmail(String email);
+
+    @AutoBean.PropertyName("institution")
+    String getInstitution();
+
+    @AutoBean.PropertyName("institution")
```
psarando: By the way, these PropertyNames for `id`, `email`, and `institution` are redundant.
The PropertyName only needs to be specified if the JSON key doesn't match (case insensitive) the method name after the get/set. So the PropertyNames with the underscores in them are required.

dennisroberts71: Ah, good to know! Thanks.

PR #322, opened 2016-08-19
CORE-8105 Fix Trash app category listing schema validation errors.

psarando:
This PR fixes schema validation errors in the `GET /apps/categories/:trash-id` app listings endpoint.

The base select query fields from the `kameleon.app-listing/get-app-listing-base-query` were refactored into a new function that could be used by the trash listing query as well, which was previously selecting all columns in the `app_listing` view, causing the listing endpoint to throw schema validation errors for the extra fields.

I also refactored some other query building steps from `get-app-listing-base-query` into new functions, to clean that function up a bit more.


ianmcorvidae:
LGTM :shipit:

psarando:
Thanks for the review 👍

PR #323, opened 2016-08-19
CORE-8132 Highlight hierarchies in App Details that match app search …

ash3rz:
…terms

If you conduct a search in the Apps window, the results returned will be matched on app name, tool name, integrator name, app description, and now also hierarchy name.  This PR highlights the matching text from the hierarchy names in the App Details window.


psarando:
Looks good to me 👍 
I wonder if there's a way to auto-expand the tree to the matching categories, but that could definitely go into another PR.
:shipit:

ash3rz:
Agreed, I started to look into that and it seemed like I'd have to do something that would involve matching twice - once to get a list of which nodes I need to have the tree auto-expand, and again when the text becomes visible to highlight the relevant text.  There might be another way to do it that I missed though.

PR #324, opened 2016-08-22
CORE-8112 Loading mask fix for App Catalog and Categorize dialog fix

ash3rz:
Currently, if you click on the Trash category in the App Catalog tab, both the editor grid and the preview grid get masked when only the editor grid should get masked.

Also, when you try to categorize an app, the Trash category is listed when it should not be.  Deleting an app should be done through either drag-and-drop to the Trash category or clicking the Delete App button.

I also renamed TreeType to ViewType (most of the modifications in this PR), since I needed a way to choose which grid to mask and selecting the grid based on a "TreeType" did not make sense.

### DIFF
--- a/ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/ontologies/OntologiesView.java
+++ b/ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/ontologies/OntologiesView.java
@@ -50,7 +50,7 @@ public interface OntologiesView extends IsWidget,
                                         RefreshPreviewButtonClicked.HasRefreshPreviewButtonClickedHandlers,
                                         RestoreAppButtonClicked.HasRestoreAppButtonClickedHandlers {
 
-    enum TreeType {
+    enum ViewType {
         ALL, EDITOR, PREVIEW
     }
 
@@ -67,21 +67,21 @@ public interface OntologiesView extends IsWidget,
 
     Ontology getSelectedOntology();
 
-    void clearTreeStore(TreeType type);
+    void clearTreeStore(ViewType type);
 
-    void addToTreeStore(TreeType type, List<OntologyHierarchy> children);
+    void addToTreeStore(ViewType type, List<OntologyHierarchy> children);
 
-    void addToTreeStore(TreeType type, OntologyHierarchy parent, List<OntologyHierarchy> children);
+    void addToTreeStore(ViewType type, OntologyHierarchy parent, List<OntologyHierarchy> children);
 
-    void maskTree(TreeType type);
+    void maskTree(ViewType type);
 
     void selectHierarchy(OntologyHierarchy hierarchy);
 
     void selectActiveOntology(Ontology activeOntology);
 
-    void deselectHierarchies(TreeType type);
+    void deselectHierarchies(ViewType type);
 
-    void reSortTree(TreeType type);
+    void reSortTree(ViewType type);
 
     void updateButtonStatus();
 
@@ -93,7 +93,7 @@ public interface OntologiesView extends IsWidget,
 
     void deselectAll();
 
-    void unmaskTree(TreeType type);
+    void unmaskTree(ViewType type);
 
 
     interface OntologiesViewAppearance {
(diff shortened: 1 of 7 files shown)


ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/ontologies/OntologiesView.java
```diff
@@ -50,10 +50,11 @@
                                         RefreshPreviewButtonClicked.HasRefreshPreviewButtonClickedHandlers,
                                         RestoreAppButtonClicked.HasRestoreAppButtonClickedHandlers {
 
-    enum TreeType {
+    enum ViewType {
         ALL, EDITOR, PREVIEW
     }
 
+    String TRASH_CATEGORY = "Trash";
```
psarando: Should this still be `static final`?

ash3rz: Aren't interface variables static and final already?  When I copy-pasted it here, IntelliJ complained that they were redundant too.

psarando: Good point, just from looking at this diff, I forgot that the view is an interface 👍

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/ontologies/views/OntologiesViewImpl.java
```diff
@@ -455,21 +455,41 @@ public void updateButtonStatus() {
         deleteButton.setEnabled(selectedOntology != null && selectedOntology != activeOntology);
         deleteHierarchy.setEnabled(selectedOntology != null && editorTree.getSelectionModel().getSelectedItem() != null);
         categorize.setEnabled(selectedOntology != null && targetApp != null && !targetApp.getAppType().equalsIgnoreCase(App.EXTERNAL_APP));
-        deleteApp.setEnabled(selectedOntology != null && targetApp != null && !targetApp.getAppType().equalsIgnoreCase(App.EXTERNAL_APP));
+        deleteApp.setEnabled(selectedOntology != null && targetApp != null && !targetApp.getAppType().equalsIgnoreCase(App.EXTERNAL_APP) && !targetApp.isDeleted());
```
psarando: Not necessary for this PR, but I would consider refactoring these conditionals into booleans.
It might make this section a little more readable.
For example
```java
boolean ontologySelected = selectedOntology != null;
boolean appSelected = targetApp != null;
boolean appExternal = targetApp.getAppType().equalsIgnoreCase(App.EXTERNAL_APP);
boolean appDeleted = targetApp.isDeleted();
...
deleteApp.setEnabled(ontologySelected && appSelected && !appExternal && !appDeleted);
```
These could even be named with an `is` prefix (isAppSelected) if preferred.

ash3rz: Ah ok, much more readable.  I'll go ahead and add it now.

ui/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/ontologies/views/OntologiesViewImpl.java
```diff
@@ -111,6 +111,14 @@
     private OntologyUtil ontologyUtil = OntologyUtil.getInstance();
     private PagingLoader<FilterPagingLoadConfig, PagingLoadResult<App>> loader;
 
+    boolean ontologySelected;
+    boolean isActiveOntology;
+    boolean editorTreeHasSelection;
+    boolean hasAppSelected;
+    boolean isExternalApp;
+    boolean isDeletedApp;
+    boolean editorTreeHasHierarchies;
```
psarando: I don't think we need to keep track of this state outside of `updateButtonStatus`, so these can just be local vars in that method.

ash3rz: Oops, you're right, one sec.

psarando:
Looks good! :shipit: 👍

ash3rz:
Thanks for the review!

PR #326, opened 2016-08-22
CORE-7544 Fix reference genome path/name validation.

psarando:
When attempting to rename or update the path of a reference genome, the `PATCH /admin/reference-genomes/` endpoint would return an error like
```json
{"error_code":"ERR_EXISTS","reason":"Another reference genome with the given name already exists."}

```

The `not=` symbol needs to be quoted when used outside of the korma `where` macro.
Otherwise the path and name validation conditionals were parsed as queries like
```sql
WHERE (? AND "genome_reference"."path" = ?)
```
which always found a match when attempting to update just the path or name of an existing reference genome.

Also removed the unused `apps.metadata.reference-genomes/get-all-reference-genomes` function.

### DIFF
--- a/services/apps/src/apps/metadata/reference_genomes.clj
+++ b/services/apps/src/apps/metadata/reference_genomes.clj
@@ -50,11 +50,6 @@
       (where {:id [in uuids]}))
     (select (reference-genome-base-query))))
 
-(defn get-all-reference-genomes
-  "Lists all of the reference genomes in the database, including those marked as deleted."
-  []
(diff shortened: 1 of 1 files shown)


dennisroberts71:
Looks good to me! :shipit:

ianmcorvidae:
Indeed. I don't know the internals of that korma macro particularly well, but it makes sense.

psarando:
Thanks for the reviews!

PR #327, opened 2016-08-24
CORE-8155: fix call signatures for job sharing validations

ianmcorvidae:
`supports-job-sharing?` was changed in 0e5aae1abac55d609d51b6070307f0e5b1e8fd64 to take a list of job steps rather than a job ID, but not all of the places it was called were updated.

This adds another function which does the lookup of job steps from job ID that the old function did, and renames both functions to be more descriptive, updating their call sites as appropriate.


dennisroberts71:
Looks good to me. :shipit:

ianmcorvidae:
Thanks for the review!

PR #329, opened 2016-08-26
Use temporary tables rather than CTEs to speed up certain paged directory listing queries

ianmcorvidae:
PostgreSQL is rubbish at predicting the number of rows for CTE Scans, which goes really badly when the answer is 16000 and it guesses 25.

It doesn't have this issue with temporary tables being created instead, so this does that.

### DIFF
--- a/libs/clj-icat-direct/src/clj_icat_direct/icat.clj
+++ b/libs/clj-icat-direct/src/clj_icat_direct/icat.clj
@@ -3,7 +3,6 @@
   (:require [clojure.string :as string]
             [korma.db :as db]
             [korma.core :as k]
-            [clojure.java.jdbc :as jdbc]
             [clj-icat-direct.queries :as q])
   (:import [clojure.lang ISeq Keyword]))
 
@@ -48,8 +47,8 @@
     (let [side-effects (drop-last 1 queries)
           final-query-and-args (last queries)]
       (doseq [q side-effects]
-        ; uses jdbc/execute! because this can run DDL statements like CREATE TEMPORARY TABLE
-        (jdbc/execute! db/*current-conn* q))
+        ; uses exec-raw directly here to exclude :results so this can run DDL statements
+        (k/exec-raw icat q))
       (apply run-query-string final-query-and-args))))
 
 (defn number-of-files-in-folder
--- a/libs/clj-icat-direct/src/clj_icat_direct/icat.clj
+++ b/libs/clj-icat-direct/src/clj_icat_direct/icat.clj
@@ -3,6 +3,7 @@
   (:require [clojure.string :as string]
             [korma.db :as db]
             [korma.core :as k]
+            [clojure.java.jdbc :as jdbc]
             [clj-icat-direct.queries :as q])
   (:import [clojure.lang ISeq Keyword]))
 
@@ -37,6 +38,19 @@
    clj-icat-direct.queries first."
   [query & args]
   (k/exec-raw icat [query args] :results))
+
+(defn- run-transaction
+  "Runs the set of passed-in query strings+args in a transaction, returning
+  what the last one returns. This is intended to be used to create temporary
+  tables to circumvent the fact postgresql is bad at doing estimates on CTEs."
+  [& queries]
+  (db/transaction
+    (let [side-effects (drop-last 1 queries)
+          final-query-and-args (last queries)]
+      (doseq [q side-effects]
+        ; uses jdbc/execute! because this can run DDL statements like CREATE TEMPORARY TABLE
+        (jdbc/execute! db/*current-conn* q))
+      (apply run-query-string final-query-and-args))))
 
 (defn number-of-files-in-folder
   "Returns the number of files in a folder that the user has access to."
@@ -212,14 +226,16 @@
                      :file   q/mk-paged-files-in-folder
                      :folder q/mk-paged-folders-in-folder
                              (throw (Exception. (str "invalid entity type " entity-type))))
-        query (query-ctor
+        queries (query-ctor
                 :user           user
                 :zone           zone
                 :parent-path    folder-path
                 :info-type-cond (q/mk-file-type-cond info-types)
                 :sort-column    (resolve-sort-column sort-column)
-                :sort-direction (resolve-sort-direction sort-direction))]
-    (map fmt-info-type (run-query-string query limit offset))))
+                :sort-direction (resolve-sort-direction sort-direction)
+                :limit limit
+                :offset offset)]
+    (map fmt-info-type (apply run-transaction queries))))
 
 
 (defn select-files-with-uuids
--- a/libs/clj-icat-direct/src/clj_icat_direct/icat.clj
+++ b/libs/clj-icat-direct/src/clj_icat_direct/icat.clj
@@ -80,12 +80,12 @@
      info types."
   [^String user ^String zone ^String folder-path ^Keyword entity-type & [info-types]]
   (let [type-cond (q/mk-file-type-cond info-types)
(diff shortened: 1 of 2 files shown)


ianmcorvidae:
(I may still be doing more on this, but I wanted to get the main part up)

libs/clj-icat-direct/src/clj_icat_direct/queries.clj
```diff
@@ -310,13 +314,13 @@
     It returns the properly formatted SELECT query."
   [^String user ^String zone ^String parent-path ^String info-type-cond]
   (let [group-query "SELECT group_user_id FROM groups"]
-    (str "WITH groups    AS (" (mk-groups user zone) "),
-               objs      AS (" (mk-unique-objs-in-coll parent-path ) "),
-               file_avus AS (" (mk-obj-avus "SELECT data_id FROM objs") ")
-         " (mk-count-objs-of-type "objs" "file_avus" group-query info-type-cond))))
+    [[(mk-temp-table "groups" (mk-groups user zone))]
+     [(mk-temp-table "objs" (mk-unique-objs-in-coll parent-path))]
+     [(mk-temp-table "file_avus" (str "SELECT object_id, meta_attr_value, meta_attr_name FROM (" (mk-obj-avus "SELECT data_id FROM objs") ") obj_avus"))]
```
ianmcorvidae: this (and other ones below) have this subselect structure because otherwise they don't work because there's some conflicting column names. extra credit work for me or a reviewer would be to see if everywhere that uses `mk-obj-avus` can just be limited to these so we don't need it

tedgin:
Nice find!  Do you find the info about CTEs sucking at estimation in an article, or did you know about this just from experience? Also, could the estimation problem be solved through the PostgreSQL tunables in postgres.conf?

libs/clj-icat-direct/src/clj_icat_direct/queries.clj
dennisroberts71: I can't guarantee that I found all of the columns that were used, but it appears that the temporary table (or CTE, depending on the context) for AVUs is aliased to two names `f` and `m`. There were some cases where `m.meta_id` was referenced directly, but those were all cases where `m` was a direct alias to `r_meta_main` instead. I can't be absolutely certain at this point, but I believe that you can limit the fields selected by `mk-obj-avus`.

libs/clj-icat-direct/src/clj_icat_direct/icat.clj
```diff
@@ -38,6 +39,19 @@
...
+(defn- run-transaction
+  "Runs the set of passed-in query strings+args in a transaction, returning
+  what the last one returns. This is intended to be used to create temporary
+  tables to circumvent the fact postgresql is bad at doing estimates on CTEs."
+  [& queries]
+  (db/transaction
+    (let [side-effects (drop-last 1 queries)
+          final-query-and-args (last queries)]
+      (doseq [q side-effects]
+        ; uses jdbc/execute! because this can run DDL statements like CREATE TEMPORARY TABLE
```
psarando: Does this mean `k/exec-raw` can't run DDL statements?

dennisroberts71:
I agree. This was a very nice find!

libs/clj-icat-direct/src/clj_icat_direct/icat.clj
dennisroberts71: I've used `k/exec-raw` for DDL statements in the past.

dennisroberts71:
I don't see any major issues. This gets a :shipit: from me.

libs/clj-icat-direct/src/clj_icat_direct/icat.clj
psarando: I'd prefer if we could use `k/exec-raw` over adding a direct dependency on `jdbc`.

ianmcorvidae: I used `run-query-string` (and therefore `k/exec-raw`) initially, but got an error "PSQLException No results were returned by the query. org.postgresql.jdbc2.AbstractJdbc2Statement.executeQuery (AbstractJdbc2Statement.java:303)" (I did a search for that, which is where I pulled it from, so might be reformatted). The searching I did suggested that the issue was using the wrong jdbc function. In particular, `exec-raw` calls out to `korma.db/do-query` which calls out to `korma.db/exec-sql`, which uses `query`. The stuff I saw recommended using `executeUpdate`, where `query` uses `executeQuery` (IIRC); nothing in korma uses `executeUpdate`. But now that I'm looking in the morning, it looks like maybe there's something about configuring whether results are returned, which uses something that might also work. I'll test if turning off results fixes this.

ianmcorvidae:
@tedgin kind of a mix, mostly observation/experience. I've seen a decent number of queries in the past which have had bad misestimations due to it choosing either 1 or 25 (I've never seen any other value, that I can think of) for the size of a CTE. I tried to find an article or something to cite about this, but couldn't really find much -- there's a few threads on the postgresql mailing lists about misestimation issues using CTEs which seemed potentially related (and one which recommended using temporary tables instead where possible), and there's stuff all over the place about CTEs being optimization fences (but that's a different problem than this, as far as I can determine -- it's possible that inlining these rather than using WITH or temp tables would have different performance characteristics too though). Given the little bit I found, I thought using temporary tables would be a nice thing to try and was pleasantly surprised at how well it worked!

libs/clj-icat-direct/src/clj_icat_direct/icat.clj
psarando: That's probably the issue.
`run-query-string` calls `(k/exec-raw icat [query args] :results)`.
Maybe using `(k/exec-raw q)` here would prevent that exception.

ianmcorvidae: Yup, it seems to. Pushing that up now.

libs/clj-icat-direct/src/clj_icat_direct/queries.clj
ianmcorvidae: Checking again in the AM when I'm fresher and I think you're right. I'll make that change, it'll make the queries clearer.

ianmcorvidae:
Just as a last comment to get our heads around this and why it works:

I just looked again and I think I understand the "bad CTE estimation" issue a bit better. https://explain.depesz.com/s/hBnY is the explain for the original query here, but only some of the CTE Scans have estimates of 1. Most are estimated at 25, which seems to also be the estimate it uses for the Index Scan which it's using to create the `objs` CTE; this is probably based on a histogram-type average value. The ones which estimate a row count of 1 seem to be an estimate based on the guess of 25 combined with something about the selectivity which brings it down to 1 or below. Basically, even one bad estimate quickly compounds since the CTEs all depend on each other.

i.e.: What this all basically boils down to is: these are huge tables, so the statistics about them are not necessarily that good for really extreme values, and since we used a chain of CTEs that depended on each other, one bad estimate could have an outsize effect, since CTEs are estimated based on the queries that make them up, not any sort of actual analysis (after all, query planning is done from only statistics, not actual values, and isn't changed once the query starts executing). Switching to temporary tables, by contrast, makes there be statistics about the exact sets of relevance each time a query's executed (since each `CREATE TEMPORARY TABLE AS SELECT` and `SELECT` gets its own independent query planning and execution stages), so the counts will be correct (at the cost of, presumably, a little bit of overhead to do the analysis, plus the cost (irrelevant in this case) of things not being part of a completely consistent MVCC snapshot without modifying the transaction isolation levels (which has its own downsides for performance)).

PR #330, opened 2016-09-01
CORE-8163 Add more static IDs to Belphegor

ash3rz:
This PR is to add more static IDs to Belphegor as requested by QA.
- Each tab in Belphegor should have its own ID (switched to `DETabPanel` class instead of Sencha's `TabPanel` which allows for the tabs to have IDs)
- The App Catalog tab didn't have any IDs so those were added
  - I created `DETree` abstract class which allows any child class to specify a `generateDebugID` method for debug IDs on the tree nodes
- The column headers in grids all got static IDs (with use from StaticIDHelper class)
- I also patched up a random NPE that happens if you have saved sessions enabled, save a session with no windows open, and then log in.

### DIFF
--- a/ui/de-lib/src/main/java/org/iplantc/de/admin/apps/client/views/grid/AdminAppsGridImpl.java
+++ b/ui/de-lib/src/main/java/org/iplantc/de/admin/apps/client/views/grid/AdminAppsGridImpl.java
@@ -128,6 +128,7 @@ public class AdminAppsGridImpl extends ContentPanel implements AdminAppsGridView
     protected void onEnsureDebugId(String baseID) {
         super.onEnsureDebugId(baseID);
 
+        acm.ensureDebugId(baseID);
         grid.asWidget().ensureDebugId(baseID + Belphegor.AppIds.GRID);
     }
 
--- a/ui/de-lib/src/main/java/org/iplantc/de/admin/apps/client/views/grid/AdminAppsGridImpl.java
+++ b/ui/de-lib/src/main/java/org/iplantc/de/admin/apps/client/views/grid/AdminAppsGridImpl.java
@@ -11,6 +11,7 @@ import org.iplantc.de.apps.client.events.selection.AppCategorySelectionChangedEv
 import org.iplantc.de.apps.client.events.selection.AppInfoSelectedEvent;
 import org.iplantc.de.apps.client.events.selection.AppSelectionChangedEvent;
 import org.iplantc.de.client.models.apps.App;
+import org.iplantc.de.client.util.StaticIdHelper;
 
 import com.google.common.base.Joiner;
 import com.google.gwt.core.client.GWT;
@@ -25,6 +26,7 @@ import com.google.inject.assistedinject.Assisted;
 
 import com.sencha.gxt.data.shared.ListStore;
 import com.sencha.gxt.widget.core.client.ContentPanel;
+import com.sencha.gxt.widget.core.client.event.ViewReadyEvent;
 import com.sencha.gxt.widget.core.client.grid.ColumnModel;
 import com.sencha.gxt.widget.core.client.grid.Grid;
 import com.sencha.gxt.widget.core.client.grid.GridView;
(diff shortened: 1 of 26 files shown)


ui/de-lib/src/main/java/org/iplantc/de/admin/apps/client/views/grid/AdminAppsGridImpl.java
```diff
@@ -125,10 +127,19 @@ public void onSelectionChanged(SelectionChangedEvent<App> event) {
     }
 
     @Override
-    protected void onEnsureDebugId(String baseID) {
+    protected void onEnsureDebugId(final String baseID) {
         super.onEnsureDebugId(baseID);
 
+        acm.ensureDebugId(baseID);
         grid.asWidget().ensureDebugId(baseID + Belphegor.AppIds.GRID);
+        grid.addViewReadyHandler(new ViewReadyEvent.ViewReadyHandler() {
```
ash3rz: I did this for all the views in Belphegor.  I'm not sure if `onEnsureDebugId` is the best place to put it since we're adding a handler to the grid, but it's only for setting the debug IDs, so I put it here for now.

ui/de-lib/src/main/java/org/iplantc/de/client/util/StaticIdHelper.java
```diff
@@ -0,0 +1,65 @@
...
+     * Sets the static IDs for the column headers in a grid
+     * This does not work if called simply within a view's onEnsureDebugId method
+     * as the headers will not have been rendered yet.
+     * This can be called successfully using the grid's ViewReadyHandler (which can
+     * then be added to the grid within the onEnsureDebugId method)
+     * @param baseID
+     * @param grid
+     * @param cm
+     */
+    public void gridColumnHeaders(String baseID, Grid<?> grid, ColumnModel<?> cm) {
```
ash3rz: QA asked specifically for IDs on each column header, and this was the best I could come up with.  From what I saw, Sencha's Grid/GridView and ColumnModel/ColumnHeader classes were too intertwined to find an easy way to extend a single class and set those IDs for each column header.

Sencha's ColumnHeader appearance gives me access to the header style class name, which I then use to find the whole column header's `table` and its inner `td` elements which are each individual column header.

ui/de-lib/src/main/java/org/iplantc/de/commons/client/widgets/DETree.java
```diff
@@ -0,0 +1,67 @@
...
+        }
+    }
+
+    @Override
+    protected String register(M m) {
+        String id = generateModelId(m);
+        if (nodes.containsKey(id)) {
+            return nodes.get(id).getDomId();
+        } else {
+            String domId = generateDebugId(m);
```
ash3rz: In Sencha's `Tree` class, the `register` method directly calls `XDOM.getUniqueId()` to create the `domId` for each TreeNode. So for `DETree` I removed that and defined a `generateDebugId` method that child classes can implement and set the IDs however they want.

ui/de-lib/src/main/java/org/iplantc/de/desktop/client/presenter/DesktopPresenterImpl.java
```diff
@@ -656,8 +656,10 @@ void postBootstrap(final Panel panel) {
    }
 
     void restoreWindows(List<WindowState> windowStates) {
-        for (WindowState ws : windowStates) {
-            desktopWindowManager.show(ws);
+        if (windowStates != null && windowStates.size() > 0) {
```
ash3rz: Random NPE fix.

psarando:
Reminder: These changes will have to be ported to https://github.com/cyverse-de/ui

ash3rz:
Oh right, whoops.  Should I close this PR and reopen there?

psarando:
At this point, either way, since the code should be the same right now; but that's what I would do.

ash3rz:
Sure, might as well get in the habit now - I'll close this and reopen in the new repo.

